PR #130, opened 2013-05-30
Pull test results out of the VM

thommay:
When running in Jenkins it's rather convenient to be able to utilise minitest's junit output. This adds some config to allow one to do so.

I don't think this should be merged as is - I dislike having to specify the result for each platform, and it doesn't yet cope with multiple platforms all that well - it'll overwrite rather than having discreet directories per platform. But I want to start a discussion as to how best to proceed.


cheeseplus:
Closing this in favor of https://github.com/test-kitchen/test-kitchen/pull/783 or an as yet unimplemented PR.

PR #201, opened 2013-09-18
Add config parameter to use a local web proxy (with health checking)

aerickson:
Adds support for the following new config parameters:

``` text
:use_local_web_proxy (defaults to false)
:local_web_proxy_port (defaults to 8123, polipo default)
```

If :use_local_web_proxy is enabled, the code will attempt to connect to that proxy and verify that it's working. The external address IP address of the host computer is determined by inspecting the default gateway. If it is working it will be used, if it's not working it will be disabled.

http_proxy and https_proxy work well for organizations that run a centralized cache, but don't serve use cases where the developer runs a local cache.

Maximum speed gains are realized by rewriting the urls in the install.sh from https to http to enable caching (proxies that work with SSL content only allow it to pass -- they cannot cache it). See my master branch (aerickson/test-kitchen) for that modification. Ideally the install.sh would just be changed to http links (the advantages of using HTTPS aren't clear).

---

``` yaml
driver_plugin: vagrant
driver_config:
  use_local_web_proxy: true
  # local_web_proxy_port: 3128 # squid default port
```

### DIFF
--- a/lib/kitchen/driver/ssh_base.rb
+++ b/lib/kitchen/driver/ssh_base.rb
@@ -16,6 +16,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+require "net/https"
+require "net/http"
+require "socket"
+require "uri"
+
+PROXY_TIMEOUT = 2
+
 module Kitchen
 
   module Driver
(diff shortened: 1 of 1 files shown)


cheeseplus:
If this can be rebased we'd be happy to review it - I'm still using the gist hack but it seems it might be generally useful to enough folks.

cheeseplus:
Given the age and conflicts on this one, closing. We've totally though about implementing our own proxy but given the need for wide cross platform support this is difficult to do without a bunch more code. Thanks for the contribution regardless!

PR #197, opened 2013-09-11
First stage for Puppet support [Rebased #83]

josephholsten:
just rebased #83 onto master. Passes tests now.

PR #291, opened 2013-12-13
[WIP] Puppet apply provisioner

ChrisLundquist:
After looking at https://github.com/test-kitchen/test-kitchen/pull/83 and https://github.com/test-kitchen/test-kitchen/pull/197

I wanted to take a stab at a lower impact / more contained provisioner. After writing a [shell provisioner](https://github.com/test-kitchen/test-kitchen/pull/286) it seemed like it could be done.

Currently this only sets up puppet 2.7 on ubuntu. It seems like moving to an omnibus installer would fix that up.

This works with hiera, but expects a `hiera.yaml` file somewhere in the directory tree. I wasn't sure if moving that to a template would be better.

The provisioner stanza in my `.kitchen.yml` looks like this
```yaml
provisioner:
  name: puppet_apply
  manifests_path: /absolute/path/to/puppet/manifests
  modules_path: /absolute/path/to/puppet/modules
  hiera_data_path: /absolute/path/to/puppet/puppet/hieradata
```

`manifests_path` and `modules_path` first look in the currently directory tree, but I'm testing this on another code base so it is a little wonky. Note, you can't seem to use `~/path/puppet`. If you do, nothing gets copied.

/cc @fnichol @jfryman
Do either of you know a good way to omnibus install puppet to fix this up?
What are your thought? 
Are there use cases that I have overlooked? 
What can we cleanup to make it more idiomatic?


cheeseplus:
Is this needed anymore with https://github.com/neillturner/kitchen-puppet ?

ChrisLundquist:
two years is a long time.

cheeseplus:
Totally, we've got some new maintainers so we're trying to clean house as part of the process. Apologies that it took so long for us to get back to you.

ChrisLundquist:
:heart: I might have done some of what I needed in https://github.com/neillturner/kitchen-puppet/pull/1
Looks like it has become more popular since then.

PR #300, opened 2013-12-20
SSH ENV as hash

ChrisLundquist:
This is the lowest impact way I see to get a better way for managing the remote env.

Sadly, openssh won't let us set the env in the channel like net/ssh warns, so we do have to prepend the variable to the command.

To me, passing env vars prepended as a string is an implementation detail, and having the interface of taking a hash seems nicer. This way, people could even do crazy thing like pass `ENV` from ruby to the remote.

[Here is how I tested this](https://gist.github.com/ChrisLundquist/8051513)

As you can see, setting the env in channel produces our error messages.

The variables are still there though, because we join and prepend the hash in front of the command.

I left the broken ssh variable in channel method there for now mostly as a reminder for people of why we have to prepend it. We can remove it, but I didn't want any one else to waste time duplicated the failed experiment.

### DIFF
--- a/lib/kitchen/driver/ssh_base.rb
+++ b/lib/kitchen/driver/ssh_base.rb
@@ -96,18 +96,17 @@ module Kitchen
         [combined[:hostname], combined[:username], opts]
       end
 
-      def env_cmd(cmd)
-        env = "env"
-        env << " http_proxy=#{config[:http_proxy]}"   if config[:http_proxy]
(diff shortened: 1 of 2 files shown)


cheeseplus:
This will need to be rebased given the number of changes to SSH and transport but we are still open to reviewing.

ChrisLundquist:
Two years is a long time. I guess I was hoping for feedback sooner. I don't think I can push this forward.

PR #311, opened 2014-01-05
Run an arbitrary command before provisioning

coderanger:
It would be nice if I could run a command (in this case `apt-get install curl`) before installing Chef. This would also help for provisioning OSes not in omnitruck.


timothy-spencer:
Given that #329 seems to not be going anywhere, is this something that could land in the spirit of "don't let best be the enemy of good"?  I am suffering because of the lack of this.

someara:
For what it's worth... kitchen-dokken lets you do this with `intermediate_instructions`

timothy-spencer:
Sadly, we use kitchen-ec2.  I have certs for a proxy that I need to trust before anything can work.  :-(

PR #336, opened 2014-01-21
Added configurable 'berksfile_path' and 'cheffile_path' options

cstewart87:
Allows users to specify which berksfile/cheffile to use for test kitchen.  Defaults to original values.  This has been helpful when using a chef-repo and each cookbook has its own .kitchen.yml.

### DIFF
--- a/lib/kitchen/provisioner/chef_base.rb
+++ b/lib/kitchen/provisioner/chef_base.rb
@@ -46,6 +46,16 @@ module Kitchen
       end
       expand_path_for :data_path
 
+      default_config :berksfile_path do |provisioner|
+        provisioner.calculate_path("berksfile")
+      end
+      expand_path_for :berksfile_path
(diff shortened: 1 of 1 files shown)


CpuID:
*bump* trying to solve the issue this sorts out at the moment :)

cheeseplus:
Given the concerns listed above and a feeling that Berksfiles/Cheffiles themselves with local paths solves the bulk of the issue I'm going to close the issue. TK is cookbook centric as is Berkshelf and unless we can address the edge cases / make it fool proof I feel like we're getting outside of our core competency.

gled-rs:
Implemented and updated the original code in this fork if someone needs the functionality:

https://github.com/gled-rs/test-kitchen-ez/commit/514a43a483d227170a042c16f7b5d48f5b2e7e49

PR #364, opened 2014-02-18
Global (aka 'all') suite support

erikh:
This adds a special directory -- `test/integration/all` -- that operates as a suite that is the base for all other suites. This allows cookbooks to have specialized suites for recipes and still normalize shared functionality between suites by putting your busser tests and helpers in the `test/integration/all` path just like you would any other suite.

When `test/integration/all` is found, the files are copied from there before copying from anywhere else. This way, a suite may override the given test with a more useful test for that suite, as it will be copied in later. Think class inheritance.

I have also added a few (small) support methods to make some of the more gory bits of path concatenation simpler.

https://github.com/gollector/chef-gollector/tree/master/test/integration is an example of how the tree looks after exploiting this patch. Note that there are no backwards compat problems that I have noticed.

There are no tests AFAICT for the busser functionality touched, or documentation best I can tell. Please inform me if there is a necessity for either.

Thanks!

### DIFF
--- a/lib/kitchen/busser.rb
+++ b/lib/kitchen/busser.rb
@@ -121,18 +121,32 @@ module Kitchen
     # @return [String] a command string to transfer all suite test files, or
     #   nil if no work needs to be performed.
     def sync_cmd
-      @sync_cmd ||= if local_suite_files.empty?
+      @sync_cmd ||= if local_suite_files.empty? && all_suite_files.empty?
         nil
       else
         sync_cmd  = []
         sync_cmd << busser_setup_env
         sync_cmd << "#{sudo}#{config[:busser_bin]} suite cleanup"
-        sync_cmd << "#{local_suite_files.map { |f| stream_file(f, remote_file(f, config[:suite_name])) }.join("; ")}"
-        sync_cmd << "#{helper_files.map { |f| stream_file(f, remote_file(f, "helpers")) }.join("; ")}"
+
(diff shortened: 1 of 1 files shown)


cheeseplus:
Given the age and issues with implementation we're going to close this for now. Something explicit with regard to setting a directory seems like a much better approach as we want to remove as much magic (mystery) as possible from the user experience.

grubernaut:
Just ran into this issue, I was able to use busser to transfer default serverspec tests via the `helpers` directory, without having to explicitly require the tests.

```
.
|-- fixtures
`-- integration
    |-- helpers
    |   `-- serverspec
    |       |-- default
    |       |   `-- default_spec.rb
    |       `-- spec_helper.rb
    `-- super-test-suite
        `-- serverspec
            `-- super_suite_spec.rb
```

Where,

`super_suite_spec.rb`:

```
require 'spec_helper'
...
# custom suite-specific tests
```

`spec_helper.rb`:

```
require 'serverspec'

set :backend, :exec
```

and `default_spec.rb` are my shared tests between suites.

I was previously able to use a shared data directory as @sethvargo mentioned, but the files are populated onto the specified platform during the provisioning step, not the verifier. This left a pretty bad user-experience, as updating any incorrectly written tests forced another `converge` run, instead of the significantly shorter `verify` run. 

Not to rehash a closed issue, just documenting here in case other users find this useful. <3

PR #393, opened 2014-03-24
Allow configuration for cookbooks and site-cookbooks paths (array of cookbook paths)

jpb:
In our chef-repo we want to be able to keep all our cookbooks under `site-cookbooks/` and have one `Cheffile` to populate the `cookbooks` directory, but have a separate test-kitchen setup for each cookbook. Allowing the cookbooks and site-cookbooks path allows us to achieve this.

For example, my configuration would look something like this:

```
# cookbooks/something/.kitchen.yml
provisioner:
  ...
  cookbooks_path: <%= File.join(ENV['CHEF_REPO'], 'cookbooks') %>
  site_cookbooks_path: <%= File.join(ENV['CHEF_REPO'], 'site-cookbooks') %>
```

### DIFF
--- a/lib/kitchen/data_munger.rb
+++ b/lib/kitchen/data_munger.rb
@@ -86,7 +86,7 @@ module Kitchen
     def convert_legacy_chef_paths_format!
       data.fetch(:suites, []).each do |suite|
         %w{data data_bags encrypted_data_bag_secret_key
-          environments nodes roles}.each do |key|
+          environments nodes roles cookbooks site_cookbooks}.each do |key|
             move_chef_data_to_provisioner_at!(suite, "#{key}_path".to_sym)
         end
       end
--- a/lib/kitchen/provisioner/chef_base.rb
+++ b/lib/kitchen/provisioner/chef_base.rb
@@ -310,11 +310,19 @@ module Kitchen
       end
 
       def cookbooks_dir
-        config[:cookbooks_path]
+        if config[:cookbooks_path]
+          File.expand_path(config[:cookbooks_path])
+        else
+          File.join(config[:kitchen_root], "cookbooks")
+        end
       end
 
       def site_cookbooks_dir
-        config[:site_cookbooks_path]
+        if config[:site_cookbooks_path]
+          File.expand_path(config[:site_cookbooks_path])
+        else
+          File.join(config[:kitchen_root], "site-cookbooks")
+        end
       end
 
       def data_bags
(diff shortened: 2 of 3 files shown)


cheeseplus:
If this can be rebased and reworked to be generally flexible as @fnichol mentions feel free - if not let us know and we can put an issue down to be prioritized against the backlog.

Cheers!

luckymike:
@cheeseplus @portertech & I were working with @jpb when he requested this. Given the age of this and what I know about that project, I wouldn't guess that revisiting it will be a priority. We had been under the impression that this was a handoff to the Test Kitchen team :) It's probably best to add it to your backlog, if you want to support it.

cheeseplus:
@luckymike well I/we're the test-kitchen team and I don't see anything really in the way of adding it aside from rebasing it to current TK

That said, Cheffile / Librarian has faded in the community and most people would accomplish this with Berkshelf which also sidesteps the clear winner issue (berks'll just blow up if there are two of the same name). That is to say I'm not against the feature but I do feel like most folks have just rubbed some Berkshelf on it as a solution.

luckymike:
What I meant to suggest is that this has probably been abandoned by the original requester, as it's been 2 years, so asking for a rebase probably won't go anywhere if you do want to include it. :) 

I hear you on Cheffile abandonment, but Berks still fails to solve the repository needs described here. Personally, I'd love to see energy go into accepting #647 to provide a better path to addressing this.

Jimflip:
So is there a simple way of telling test-kitchen the location of my cookbooks?

mwrock:
@Jimflip the easiest way to accomplish this is via a berksfile. Lets say you had 3 local cookbooks youi wanted to reference, your `berksfile` might look like:
```
source 'https://supermarket.chef.io'

group :integration do
  cookbook 'cookbook-A', :path => '../cookbooks/cookbook-A'
  cookbook 'cookbook-B', :path => '../cookbooks/cookbook-B'
  cookbook 'cookbook-C', :path => '../cookbooks/cookbook-B'
end
```

angelichorsey:
@mwrock What if I don't want to use a Berksfile?

So I thought the issue was with chef_zero not interpreting the cookbooks_path values correctly, but it seems that chef_solo cannot either. If this functionality isn't available then perhaps someone could tell the folks at Chef to remove it from their documentation: https://docs.chef.io/config_yml_kitchen.html#multiple-cookbooks

Incredibly frustrating working through this nonsense.

cheeseplus:
@angelichorsey if you don't want to use a Berksfile you don't have to - however many years after this ticket was opened, Librarian, has effectively fallen by the wayside. Your comment doesn't seem to negate the documentation either. Multiple cookbook paths are still allowed and fwiw both @mwrock and I happen to be employed at Chef so not clear what is wrong here.

PR #424, opened 2014-05-02
Set compression flag for SSH session.

sawanoboly:
Now, compression flag of SSH is dependent on the user configuration such as ssh_config.
It may take some time to transfer the file.

Example: About 100MB file transfer.

```
# case compression=true
real	0m16.905s
user	0m7.740s
sys	0m0.505s

# case compression=false
real	0m46.755s
user	0m3.001s
sys	0m0.401s
```

### DIFF
--- a/lib/kitchen/driver/ssh_base.rb
+++ b/lib/kitchen/driver/ssh_base.rb
@@ -98,6 +98,7 @@ module Kitchen
         opts[:port] = combined[:port] if combined[:port]
         opts[:keys] = Array(combined[:ssh_key]) if combined[:ssh_key]
         opts[:logger] = logger
+        opts[:compression] = true
 
         [combined[:hostname], combined[:username], opts]
       end


sethvargo:
How backwards compatible is this? Do we need a certain version of SSH

sawanoboly:
Sorry for long delay.
Its feature has been supported since 2007.  At the time, Maybe the net/ssh version is 1.99.

https://github.com/net-ssh/net-ssh/blame/441d07882645114b5acc764d2b0b83930c6cacd7/lib/net/ssh/transport/state.rb#L23

sethvargo:
:+1:

fnichol:
Okay, now that we have `Transport::SSH` this feels like a much easier thing to reason about. I'll take a look at what this looks like, thanks for the update!

fnichol:
Let do it! Updated implementation in #675

sawanoboly:
Thanks :+1:

lamont-granquist:
@sawanoboly heads up that i just flipped your default to turning compression off after benchmarking the file upload from travis using kitchen-ec2 and finding that compression made the time worse. 

its likely that this is the more common use case where we're cpu limited and transferring lots of small files -- if someone has a 100MB compressible file checked into a cookbook that they're uploading with test-kitchen that seems like a bad use case and is clearly on the 20% side of the 80/20 rule, so the you'll have to set compression to true in your .kitchen.yml file.

PR #438, opened 2014-05-14
gem_src: Add gem_source configuration option.

MarkGibbons:
Allow a gem server to be specified for the busser gem install.

When running test-kitchen in an environment without access to rubygems.org I need to be able to specify the gem mirror where busser and related gems may be found. This request adds a gem_source option to the busier config in .kitchen.yml so that the source may be specified.  The config option is appended to the busser gem install command as "--source <gem source>.

Thanks,
  Mark


senecav:
I support the pull request as well. We use a Nexus server at my place of work to store the gems locally as we have limited access to the internet from certain environments. It would be great to be able to specify a local source for the gem install without having to create a gem_source cookbook as I echo kongslund's comments about this changing the parameters of the testing.

codecrunchers:
Facing the same issue - this would be a great feature

PR #442, opened 2014-05-19
The bootstrap.sh script must be made executable on the remote machine.

ringods:
This change makes the shell provisioner work with Windows as a host machine too.

The old implementation makes the assumption that one is running on a Unix host, making the bootstrap script executable on the host machine and then scp-ing it to the guest, meanwhile maintaining the file permissions.

I changed this so the script is generated in the host sandbox folder, then transferring it to the (Unix!) guest and making it executable there.

### DIFF
--- a/spec/kitchen/provisioner/shell_spec.rb
+++ b/spec/kitchen/provisioner/shell_spec.rb
@@ -210,7 +210,6 @@ describe Kitchen::Provisioner::Shell do
           provisioner.create_sandbox
 
           sandbox_path("my_script").file?.must_equal true
-          sandbox_path("my_script").executable?.must_equal true
           IO.read(sandbox_path("my_script")).must_equal "gonuts"
         end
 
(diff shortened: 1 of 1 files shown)


cheeseplus:
Does this condition still exist and if so can we solve it without chmod?

b00ga:
I happened to be looking through PRs for test kitchen and saw this one. This PR addresses the issue I recently reported in issue #931 

The only way I was thinking that this could be solved without chmod is to explicitly call scripts with their interpreter and not rely on the shebang line (#!/bin/sh). So instead of calling "sudo -E /tmp/kitchen/bootstrap.sh" and relying on bootstrap.sh to be chmodded 755/+x, call it as "sudo -E sh /tmp/kitchen/bootstrap.sh". I'd suggest in this scenario to make 'sh' the default, but make it a config option. Here the user can select any interpreter on the target guest, a different shell (bash, ksh, zsh on Linux, cmd.exe or maybe even bash on Windows) or possibly even any other interpreted language's interpreter (Ruby, Python, Perl, PHP, etc, specifying the bootstrap script as bootstrap.rb/bootstrap.py/etc).

-Shawn

cheeseplus:
Going to fix this one way or another so linking it to the upcoming milestone even if we close it.

tas50:
This same issue was solved with #1381

PR #447, opened 2014-05-27
add puppet to the list of ignored names to allow for a per suite puppet ...

ehaselwanter:
...test configuration

otherwise it fails with at the busser install step:

```
sudo -E ${gem_bindir}/busser setup; sudo -E /tmp/busser/bin/busser plugin install busser-puppet busser-serverspec']
```

I created a workaround with a fake gem https://rubygems.org/gems/busser-puppet
And I think this should be somehow configurable, but can't think of how ...

### DIFF
--- a/lib/kitchen/busser.rb
+++ b/lib/kitchen/busser.rb
@@ -224,7 +224,7 @@ module Kitchen
     end
 
     def non_suite_dirs
-      %w{data data_bags environments nodes roles}
+      %w{data data_bags environments nodes roles puppet}
     end
 
     def busser_setup_env


neillturner:
i've prefer to remove the old inline kitchen puppet code and stick with the separate gem as it allows faster modification and improvement. I actually think if you pulled out the chef provisioner from the main code it would also allow chef provisioner to be modified an improved faster but that your decision. -)

cheeseplus:
I am with @neillturner on this one and as such I'm going to close this as we're planning on abstracting out the chef provisioner as suggested.

Issue #439, opened 2014-05-14
Documentation on kitchen functions


dgarstang:
Would it possible to provide some documentation on what all the kitchen commands are and specifically what they do? i.e. 'converge', 'verify', 'test' etc. I can't find this anywhere including the getting started guide.

PR #450, opened 2014-05-28
Help update

MarkGibbons:
Modified the output of the kitchen help commands for create, converge, setup, verify, destroy and test.  The text in the messages is what I'm using to explain how test kitchen works to test-kitchen beginners.  This update may help cover some of the requests raised in issue #439.

### DIFF
--- a/lib/kitchen/cli.rb
+++ b/lib/kitchen/cli.rb
@@ -95,11 +95,23 @@ module Kitchen
       perform("diagnose", "diagnose", args, :loader => @loader)
     end
 
-    [:create, :converge, :setup, :verify, :destroy].each do |action|
+    {
+      create:   'Change instance state to create. Start one or more instances',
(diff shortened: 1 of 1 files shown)


paulbrianstephenson:
@sethvargo - Apologies for bringing up this old issue again, however, I'm a new test-kitchen user and I've also struggled to 'easily' understand the difference between 'kitchen converge' and 'kitchen verify'.

However, I found @RoboticCheese 's description above to be very helpful and concise:

"converge" - Run your (Chef or whatever) provisioner
"verify" - Run the tests set up in test/integration/#{SUITE_NAME}/#{TEST_FRAMEWORK_NAME}
"test" - Do a full test cycle: destroy any already-existing platform, complete every action on it (including a converge and a verify) and destroy it at the end.

I think 'kitchen help' would be more user friendly if it used similar descriptions. I don't think saying "Change instance state to converge", "Change instance state to verify", etc... adds any value.

Just my 2c. Loving using the tool though!

PR #451, opened 2014-05-29
Adding Node Support to Test Kitchen/.kitchen.yml

doubt72:
This PR adds node support to TK and allows nodes to be specified in the .kitchen.yml; these changes are needed to support testing on multiple nodes (or something like it would be).  The current kitchen-metal requires these changes to function properly (see https://github.com/doubt72/kitchen-metal).

This is a fairly big change, so I'd expect peeps to have some comments/suggestions here before this is even considered for merging; points of contact for this are me (though my priority is another project at the moment) or @jkeisier (who is still working on Metal actively at this moment).


cheeseplus:
Is this still generally useful even with kitchen-metal not being actively maintained?

hh:
@cheeseplus is there a recommended approach to using test-kitchen and chef-provisioning?

cheeseplus:
Given the branch conflicts and state of kitchen-metal I'm closing this one. Would be open for a reimplementation.

PR #453, opened 2014-06-01
Exit the loop once the command exits.

coderanger:
This matters if there are multiple channels active at once.


cheeseplus:
Closing in favor of https://github.com/test-kitchen/test-kitchen/pull/724

PR #426, opened 2014-05-03
Add SSH ProxyCommand functionality

bdclark:
Similar to vagrant's `ssh_info[:proxy_command]`, this adds a `:proxy_command` argument for SSH where an SSH proxy is required for a driver to communicate with remote instance(s).

I have tested (and am currently using) this with vagrant driver and vagrant-libvirt plugin where hypervisor is remote and instances live on hypervisor's internal private network.  Another (vagrant-specific) use-case is when `config.ssh.proxy_command` is placed in a user's Vagrantfile (or custom Vagrantfile ERB in kitchen-vagrant's case) to access remote instances when policy or network configuration requires it.

PR #463, opened 2014-06-16
Add SOCKS support to test_ssh

berniedurfee:
This pull request adds SOCKS support to Test Kitchen, so that you can spin up EC2 instances from behind a firewall.

This is only a 'partial' pull request, it'll need Pull Request #426 to be included, which adds the ability to add a ProxyCommand argument to the SSH command. That pull request won't work with EC2, unless the changes in this pull request are included as well.

This does introduce a dependency on 'socksify'.

Please let me know if this is a viable pull request. We, along with many others in big enterprise land live behind firewalls. Without this, Test Kitchen is not really all that usable to us.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,13 @@ subcommands and their usage:
 $ kitchen help test
 ```
 
+## Using Test Kitchen from behind a SOCKS proxy
+
+You can specify the following parameters to use Test Kitchen with remote hosts (e.g. EC2) from behind a firewall:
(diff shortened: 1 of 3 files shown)


cheeseplus:
Could this be rebased?

berniedurfee:
I deleted the original fork, so I don't think it can be done?

I created a new PR #916 to replace this against HEAD.

PR #466, opened 2014-06-26
Add compression before uploading files

MurgaNikolay:
driver: kitchen-ec2

Before:
-----> Kitchen is finished. (11m53.12s)

After:
-----> Kitchen is finished. (0m41.31s)


jcii:
+1 for this...what can I/we do to move it forward?

mwrock:
There has been quite a bit of churn in Test-Kitchen since this was submitted. One of the key changes affecting this PR is the decoupling of transport from driver. This PR overhauls upload in `SSHBase`, but the responsibility of upload has been moved to the transport.

What needs to be done to make this comply with today's modern Test-Kitchen is move the new upload logic to `Kitchen::Transport::SSH`. I think that would be fairly straight forward to do.

While as @sethvargo points out above that this is linux centric, the winrm implementation does currently leverage compression so this can remain completely an SSH concern.

cheeseplus:
Going to close this one for a a few reasons:

1) There are now external implementations that do this effectively: https://github.com/coderanger/kitchen-sync
2) We're trying to keep core lean and tight and want to keep these solutions as plugins - they may eventually make into core (think chef resources)
3) This requires a substantial rebase to comply with the Transport abstraction work
3) SSH is already gzipping so where this actually saves time is round trips, which is still important

Thanks for the PR and the effort, we really appreciate it - we're definitely going to be encouraging the usage of some of the improved transport plugins like `kitchen-sync` but still will have them be optional in a purely TK context

lamont-granquist:
FWIW most of the time spend in uploading files seems to be CPU grinding in a single core, and I've turned compression /off/ in ssh by default -- due to testing that revealed this made the process fast.  I also made test-kitchen use net-scp async and do multiple uploads in parallel, but this did not make any difference in performance of uploads, again indicating that maxing out a cpu core is the bottleneck.

I'm investigating the best compression/hmac/kex algorithms for performance in order to determine how to speed up file transfers.  I suspect that the approach in this PR was never the correct path to go down though.

PR #484, opened 2014-07-30
Add ability to download and install chef from GitHub.

mcquin:
This replaces https://github.com/test-kitchen/test-kitchen/pull/473, with some minor changes:

* Uses the tag `:github`, instead of `:chef_git_url`, and set `:github "<user>/<repo>"`
* Requires `:branch` key if `:github` key is used. `:branch` can be set to a branch name or commit sha.
* Overrides `:require_chef_omnibus` to `true` when downloading GitHub source code, if set to `false` or `nil`
* Will fail immediately if the source code cannot be downloaded from the generated url or if the chef gem cannot be built.

Works on Ubuntu 12.04 on both vagrant/virtualbox and ec2. Works on CentOS on vagrant/virtualbox and I'm pretty sure ec2, too.

@fnichol, @sethvargo, @sersut, @lamont-granquist, @adamedx, @danielsdeleo

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,3 +1,7 @@
 Style/FileName:
   Exclude:
     - 'support/**/*'
+
+Style/ClassLength:
+  Exclude:
+    - 'lib/kitchen/provisioner/chef_base.rb'
--- a/lib/kitchen/provisioner/chef_base.rb
+++ b/lib/kitchen/provisioner/chef_base.rb
@@ -141,13 +141,45 @@ module Kitchen
         ))
       end
 
-      # Generates the shell code to conditionally install a Chef Omnibus
-      # package onto an instance.
+      # Generates the shell code to conditionally install Chef on the system.
       #
       # @return [String] shell code
       # @api private
       def chef_install_function
-        version = config[:require_chef_omnibus].to_s.downcase
+        install_command = install_chef_omnibus
+
+        if repo = config[:github]
+          chef_src_dir = "/tmp/src/chef"
+          chef_install_dir = "/opt/chef"
+
+          unless branch = config[:branch]
+            raise(UserError,
+              "The #{ENV['KITCHEN_YAML'] || '.kitchen.yml'} defines a 'github' key" \
+              " but does not define a 'branch' key." \
+              " Please add: `branch <git_sha>` OR `branch <branch_name>`" \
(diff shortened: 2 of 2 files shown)


tyler-ball:
@fnichol and I are trying to start closing some pending Test Kitchen PRs, and I picked this one first because I was familiar with it.

The current code doesn't handle multi platform support very well - would the shell code work on FreeBSD or Windows?  Because we don't think it would, it shouldn't go inside TK core until we get better testing around these cases.

But rather than enhance this PR to support all TK platforms we think this should be separated into its own TK Provisioner.  The use case for this is currently very Chef specific.  So we propose adding a new Provisioner under https://github.com/opscode/chef/tree/master/kitchen-tests that has this logic in it.  We can do the work to migrate this PR to the Chef repo.

As we start adding more functionality to TK, the pinned branch in https://github.com/opscode/chef/blob/master/kitchen-tests/.kitchen.yml will get more out of sync, so I think it is worth doing this work.  There isn't a pressing need, but it is an opportunity for @fnichol and I to get some pair programming time on something we're we can share knowledge.

@mcquin let me know if you have any issue with my proposed plan!

lamont-granquist:
I added OpenSUSE support in a commit here:

https://github.com/lamont-granquist/test-kitchen/tree/lcg/install_chef_from_github

which doesn't seem to have made it into this PR (i thought i did a PR against @mcquin's branch but it doesn't seem to have made it transitively into this PR).

lamont-granquist:
And writing a new provisioner sounds entirely sensible to me.

mcquin:
@tyler-ball I think adding a provisioner is a really good idea. I think that the people who are going to want this feature are those who are developing Chef. It doesn't entirely make sense to have this in the chef-solo/zero provisioners kitchen provides.

I'm very excited about this! :D

fnichol:
To update the group here, @jdmundrawala has been working on [jdmundrawala/kitchen-appbundle-updater](https://github.com/jdmundrawala/kitchen-appbundle-updater) which aims to solve this problem via a custom Provisioner called `chef_github`. Now that we're making solid progress here, I'm going to close this one out and we'll keep the work going in [jdmundrawala/kitchen-appbundle-updater](https://github.com/jdmundrawala/kitchen-appbundle-updater).

Thanks all!

mcquin:
Excellent! Looking forward to it, @jdmundrawala. Thanks for the update, @fnichol.

PR #490, opened 2014-08-14
Generate an Ohai hint file for the current driver

schisamo:
This will ensure Ohai loads all cloud-provider specific attribute data during the Chef run.

/cc @test-kitchen/owners


Jimflip:
would be good to see this in next release so I can remove some ugly code doing the same thing from my cookbook

cheeseplus:
Can we get this rebased or should re repin the issue?

palfrey:
Any news on this? Anything I can do to help?

cheeseplus:
@palfrey this already exists on the drivers themselves, in the case of `kitchen-ec2` so not sure this is something we need any longer. Also, looking at drivers this would hint non-existent things for many.

cheeseplus:
Given the enhancements to Ohai itself, this is no longer even necessary or advised. Closing.

PR #501, opened 2014-09-18
Kernel.exec ~ expansion bug

davent:
When setting an SSH key in .kitchen.yml with a path starting with a ~, Kernel.exec will not expand ~ to $HOME when passed as an element in an Array.


cheeseplus:
Could this be rebased?

cheeseplus:
Given the age and lack of response, closing this one - I also am pretty sure this wouldn't work on Windows.

PR #516, opened 2014-10-07
Move Shell provisioner data to provisioner as well

paracycle:
Each test suite should be able to specify its own `script` and
`data_path` if the provisioner in use is the Shell provisioner.

### DIFF
--- a/lib/kitchen/data_munger.rb
+++ b/lib/kitchen/data_munger.rb
@@ -44,7 +44,7 @@ module Kitchen
       convert_legacy_require_chef_omnibus_format!
       convert_legacy_busser_format!
       convert_legacy_driver_http_proxy_format!
-      move_chef_data_to_provisioner!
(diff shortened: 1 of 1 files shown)


cheeseplus:
Could this be rebased? We're planning on moving Chef out of core like other provisioners so this lines u well either way.

paracycle:
That's great news. I did a quick mechanical rebase of the original commit but didn't have time to check if anything else needs updating in the new code as well. Please let me know how you want to proceed.

cheeseplus:
I kicked the build to see if this will work but we will need it rebased again to get it in but I'd still love to see this in.

paracycle:
@cheeseplus I rebased the commit on the latest master, however, it seems there is a problem with `nio4r` dependency on build systems and that seems to be failing (some of the) test runs.

PR #520, opened 2014-10-10
Add cmd_prefix ssh driver option to allow windows-via-ssh to work

tyrken:
A slightly less hacky form of the workaround at http://aespinosa.github.io/blog/2014-05-08-workarounds-to-converge-test-kitchen-on-windows.html, to allow SSH control of windows guests

### DIFF
--- a/lib/kitchen/driver/ssh_base.rb
+++ b/lib/kitchen/driver/ssh_base.rb
@@ -147,6 +147,7 @@ module Kitchen
       def run_remote(command, connection)
         return if command.nil?
 
+        command = "#{config[:cmd_prefix]} #{command}" if config[:cmd_prefix]
         connection.exec(env_cmd(command))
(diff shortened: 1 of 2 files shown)


fnichol:
@tyrken Hi there! If you're still working with Test Kitchen and Windows together I hopefully have some good news and less than good news. The good news is: we have 1.4.0.beta.* releases that now work natively with Windows Instances over WinRM out of the box. The less than good news? The SSH communication may not be fully fleshed out when communicating with a Windows Instance.

I'm hoping that the WinRM route is more desirable anyway and seeing as how the code underneath has changed considerably (we now have a plugin type of Transport for SSH and WinRM), I'm going to close this out.

However, if I haven't addressed what you're trying to do, could you reply here, reopen, or file a new issue? Thank you so much!!!

PR #530, opened 2014-10-23
Stop uploading empty directories

### LINKED ISSUE #429: Filtering non-cookbook files leave empty directories that are still scp-ed
We use test-kitchen & kitchen-ec2 as part of our CI testing on travis. By default travis will bundle in to `vendor/bundle/` in the cloned repository.

We were getting builds that would appear to hang randomly while transferring files for the cookbook being tested. I SSH-ed in to the box while it was transferring files, ran strace against the scp process, and there I saw that it was copying bunch of "files" in the installed bundle directory. Investigating further, I found that they were actually all empty directories.

[Looking at the code](https://github.com/test-kitchen/test-kitchen/blob/master/lib/kitchen/provisioner/chef_base.rb#L273) it looks like all of the resolved cookbooks are copied to a temp directory and then ignorable files are removed. However, directories that these files were in are left behind, due to [filtering non-files from the removal list](https://github.com/test-kitchen/test-kitchen/blob/master/lib/kitchen/provisioner/chef_base.rb#L280).

Uploading files to the instance happens in [`Kitchen::SSH`](https://github.com/test-kitchen/test-kitchen/blob/master/lib/kitchen/ssh.rb#L70) using `Net::SCP`. This will happily upload those "empty" directories that we don't care about anymore. It also doesn't yield to the progress block for these directories, so it is not obvious this is happening.

The issue is caused by the ~2000 directories that get created by a pretty reasonable set of installed gems, namely berkshelf, foodcritic, chefspec, & test kitchen. For now I've worked around this issue by having bundle install to another location, outside of the work directory.

I'm not sure what the best way to "fix" this would be, so this is just a bug report for now. I think it would make sense to figure out a way to avoid having these directories end up in the upload at all, whether that means a second round of deleting or switching to a white-list copy.

It appears [some discussion happened](https://github.com/test-kitchen/test-kitchen/issues/35) a while back about a better way to upload cookbooks, which might also solve the slowness issue even if these ignorable directories were still uploaded

whiteley:
Fixes #429

### DIFF
--- a/lib/kitchen/provisioner/chef_base.rb
+++ b/lib/kitchen/provisioner/chef_base.rb
@@ -292,6 +292,8 @@ module Kitchen
       def filter_only_cookbook_files
         info("Removing non-cookbook files before transfer")
         FileUtils.rm(all_files_in_cookbooks - only_cookbook_files)
+        Dir.glob(File.join(tmpbooks_dir, "**/"), File::FNM_PATHNAME).
+          reverse_each { |fn| FileUtils.rmdir(fn) if Dir.entries(fn).size == 2 }
       end
 
       # Generates a list of all files in the cookbooks directory in the


whiteley:
`kitchen-sync` is a great plugin for speeding up transfers, but why are we compressing or pipelining empty directories when this would help all upload methods?

sethvargo:
Thanks!

L2G:
This change is causing new errors for me. See the example below. The path ending with `test/integration/joyent/bats` is one of the directories that was pruned, but the error suggests that something somewhere else in the code didn't get the message that it was pruned.

```
-----> Starting Kitchen (v1.3.2.dev)
-----> Converging <prod-redis28-master-centos-66>...
       Preparing files for transfer
       Preparing dna.json
       Resolving cookbook dependencies with Berkshelf 3.1.5...
       Preparing data_bags
       Preparing environments
       Preparing roles
       Preparing validation.pem
       Preparing client.rb
-----> Chef Omnibus installation detected (install only if missing)
       Transferring files to <prod-redis28-master-centos-66>
>>>>>> ------Exception-------
>>>>>> Message: Failed to complete #converge action: [No such file or directory @ rb_file_s_stat - /var/folders/b_/y337by3540l885cbfn8d85ww0000gp/T/prod-redis28-master-centos-66-sandbox-20150123-98432-1736qnt/cookbooks/itoys-tsung/test/integration/joyent/bats]
>>>>>> ----------------------
```

whiteley:
@L2G very interesting, I've been using this for months with cookbooks containing different types of test files including bats without issue. Obviously, you're seeing a problem though and I'd like to help figure it out. Would you be able to post a listing of the contents of your `test/` directory? Something like:

```console
$ find test -ls                                                                                                                                                                 [18:22:28]
4355099        0 drwxr-xr-x    4 mwhiteley        wheel                 136 Jan 23 18:20 test
4355100        0 drwxr-xr-x    3 mwhiteley        wheel                 102 Jan 23 18:20 test/fixtures
4355101        0 drwxr-xr-x    3 mwhiteley        wheel                 102 Jan 23 18:20 test/fixtures/test-kitchen
4355102        0 drwxr-xr-x    3 mwhiteley        wheel                 102 Jan 23 18:20 test/fixtures/test-kitchen/scripts
4355103        8 -rw-r--r--    1 mwhiteley        wheel                 384 Jan 23 18:20 test/fixtures/test-kitchen/scripts/install-smartos.sh
4355104        0 drwxr-xr-x    3 mwhiteley        wheel                 102 Jan 23 18:20 test/integration
4355105        0 drwxr-xr-x    3 mwhiteley        wheel                 102 Jan 23 18:20 test/integration/default
4355106        0 drwxr-xr-x    3 mwhiteley        wheel                 102 Jan 23 18:20 test/integration/default/bats
4355107        8 -rw-r--r--    1 mwhiteley        wheel                  50 Jan 23 18:20 test/integration/default/bats/install_a_package.bats
```

Thanks!

L2G:
@whiteley: I will see what I can do. I need to make sure I'm not revealing any sensitive info.

Are you interested in the "main" `test/` directory, or the one that is called out in that error message (which is different)?

L2G:
Oh, indeed, this turned up something interesting:

`11418104        8 lrwxr-xr-x    1 larrygilbert     staff                  15 Oct 31 17:40 test/integration/joyent/bats -> ../default/bats`

Maybe both the symlink and its target are being collected by that `Dir.glob`, the target directory is getting removed, and then when it gets to the symlink (which previously looked like a directory), the symlink doesn't point to anything so an error gets thrown.

PR #562, opened 2014-12-11
winrm transport should use a single (or minimal) shell when transferring files. transfer via a zip file to optimize round trips

mwrock:
This can significantly improve file upload performance over winrm. It eliminates a authentication handshake and winrs process creation for each call and compresses the files to a single contiguous block to rudece the number of round trips necessary to transfer the files. winrm limits roundtrips to 8k chunks do optimizing the payload is crucial.


mwrock:
This is green now. There were a ton of style violations that had built up and I ad to move alot around to make them happy. The final one was a complaint about the chef_base class being too long. The obvious abstraction to abstract was the shell. We have quite a few case statements scattered about that branches on `bourne` vs. `powershell`. I had noticed this being awkward and created `BourneShell` and `PowershellShell`. This is an **extremely** rough implementation just to get this green. I'm about to refactor and move them so that they do not hang off of the provisioner. I also need to double check that the code actually works. Hope to wrap this up tonight.

mwrock:
finished the refactor but it broke a ton of tests that I'm fixing up. Looks like pushing it is gonna be more likely to happen over the weekend.

mwrock:
Finished refactor and smoke tested on windows host against both windows 2012 and ubuntu 12 guests. Needs more testing and some test cleanup and api comments.

mwrock:
Looks like this broke ruby 1.9.3 with Module.get_const

lib/kitchen/busser.rb
```diff
@@ -40,23 +56,10 @@ class Busser
     # @option opts [String] :kitchen_root local path to the root of the project
     # @option opts [String] :instance_ruby_bindir path to the directory
     #   containing the Ruby binary on the remote instance
-    # @option opts [TrueClass, FalseClass] :sudo whether or not to invoke
-    #   sudo before commands requiring root access (default: `true`)
```
jdmundrawala: Will removing this break backwards compatibility?

lib/kitchen/busser.rb
```diff
@@ -100,65 +86,30 @@ def diagnose
...
-          #{busser} plugin install #{plugins.join(" ")}
-        CMD
-      else
-        raise "[#{self}] Unsupported shell: #{shell}"
-      end
-      Util.wrap_command(cmd, shell)
+      cmd = "#{busser_setup_env}\n"
+      cmd << shell.busser_setup(config[:ruby_bindir], config[:root_path], gem_install_args)
+      cmd << "\n#{shell.sudo(config[:busser_bin])} plugin install #{plugins.join(" ")}"
+      shell.wrap_command(cmd)
```
jdmundrawala: :+1:

lib/kitchen/provisioner/chef/powershell_shell.rb
```diff
@@ -0,0 +1,117 @@
...
+        # A command for initializing an instance for working with chef
+        # the calling provisioner may have a omnibus root configured by
+        # the user that is different from this module's chef_omnibus_root.
+        #
+        # @param root_path [String] root path to chef omnibus
+        # @return [String] command to run on instance initializing
+        # chef environment
+        def init_command(root_path)
+          cmd = <<-CMD.gsub(/^ {10}/, "")
+            if (-Not (Test-Path #{root_path})) {
```
jdmundrawala: Is it safe to leave the paths unquoted?

lib/kitchen/provisioner/chef/bourne_shell.rb
```diff
@@ -0,0 +1,107 @@
...
+        # the calling provisioner may have a omnibus root configured by
+        # the user that is different from this module's chef_omnibus_root.
+        #
+        # @param root_path [String] root path to chef omnibus
+        # @return [String] command to run on instance initializing
+        # chef environment
+        def init_command(root_path)
+          dirs = %w[cookbooks data data_bags environments roles clients].
+            map { |dir| File.join(root_path, dir) }.join(" ")
+          lines = ["#{sudo("rm")} -rf #{dirs}", "mkdir -p #{root_path}"]
```
jdmundrawala: is it safe to leave the paths unquoted?

lib/kitchen/shell/powershell.rb
```diff
@@ -0,0 +1,75 @@
...
+        CMD
+      end
+
+      # (see Base#add_to_path)
+      def add_to_path(dir)
+        <<-CMD.gsub(/^ {10}/, "")
+          $env:PATH="$env:PATH;#{dir}"
+        CMD
+      end
+
```
jdmundrawala: what happened to the helper file for powershell

lib/kitchen/transport/base.rb
```diff
@@ -124,9 +124,9 @@ def wait_for_connection
       # Returns the desired shell to use.
       # [Idea] Let's see if this help for the Provisioner to chose the right code
       #
-      # @return [String] the desired shell for this transport
+      # @return [Shell] the desired shell for this transport
       def shell
-        config.fetch(:shell, nil)
+        @shell ||= Shell.for_plugin(config.fetch(:shell, "bourne"))
```
jdmundrawala: I think you have a `DEFAULT_SHELL` constant somewhere

lib/kitchen/transport/base.rb
```diff
@@ -156,6 +156,8 @@ def default_port
       def finalize_config!(instance)
         super
         load_needed_dependencies!
+        config[:http_proxy] = instance.driver[:http_proxy] unless config[:http_proxy]
```
jdmundrawala: `||=`?

lib/kitchen/transport/ssh.rb
```diff
@@ -34,6 +34,19 @@ class Ssh < Kitchen::Transport::Base
       default_config :sudo, true
       default_config :shell, "bourne"
```
jdmundrawala: maybe use `DEFAULT_SHELL` here

lib/kitchen/transport/winrm_file_transfer/remote_file.rb
```diff
@@ -0,0 +1,228 @@
...
+          @closed = false
+          @service = service
+          @shell = Shell.new(logger, service)
+          @local_path = local_path
+          @remote_path = full_remote_path(local_path, remote_path)
+          @logger = logger
+          logger.debug("Creating RemoteFile of local '#{local_path}' at '#{@remote_path}'")
+        ensure
+          if !shell.nil?
+            ObjectSpace.define_finalizer(self, self.class.close(shell))
```
jdmundrawala: This looked scary until I saw that this class had both close and self.close. Maybe rename to finalizer?

mwrock:
@jdmundrawala: Thanks a ton for the comments and I'll respond later but I generally agree on all points and yes the paths should be quoted. I have a meeting with our chef "success engineer" today so will likely miss the zoom call.

jdmundrawala:
no problem, i' still going through... still have a few more files to look
at. In general, i think we should merge it, everything seems in a sane
place (but that's is fletcher's call)

jdmundrawala:
Did you fix the issue where writing to stderr was considered an error? Because I'm not seeing that being an issue in this branch

lib/kitchen/provisioner/chef_zero.rb
```diff
@@ -56,47 +55,40 @@ def prepare_command
...
+        add_argument(cmd, :chef_zero_port, "--chef-zero-port %config%")
+        add_argument(cmd, :json_attributes, "--json-attributes #{config[:root_path]}/dna.json")
+        add_argument(cmd, :log_file, "--logfile %config%")
+
+        shell.wrap_command(cmd.join(" "))
       end
 
       private
 
+      def add_argument(line, config_key, arg)
```
tyler-ball: I think this can use some doc because of the `%config%` substitution

lib/kitchen/shell/base.rb
```diff
@@ -0,0 +1,128 @@
...
+      # @return [String] command to run on test instance that
+      # installs and sets up busser
+      def busser_setup(busser_root, gem_install_args) # rubocop:disable Lint/UnusedMethodArgument
+      end
+
+      # Command for setting environment variables on the test instance
+      #
+      # @param key [String] environmnt variable name
+      # @param value [String] value to be assigned to environment variable
+      # @return [String] command for setting an environmrnt variable on
```
tyler-ball: misspelled environment

lib/kitchen/shell/bourne.rb
```diff
@@ -0,0 +1,82 @@
...
+        <<-CMD.gsub(/^ {10}/, "")
+          #{key}="#{value}"
+          export #{key}
+        CMD
+      end
+
+      # (see Base#add_to_path)
+      def add_to_path(dir)
+        <<-CMD.gsub(/^ {10}/, "")
+          export PATH="#{dir}:$PATH"
```
tyler-ball: Has there ever been a use case for appending at the end of the path instead of the beginning?

lib/kitchen/shell/powershell.rb
```diff
@@ -0,0 +1,75 @@
...
+      def set_env(key, value)
+        <<-CMD.gsub(/^ {10}/, "")
+          $env:#{key}="#{value}"
+        CMD
+      end
+
+      # (see Base#add_to_path)
+      def add_to_path(dir)
+        <<-CMD.gsub(/^ {10}/, "")
+          $env:PATH="$env:PATH;#{dir}"
```
tyler-ball: The bourne `add_to_path` was doing a prepend and this is doing an append - is that operating system convention differences?

tyler-ball:
The code looks good - its a big review, nice work!  I'm going to try and use this branch to spin up a Windows host using TK and Vagrant.

:+1:

lib/kitchen/busser.rb
mwrock: I'm assuming you are referring to dropping sudo?

I don't think it does, but that could be open to interpretation. The former code enforced two separate sudo configurations. One here in the busser and another in the provisioner. I think it is more clear to have one and if there are kitchen.yml files out there that specify sudo in a busser config, it will simply be ignored.

My guess is that if sudo was specified for the busser it was also specified for the provisioner. I could be wrong and others may have more insight, but i *think* this is safe.

lib/kitchen/provisioner/chef/bourne_shell.rb
mwrock: I think it is in this case. This is pretty much a copy from https://github.com/test-kitchen/test-kitchen/blob/master/lib/kitchen/provisioner/chef_base.rb#L96

lib/kitchen/provisioner/chef/powershell_shell.rb
mwrock: This on the other hand is NOT safe.

lib/kitchen/provisioner/chef_zero.rb
mwrock: I'm gonna try to do this a better way. It gets the job done and appeases rubocop but is just kinda weird.

lib/kitchen/shell/bourne.rb
mwrock: Honestly I dont know why I prepended this. Switching to append.

lib/kitchen/shell/powershell.rb
mwrock: I dont think so.

lib/kitchen/shell/powershell.rb
mwrock: Still around: https://github.com/mwrock/test-kitchen/blob/zip/lib/kitchen/provisioner/chef/powershell_shell.rb#L77

lib/kitchen/transport/base.rb
mwrock: ah yes. fixing.

lib/kitchen/shell/base.rb
mwrock: thanks. fixing.

lib/kitchen/transport/base.rb
mwrock: nice. fixed

lib/kitchen/transport/ssh.rb
mwrock: right. fixing

lib/kitchen/transport/winrm_file_transfer/remote_file.rb
mwrock: reasonable. fixed

mwrock:
@jdmundrawala I think the stderr issue you are asking about was where the "progress" stream was being handled as a error stream and the "fix" was to effectively turn that stream off which is in https://github.com/mwrock/test-kitchen/blob/zip/lib/kitchen/transport/winrm.rb#L258

jdmundrawala:
@mwrock I was referring to https://github.com/mwrock/test-kitchen/blob/zip/lib/kitchen/transport/winrm.rb#L55

mwrock:
All edits from the review of @jdmundrawala and @tyler-ball is pushed to this PR.

@jdmundrawala I think that logic makes sense since powershell does not return exit codes unless it calls an executable that does. You'd want to consider the presence of a non empty error stream an error condition. This used to cause false positives due to the nature of how winrm translates the many streams of powershell. It would forward the "progress stream" to stderr and thus trigger errors from harmless messages. Since the progress stream is visibly swallowed in a winrm session, it made sense to just turn off this stream on the powershell side to prevent this from happening.

jdmundrawala:
yay!

mwrock:
Sweet!

fnichol:
Thank you to @mwrock, @jdmundrawala, and @tyler-ball for your super-hero work on this!!

I'm in the process of reviewing and rebasing against current master (circa 1.3.1) then will spin off clean up and refactoring work.

jtimberman:
:cake:! :fireworks:! :octopus:!

Issue #339, opened 2014-01-28
Multiple run on the same box


gionn:
Hi,

I was wondering how to implement a scenario where I need to test an upgrade recipe for an application, and it would be great to be able to define different suites (or whatever name you want to give to it) to be ran in a certain order on the same box, with different run_list and attributes for each run.

AFAIK in this moment I can't, since every suite will work on its dedicated instance, and so I can't have multiple runs on the same box with attributes and run_list changed.

Thanks.

Issue #324, opened 2014-01-13
Reboot during Test Kitchen run?


fancyremarker:
I'm wondering whether the following is possible to test via Test Kitchen: in my cookbook, I install a new kernel image, via:

``` ruby
# Upgrade Kernel to 3.8
apt_package 'linux-image-generic-lts-raring' do
  action    :install
  notifies  :run, 'execute[reboot]', :delayed
  not_if    'uname -a | grep -q 3.8.0'
end
```

I then test for the 3.8.0 kernel with the following test:

``` sh
@test "Kernel == 3.8.0" {
  uname -a | grep -q 3.8.0
}
```

When I run `kitchen test`, the test suite clearly doesn't expect to encounter a reboot, and so I get the following backtrace (trimmed):

```
se64: [2014-01-13T17:03:13+00:00] INFO: execute[reboot] ran successfully
se64: 
se64:     - execute reboot
se64: 
se64: [2014-01-13T17:03:13+00:00] INFO: Chef Run complete in 91.119478968 seconds
se64: [2014-01-13T17:03:13+00:00] INFO: Running report handlers
se64: [2014-01-13T17:03:13+00:00] INFO: Report handlers complete
se64: Chef Client finished, 22 resources updated
se64: Finished converging <default-precise64> (5m5.63s).
se64: -----> Setting up <default-precise64>...
se64: [SSH] connection failed, retrying (#<Net::SSH::Disconnect: connection closed by remote host>)
se64: [SSH] connection failed, retrying (#<Net::SSH::Disconnect: connection closed by remote host>)
se64: [SSH] connection failed, terminating (#<Net::SSH::Disconnect: connection closed by remote host>)
se64: Setup failed on instance <default-precise64>.
se64: ------Exception-------
se64: Class: Kitchen::ActionFailed
se64: Message: connection closed by remote host
se64: ---Nested Exception---
se64: Class: Net::SSH::Disconnect
se64: Message: connection closed by remote host
se64: ------Backtrace-------
se64: /Users/frank/.gem/ruby/2.0.0/gems/test-kitchen-1.1.1/lib/kitchen/driver/ssh_base.rb:111:in `rescue in run_remote'
```

Is there any way for this sort of test to be executed in Test Kitchen? FWIW, the recipes run fine (the kernel is updated, the system rebooted) when I run `vagrant up` with the following Vagrantfile:

``` ruby
Vagrant.configure('2') do |config|
[...]

Issue #162, opened 2013-07-06
Allow for "double-converges" on specific test suites


patcon:
Context: http://community.opscode.com/chat/chef#id-391711

Working with a recipe that is particularly prone to failing on the second run, given some complex use of attribute load order.

Thinking it would be nice to be able to converge certain suites twice, before verifying. I believe that [the awesome chef-stage cookbook](https://github.com/ranjib/chef-stage) could probably be used for this as well, but it seemed something that test-kitchen might be able to easily implement, negating the need for specialized `test_mycookbook` support cookbooks.

cc: @ranjib

tjnicholas:
+1

This would be really helpful. Especially in combination with windows support (where reboots are more commonplace). 

I think it would need to support whitelisting of 'failed' convergence runs as sometimes using 'raise' is needed to end the run when there is no point continuing prior to reboot (after upgrading powershell for example).

emiddleton:
@tjnicholas the cleaner way is to set a flag and process the reboot at the end of the run.  You still need to handle killing the connecting service so you don't end up with a race condition.  I use this approach here https://github.com/emiddleton/cookbook-wakame-vdc to upgrade a kernel.

tjnicholas:
@emiddleton Understood. Normally that's what I would do but in Windows land some time there really isn't any point in trying to continue. 

If you're upgrading from powershell 2 to powershell 4 (common) and use powershell from chef then there is little reason to think you'll get through the rest of the converge without an exception being thrown anyway. 

Certainly not ideal, but predictable and repeatable is worth an ugly exit code.

TinaTina:
:+1:

CpuID:
+1, this would be a really nice feature. The approach taken in https://github.com/test-kitchen/test-kitchen/pull/564 seems even more flexible again.

TinaTina:
remember back then we I tried so hard to automate testing for idempotence! :joy:

bdrust:
+1, would be nice.

iiro:
+1 definitely

cheeseplus:
Closing this issue (pruning a bunch of issues asking for the same thing) in favor of https://github.com/test-kitchen/test-kitchen/issues/329 and/or https://github.com/test-kitchen/test-kitchen/pull/564

petere:
I don't think those will actually address this issue.

PR #564, opened 2014-12-16
add support for multiple step converge

emiddleton:
This change allows multiple provisioner runs to be made on a single converge.  This can be used to test migration between cookbooks versions #339, to install a new kernel #324 (using cookbook-reboot-handler patched with emiddleton/cookbook-reboot-handler@a94e40edca2dccf3dcc17334825b6489edde6103) and to make converge more robust #162

Converge steps are defined as an array of step hashes as shown below.  (step hashes support all suite keys except name: and steps:).

``` yaml
platforms:
  - name: centos-6.4

suites:
  - name: mysuite
    steps:
      - run_list:
        - recipe[wakame-vdc::default]
      - run_list:
        - recipe[wakame-vdc::default]
```

Each step result in a new instance with _step_n attached to the suite name i.e

``` shell
# kitchen list

Instance                  Driver   Provisioner  Last Action
mysuite-step-1-centos-64  Vagrant  ChefSolo     <Not Created>
mysuite-step-2-centos-64  Vagrant  ChefSolo     <Not Created>
```

This allows test to be run at each step in the convergence process.

### DIFF
--- a/lib/kitchen/command/list.rb
+++ b/lib/kitchen/command/list.rb
@@ -61,7 +61,7 @@ module Kitchen
         [
           color_pad(instance.name),
           color_pad(instance.driver.name),
-          color_pad(instance.provisioner.name),
+          color_pad(instance.provisioners.last.name),
           format_last_action(instance.last_action)
         ]
       end
--- a/lib/kitchen/config.rb
+++ b/lib/kitchen/config.rb
@@ -120,6 +120,13 @@ module Kitchen
 
     private
 
+    def each_step(suite_name)
+      name, step = * suite_name.split(/_step_/)
+      (1..step.to_i).map do |count|
+        yield suites.find { |suite| suite.name == "#{name}_step_#{count}" }
+      end
+    end
+
     # Builds the filtered list of Instance objects.
     #
     # @return [Array<Instance] an array of Instances
@@ -241,7 +248,7 @@ module Kitchen
         :logger       => new_logger(suite, platform, index),
         :suite        => suite,
         :platform     => platform,
-        :provisioner  => new_provisioner(suite, platform),
+        :provisioners => new_provisioners(suite, platform),
         :state_file   => new_state_file(suite, platform)
       )
     end
@@ -265,8 +272,26 @@ module Kitchen
       )
     end
 
-    # Builds a newly configured Provisioner object, for a given Suite and
-    # Platform.
+    # Builds an array of newly configured Provisioner objects,
+    # for a given Suite and Platform.
+    #
+    # @param suite [Suite,#name] a Suite
+    # @param platform [Platform,#name] a Platform
+    # @return Array[Provisioner] a new Provisioner object
+    # @api private
+    def new_provisioners(suite, platform)
+      # check if suite name otherwise is a suite step
+      if data.steps?(suite.name)
+        each_step(suite.name) do |step|
+          new_provisioner(step, platform)
+        end
+      else
+        [new_provisioner(suite, platform)]
+      end
+    end
+
+    # Builds a newly configured Provisioner object, for a given Suite
+    # and Platform.
     #
     # @param suite [Suite,#name] a Suite
     # @param platform [Platform,#name] a Platform
--- a/lib/kitchen/data_munger.rb
+++ b/lib/kitchen/data_munger.rb
@@ -104,7 +104,21 @@ module Kitchen
     #
     # @return [Array<Hash>] an Array of Hashes
     def suite_data
-      data.fetch(:suites, [])
+      @suite_data ||= data.fetch(:suites, []).inject([]) do |suites, suite|
+        if suite[:steps]
+          suite[:steps].each_with_index do |sub_suite, index|
+            sub_suite[:name] = "#{suite[:name]}_step_#{index + 1}"
+            suites << sub_suite
+          end
+        else
+          suites << suite
+        end
+        suites
+      end
+    end
+
+    def steps?(suite_name)
+      !data.fetch(:suites, []).find { |suite| suite[:name] == suite_name }
     end
 
     private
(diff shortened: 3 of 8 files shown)


tyler-ball:
My concern with this PR is that I don't know if it solves the referenced issues.  Converting the `steps` config to multiple VM definitions seems like a new way to list different suites.  Isn't the idea that I want to perform step 1 on the same VM as I perform step 2?

emiddleton:
@tyler-ball it solves the referenced issues.  step2 is step1 followed by step2.  Step1 is included so there is a way to test the initial converge worked before testing the second converge.  You could optimize this by copying the step1 vm and converging from there for step2 but that looked more involved and this is already a pretty big change.

tyler-ball:
Hey @emiddleton - thanks for the reply, sorry it took me so long to get back to you.  I understand now that instance 2 runs both step 1 _and_ step 2.

I'm going to tag this for 2.0 and move it back into the backlog.  For 1.4 we are making some significant changes to the responsibility of Provisioners and Drivers that will affect the code here.

While this does solve some of the linked problems, it is also solving them in a prescribed way.  I think of this as 'Deployment Strategy Testing'.  This does solve help test 1 deployment strategy but eventually we would like to tackle that issue on its own.  There are billions of deployment strategies and this approach only tests 1.

We also have some smaller scale changes we can make inside the provisioners that will allow chef-client to be converged multiple times.  That is a specific issue we want to solve with a smaller, targeted patch.

Finally, after we finish the current work to shift around provisioning responsibility, you will have the option of writing your own provisioner that is capable of performing these multi-converge and multi-test execution.  This would allow you to release your own Gem containing this multi-capable provisioner.

emiddleton:
@tyler-ball The most important use case for multiple convergence is to test state transitions that will effect a production system not deployment strategies which are far less critical. Breakage in deployment causes delays, breakage on production causes system downtime.  At the moment the only way to unit test if upgrading a cookbook version or any other cookbook configuration change will break a deployed system is to manually create each possible initial state.

Everyone I have spoken to just creates a production replica and doesn't bother with cookbook testing, which is a shame because it would be much easier to diagnose problems earlier and would result in much more robust cookbooks.

patcon:
:+1:

manuelmazzuola:
+1

doertedev:
Cool. I have a cookbook to run on a virgin Debian machine. I connect via root and do $things. The last thing is switching the sshd_config to not allow root anymore and set SSHD on a diff. port.
Would this allow me, if not asserting that the connection actually breaks, to just switch to diff. connection params in the second step?

beeisabelm:
:+1:

CpuID:
@emiddleton - I am thinking of running this PR via a forked gem until it is merged, do you feel it should be fine under 1.4, as per the comments from @tyler-ball above?

wjordan:
For anyone still looking to use this feature in 1.4, I merged this PR (resolving the merge conflicts) along with a couple other open bugfix/feature PRs into a fork at [wjordan/test-kitchen#cdo](https://github.com/wjordan/test-kitchen/tree/cdo).

howdoicomputer:
Hey so @emiddleton and @wjordan thank you so much for your work putting this together. I've been working on a Proxmox cookbook and, because Proxmox includes a heavily patched version of the Linux kernel, I've had a need for double convergence for a single suite. I've forked, merged into master, and tagged this cookbook into an internal repository within my company for use within the one cookbook I've been working on.

The only alternative that I can see at the moment was to build a Vagrant box for a post-kernel-installation recipe and then pull two boxes to test this.

Again, thank you so much.

howdoicomputer:
Also, to note: this breaks busser somehow. It won't pass on over my serverspec files. I'm looking into it.

wjordan:
If anyone (like myself) still depends on this PR and wants to use it with tk v1.5.0, I've made a rebased commit 721244eba68762c37ab8796f639885743ac0b769 available in branch [wjordan/test-kitchen#multi-step-converge](https://github.com/wjordan/test-kitchen/tree/multi-step-converge).

Hoping this feature integration gets moved out of the backlog soon. Or, if there's been any progress on the two work items @tyler-ball mentioned in an earlier [comment](https://github.com/test-kitchen/test-kitchen/pull/564#issuecomment-77267001) (1. allow chef-client to be converged multiple times; 2. shift around provisioning responsibility, so a separate provisioner gem could be written), I'd be happy to do the remaining work necessary to extract this feature into its own provisioner gem with a small bit of guidance.

tpcwang:
+1. I would really like to use a custom provisioner that allows for reboot during converge to help test things like kernel module installation and service starting. Is this issue still targeted for 2.0?

cheeseplus:
The feature has been "Accepted" as a new feature and is targeted for a major release, most likely 2.0 - that said I don't know anyone on TK core not doing any active development toward this goal as we're far more focused on bugs (vs features) currently.

AReshma:
This looks fantastic. Waiting for this feature.

wesm-outreach:
Any update on this? This has been open for years, and it looks like it's fairly high demand.

cheeseplus:
Lots of people want the feature but designing with the future and multi-node in mind is essentially where this is getting hung up as they are somewhat bound together as we're going to need to restructure some low level parts of kitchen. For reboot support we've already got that in chef-client and test-kitchen supports that so some of the +1s and issues in the thread have been resolved on that front.

The feature as written breaks enough downstream that shoehorning it in at this point is less than desirable to doing a ground up implementation alongside multi-node. These are non-trivial features and need careful consideration before being introduced and unfortunately I've largely been the only person maintaining for the better part of 2 years so large feature implementations have simply been backlogged behind bug fixes and just trying to keep things up-to-date.

Even if this were rebased, I'm not sure we'd take it until we've sorted out how this will fit in with other future plans.

cheeseplus:
We have https://github.com/chef/chef-rfc/blob/master/rfc084-test-kitchen-multi.md which covers this so closing the PR since it's part of our 2.x roadmap.

PR #566, opened 2014-12-18
Force busser to be run with correct/extant ruby under powershell

gh2k:
When spinning up an omnibus-installed chef, busser.bat points at a ruby that doesn't exist. As a result, when it's run by powershell it bombs with an error saying ruby doesn't exist.

This modifies override for busser_bin to force usage the ruby in ruby_bindir.

It's still not a very smart way to do this, but it got busser working on my setup.


gh2k:
Things have moved on and this is no longer relevant

PR #587, opened 2015-01-23
Busser: improve recursion of files in test suites

L2G:
This should allow directory symlinks to be dereferenced now.

### DIFF
--- a/lib/kitchen/busser.rb
+++ b/lib/kitchen/busser.rb
@@ -28,6 +28,11 @@ module Kitchen
   # @author Fletcher Nichol <fnichol@nichol.ca>
   class Busser
 
+    # Pattern to use with Dir.glob to return all paths in the current
+    # directory and all subdirectories, dereferencing symlinks along the
+    # way.
+    GLOB_RECURSE_ALL = '{*,**/*/**/*}'
+
     # Constructs a new Busser command generator, given a suite name.
     #
     # @param [String] suite_name name of suite on which to operate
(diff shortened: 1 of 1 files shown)


L2G:
I need to rework this; sorry.

coderanger:
What are the security implications of this? Probably moot since busser runs everything as root anyway, but symlink traversal sets off my security brain, should at least state what the attack surface is.

L2G:
Right now, symlinks to files are already being followed. This change would make it so that symlinks to directories are followed as well. Maybe there could be a security setting for either or both kinds of symlinks.

The idea of symlinked directories came about because right now there doesn't seem to be an easy way to share tests among multiple test suites. The alternatives are copying or symlinking individual test files.

coderanger:
This is a core thing in RSpec (assuming you are using Serverspec), see https://github.com/coderanger/shared-spec-example for an example.

L2G:
Thanks for the tip. I do know how to use RSpec shared examples. In fact, I am trying to leverage that feature a little further. What I am trying to do is use the same shared examples in my Test Kitchen suites and in live checks on remote servers (the latter being run by Serverspec tasks in Rake). Is there a better way to make a common set of shared examples available to all serverspec suites in a Test Kitchen? Is that what the `test/integration/helpers/serverspec` directory is for?

coderanger:
The helpers folder gets copied over for all suites, yes. You can also put your shared tests in a gem, and have Test Kitchen pull them in.

PR #588, opened 2015-01-28
Change getchef.com chef.io in Powershell provisioner

jdmundrawala:
This was causing us to try to use install.sh instead of install.msi

cc @fnichol @btm @mwrock @tyler-ball @afiune @mattstratton @cmluciano

Also, we need a github group because it's hard pinging everybody individually.


cmluciano:
👍

PR #590, opened 2015-01-29
Powershell no longer re-installs chef if version constraint is only major version

tyler-ball:
If `.kitchen.yml` specifies `require_chef_omnibus: 12`, Powershell was continually installing the latest version of Chef 12.  This fixes that bug.

\cc @fnichol @test-kitchen/windows 

Bug originally listed https://github.com/opscode-cookbooks/windows/pull/172#discussion_r23739543

### DIFF
--- a/lib/kitchen/provisioner/chef/powershell_shell.rb
+++ b/lib/kitchen/provisioner/chef/powershell_shell.rb
@@ -16,6 +16,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+require "httpclient"
+
 module Kitchen
 
   module Provisioner
@@ -88,6 +90,7 @@ module Kitchen
         # @option opts [String] :chef_omnibus_url URL to download the chef installer
         # @return [String] command to install chef
         def install_function(version, config)
+          version = expand_version(version)
           install_flags = %w[latest true].include?(version) ? "" : "v=#{version}"
 
           # If we have the default URL for UNIX then we change it for the Windows version.
@@ -111,6 +114,29 @@ module Kitchen
             }
           INSTALL
         end
+
+        private
+
+        # Converts a non-complete version (EG, 12) into a complete version (EG, 12.0.3)
+        def expand_version(version)
+          # TODO: all windows architectures and versions use the same Chef package
+          # when this changes, we will need to hit a different omnitruck URL per
+          # os version and architecture.
+          # https://docs.chef.io/api_omnitruck.html#supported-platforms
+          omnitruck = "http://www.chef.io/chef/metadata?p=windows&pv=2008r2&m=x86_64&v=#{version}"
+          client = HTTPClient.new(
+              :default_header => { "Accept" => "application/json" },
+              :follow_redirect => true
+          )
+          resp = client.request(:get, URI.parse(omnitruck))
+          if resp.code == 200
+            require "json"
+            body = JSON.parse(resp.body)
+            version = body["relpath"].gsub(/.+\/chef-(client|windows)-([^\-]+)-.+$/, '\2')
+          end
+          version
+        end
+
       end
     end
   end


lib/kitchen/provisioner/chef/powershell_shell.rb
```diff
@@ -90,6 +92,18 @@ def chef_helper_file
         def install_function(version, config)
           install_flags = %w[latest true].include?(version) ? "" : "v=#{version}"
 
+          # TODO all windows architectures and versions use the same Chef package
```
jdmundrawala: can this be pulled out into its own function?

cmluciano:
This looks similar to what I was thinking, but I got stuck on your TODO line. Is there a large difference among OS versions? As in, does each OS have it's own custom omnibus package?

tyler-ball:
@cmluciano No, every Windows install gets the same omnibus package.  Thats why I don't need to specify a different `pv` or `m` in the omnitruck URL.  Should I update the TODO to better reflect that?

cmluciano:
No I think it's fine as written. Thanks!

cmluciano:
Should this potentially live under a more commonly named base-helper.rb instead of powershell?

lib/kitchen/provisioner/chef/powershell_shell.rb
```diff
@@ -111,6 +114,26 @@ def install_function(version, config)
...
+          # os version and architecture.
+          # https://docs.chef.io/api_omnitruck.html#supported-platforms
+          omnitruck = "http://www.chef.io/chef/metadata?p=windows&pv=2008r2&m=x86_64&v=#{version}"
+          client = HTTPClient.new(:default_header => { "Accept" => "application/json" })
+          resp = client.request(:get, URI.parse(omnitruck))
+          if resp.code == 200
+            require "json"
+            body = JSON.parse(resp.body)
+            version = body["relpath"].gsub(/.+\/chef-(client|windows)-([^\-]+)-.+$/, '\2')
+          end
```
jdmundrawala: this needs to handle other http codes as well (400/500 etc).
Also, i don't belive this will handle redirects, there might be a parameter you can pass in to HTTPClient

tyler-ball:
@cmluciano I only think it should go in a common helper if it is going to be used in multiple places.  If that is the case, it needs to be more robust to support more than just windows.  But in that case I don't know how to specify to the function what OS is being used - we would need to introspect the vagrant vm or require users to specify it in the `.kitchen.yml`.  I'm not super familiar with the code base - is there a better location to call this helper?

@jdmundrawala I purposefully don't change the provided `version` on anything besides a 200.  In that case the version specified in the `.kitchen.yml` will be passed directly to the provisioner, like it was before this code was introduced.

I'll look into following redirects.

jdmundrawala:
ooo, i see. my bad... :+1:

cmluciano:
👍

lib/kitchen/provisioner/chef/powershell_shell.rb
```diff
@@ -111,6 +114,26 @@ def install_function(version, config)
...
+        private
+
+        # Converts a non-complete version (EG, 12) into a complete version (EG, 12.0.3)
+        def expand_version(version)
+          # TODO: all windows architectures and versions use the same Chef package
+          # when this changes, we will need to hit a different omnitruck URL per
+          # os version and architecture.
+          # https://docs.chef.io/api_omnitruck.html#supported-platforms
+          omnitruck = "http://www.chef.io/chef/metadata?p=windows&pv=2008r2&m=x86_64&v=#{version}"
+          client = HTTPClient.new(:default_header => { "Accept" => "application/json" }, :follow_redirect => true)
```
fnichol: Looks like this line could use a quick wrap, perhaps something like below? Otherwise, CI is happy

```ruby
client = HTTPClient.new(
  :default_header => { "Accept" => "application/json" },
  :follow_redirect => true
)
```

tyler-ball:
@fnichol Thanks - fixed Rubocop and rebased/squashed.  If this looks good to you, I'll merge it!

afiune:
I like this! :+1: Good job @tyler-ball :smile:

PR #592, opened 2015-02-01
Replace sudo -E

### LINKED ISSUE #307: Use of sudo -E breaks compatibility with CentOS 5
the version of sudo in CentOS 5 does not support the -E (preserve environment) flag, so omnibus installation fails.

ref: https://gist.github.com/nathwill/8137556

aaron-lane:
This fixes #307; any other changes are intended to be refactors to improve readability and reduce complexity.

`sudo su -m` was chosen to maintain compatibility with the current usage of `sudo -E`.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,3 +1,4 @@
-Style/FileName:
+AllCops:
   Exclude:
+    - 'spec/**/*'
     - 'support/**/*'
--- a/lib/kitchen/busser.rb
+++ b/lib/kitchen/busser.rb
@@ -97,20 +97,16 @@ module Kitchen
     def setup_cmd
       return if local_suite_files.empty?
 
-      ruby    = "#{config[:ruby_bindir]}/ruby"
-      gem     = sudo("#{config[:ruby_bindir]}/gem")
-      busser  = sudo(config[:busser_bin])
-
       cmd = <<-CMD.gsub(/^ {8}/, "")
         #{busser_setup_env}
-        gem_bindir=`#{ruby} -rrubygems -e "puts Gem.bindir"`
 
-        if ! #{gem} list busser -i >/dev/null; then
-          #{gem} install #{gem_install_args}
+        if ! #{busser_installed?}; then
+          #{busser_install}
         fi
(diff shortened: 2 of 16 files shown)


jtimberman:
Is `su -m` portable to all Linux/Unix-like platforms?

aaron-lane:
As far as I have seen, but other platform jockeys can certainly weigh in. I know the initial suggestion was to clear the environment with ```su -l```, but some of the scripts that TK generates assume access to the nonroot environment and I didn't want to go that far in changing behavior.

MarkGibbons:
I don't see the sudo -m option in the source at http://www.sudo.ws or in the man page http://www.sudo.ws/sudo/man/1.8.11/sudo.man.html.  The versions of sudo I have on rhel 5, rhel 6 and solaris doesn't support a -m option.

aaron-lane:
@MarkGibbons: The original command used was ```sudo -E```, but the change is to ```sudo su -m -c```, which should function in the same manner.

robcoward:
-1 - Not sure what you are trying to achieve, but -m -c don't appear to be valid options with the version of sudo shipped with CentOS6.5

aaron-lane:
@robcoward: Again, these are not sudo options; the full command is ```sudo su -m -c```.

martinisoft:
:+1: from FreeBSD land. I can confirm FreeBSD at least version 9.3-RELEASE and newer has support for those options in su. See https://www.freebsd.org/cgi/man.cgi?query=su&apropos=0&sektion=0&manpath=FreeBSD+10.1-RELEASE&arch=default&format=html

fnichol:
After talking this one over, I think we have an alternate solution that still leaves things configurable and maximally portable:

How about adding `:sudo_command` configurable option to Provisioner::Base (and Busser) that defaults to `"sudo -E"` which lets you override it at any level (all instances, a specific platform, etc.)?

Is this something you'd like to tackle or should we add an issue for tracking?

aaron-lane:
Sorry, I was offline for a few days.

I can definitely take a crack at implementing your idea.

Issue #85, opened 2013-04-23
Add an option to clean up log files generated


miketheman:
Right now, they continue serially forever.

It would help to have a "--log-overwrite" option that allows all logs to be cleaned out and started fresh for this session.

tyler-ball:
I'm going to add a "--[no-]log-overwrite" parameter (and config option) and then default it to log overwriting.  This means that the default behavior will change, but it will provide an option for backward compatibility.

PR #600, opened 2015-02-06
Adding --log-overwrite CLI option

tyler-ball:
Fixes https://github.com/test-kitchen/test-kitchen/issues/85

\cc @fnichol @sethvargo @miketheman

### DIFF
--- a/lib/kitchen.rb
+++ b/lib/kitchen.rb
@@ -70,16 +70,27 @@ module Kitchen
     #
     # @return [Logger] a logger
     def default_logger
-      Logger.new(:stdout => $stdout, :level => env_log)
+      Logger.new(:stdout => $stdout, :level => Util.to_logger_level(env_log))
     end
 
     # Returns a default file logger which emits on standard output and to a
     # log file.
     #
+    # @param [Symbol] level logging level
+    # @param [Boolean] log_overwrite logging level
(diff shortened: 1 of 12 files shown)


miketheman:
Looks good to me. The build seems to have errored out on Ruby 1.9.3 trying to install byebug, which is only compatible with Ruby 2.x and up - likely a dependency of one of the pry gems added to the gemspec.
You might want to consider removing those to have the build continue to test on 1.9.3, since it appears that the rest of the project continues to run there.

tyler-ball:
Thanks @miketheman - I got that and a few other bugs fixed.  Rebased and squashed.

lib/kitchen.rb
```diff
@@ -70,16 +70,27 @@ def source_root
     #
     # @return [Logger] a logger
     def default_logger
-      Logger.new(:stdout => $stdout, :level => env_log)
+      Logger.new(:stdout => $stdout, :level => Util.to_logger_level(env_log))
```
tyler-ball: I actually thought it would be cleaner to move all this logging config from `lib/kitchen.rb` to `lib/kitchen/logger.rb` - thoughts?

fnichol:
LGTM!

miketheman:
Whoop whoop!

PR #603, opened 2015-02-06
Proxy Implementation for Windows Chef Omnibus

afiune:
As a corporate end user I need Chef Omnibus for Windows to work behind a proxy so I can be able to use it on my company with proxy enabled.

### DIFF
--- a/support/chef_helpers.ps1
+++ b/support/chef_helpers.ps1
@@ -23,7 +23,11 @@ function should_update_chef($version) {
 # @param $chef_msi Temporal MSI path
 function download_chef($chef_url, $chef_msi) {
   Write-Host -NoNewline "`r`tDownloading Chef..."
-  (New-Object System.Net.WebClient).DownloadFile($chef_url, $chef_msi)
+  $proxy = new-object System.Net.WebProxy
+  $proxy.Address = $env:http_proxy
+  $wc = new-object system.net.WebClient
+  $wc.proxy = $proxy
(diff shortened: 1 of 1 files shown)


support/chef_helpers.ps1
```diff
@@ -23,7 +23,11 @@ function should_update_chef($version) {
 # @param $chef_msi Temporal MSI path
 function download_chef($chef_url, $chef_msi) {
   Write-Host -NoNewline "`r`tDownloading Chef..."
-  (New-Object System.Net.WebClient).DownloadFile($chef_url, $chef_msi)
+  $proxy = new-object System.Net.WebProxy
+  $proxy.Address = $env:http_proxy
```
jdmundrawala: Will this work if $env:http_proxy is null?

jdmundrawala:
I think it might be nice to log that a proxy is being used

support/chef_helpers.ps1
afiune: Yes it does. It is translated to: `there is no proxy` :smile:

afiune:
\cc @fnichol @test-kitchen/windows

tyler-ball:
:+1: 

@afiune @jdmundrawala If I understand the comment correctly, it will already output `there is no proxy` if  proxy is nil, correct?  That seems a good enough debug log.

fnichol:
Didn't realize this needed to be explicit.

PR #607, opened 2015-02-09
Using Metal as the Provisioner [provisional]

jkeiser:


cheeseplus:
This would be something we'd want as a separate, external provisioner gem.

Issue #380, opened 2014-03-07
Long runs timeout


sethvargo:
If Test Kitchen (and/or shellout) doesn't receive something in a certain about of time, it closes the connection. This very well may be vagrant, but I'm not sure.

Either way, we should persist the connection if output is being received. I personally experienced this while compiling a Ruby from source - TK just severed the connection to vagrant mid-build.

dAnjou:
@lonniev It seems this has been fixed in >=11.14 https://github.com/test-kitchen/test-kitchen/issues/380#issuecomment-41359083

cheeseplus:
This one is pretty old and not something we've seen in quite some time. Between vagrant, drivers, transport improvements all changing I'm closing this one but folks can always open a new issue (with lots of specific detail) should it present itself

PR #608, opened 2015-02-11
Add SSH keepalive options, so that long-running commands don't time out

irvingpop:
NOTE: WIP,  looking for help to make this configurable.

While working with a Chef user (Windows host, Linux guest), we noticed that his test-kitchen runs failed during very long resources where no output was emitted for several minutes. 

We noticed that net-ssh has `:keepalive` defaulted to `false`.   By changing it to true and lowering the `:keepalive_interval` we were able to eliminate the failures.

### DIFF
--- a/lib/kitchen/driver/ssh_base.rb
+++ b/lib/kitchen/driver/ssh_base.rb
@@ -120,6 +120,8 @@ module Kitchen
         opts[:port] = combined[:port] if combined[:port]
         opts[:keys] = Array(combined[:ssh_key]) if combined[:ssh_key]
         opts[:logger] = logger
+        opts[:keepalive] = true
+        opts[:keepalive_interval] = 60
 
         [combined[:hostname], combined[:username], opts]
       end


irvingpop:
This might also address the issues in:  https://github.com/test-kitchen/test-kitchen/issues/380

fnichol:
@irvingpop How does the above commit look? Note that this is in the new Transport abstraction code which will land in the next release of Test Kitchen (the one introducing Windows support). If that seems to cover it, then we're done, wee!

irvingpop:
hey @fnichol that looks totally awesome!

irvingpop:
Closing in favor of @fnichol version

fnichol:
Excellent, thanks all!

SydOps:
anyone here could teach me how to set the `keepaliva`  to true?

I guess my current problem in https://github.com/test-kitchen/kitchen-ec2/issues/276 is related

PR #609, opened 2015-02-11
Del :ssh_retries & :ssh_timeout before ssh connect

jdevesa:
`Net::SSH::start` does not support Kitchen's SSH `:ssh_retries` and
`:ssh_timeout` options. Remove them from the `options` hash before
establish the connection.

Please note that `Kitchen::SSH.new(*build_ssh_args(state))` never builds 
the `:ssh_retries` and `:ssh_timeout` options and they only are passed through
`Kitchen::SSH` when [testing the connection](https://github.com/test-kitchen/test-kitchen/blob/master/lib/kitchen/driver/ssh_base.rb#L180) 

where `Net::SSH::start` is not called.

### DIFF
--- a/lib/kitchen/ssh.rb
+++ b/lib/kitchen/ssh.rb
@@ -195,6 +195,9 @@ module Kitchen
         Net::SSH::Disconnect, Net::SSH::AuthenticationFailed
       ]
       retries = options[:ssh_retries] || 3
+      timeout = options[:ssh_timeout] || 1
+      options.delete(:ssh_retries) if options.has_key?(:ssh_retries)
+      options.delete(:ssh_timeout) if options.has_key?(:ssh_timeout)
 
       begin
         logger.debug("[SSH] opening connection to #{self}")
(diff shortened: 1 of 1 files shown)


fnichol:
Thanks for this, @jdevesa. It took me until last week to discover this fact as well (that it was never wired through).

In the meantime, we've been working on getting Windows supported in the codebase which has meant further (re)working of the SSH code--which is now a Transport strategy. You can see a nearly completed version of the [SSH Transport](https://github.com/fnichol/test-kitchen/blob/22738ce11350ea5223fe03c0ef2f922719c9d3c5/lib/kitchen/transport/ssh.rb) which will be included in the next release containing new [configuration options](https://github.com/fnichol/test-kitchen/blob/22738ce11350ea5223fe03c0ef2f922719c9d3c5/lib/kitchen/transport/ssh.rb#L41-L44) to deal with these timeouts and retries.

So we could merge in your PR (assuming the specs are made to pass), but know that the `:ssh_retries` and `:ssh_timeout` options will go away to be replaced with:

```yaml
transport:
  name: ssh
  connection_retries: 45
  connection_retry_sleep: 2
```

jdevesa:
Gotcha! Thanks for the information. I'll make sure to pin my current version of test-kitchen and change to the new version as soon as gets released!

Great work, btw.

fnichol:
Excellent, I'll close this out then. If I misread, then please reopen!

PR #610, opened 2015-02-11
Windows2003 guest fix

GolubevV:
May be not relevant for all Windows versions, however, the following changes in the transport workflow  made the 'windows-guest-support' branch of test kitchen work with my Windows 2003R2 box on Vagrant.
Please,contact me if you need any more info about the topic.

### DIFF
--- a/lib/kitchen/transport/winrm_file_transfer/remote_zip_file.rb
+++ b/lib/kitchen/transport/winrm_file_transfer/remote_zip_file.rb
@@ -93,7 +93,7 @@ module Kitchen
             $shellApplication = new-object -com shell.application
 
             $zipPackage = $shellApplication.NameSpace('#{remote_path}')
-            Remove-Item -Recurse -Force $dest | Out-Null
+            if ( Test-Path $dest ) { Remove-Item -Recurse -Force $dest | Out-Null }
             mkdir $dest -ErrorAction SilentlyContinue | Out-Null
             $destinationFolder = $shellApplication.NameSpace($dest)
             $destinationFolder.CopyHere($zipPackage.Items(),0x10) | Out-Null
--- a/lib/kitchen/transport/winrm_file_transfer/remote_zip_file.rb
+++ b/lib/kitchen/transport/winrm_file_transfer/remote_zip_file.rb
@@ -55,6 +55,7 @@ module Kitchen
           logger.debug("iterating files in '#{glob}'")
           Zip::File.open(archive, "w") do |zipfile|
             Dir.glob(glob).each do |file|
+              next if File.directory?(file)
               logger.debug("adding zip entry for '#{file}'")
               entry = Zip::Entry.new(
                 archive,
--- a/lib/kitchen/transport/winrm_file_transfer/remote_zip_file.rb
+++ b/lib/kitchen/transport/winrm_file_transfer/remote_zip_file.rb
@@ -93,6 +93,7 @@ module Kitchen
             $shellApplication = new-object -com shell.application
 
             $zipPackage = $shellApplication.NameSpace('#{remote_path}')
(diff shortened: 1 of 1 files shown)


tyler-ball:
This LGTM - can I get :+1: from someone in @test-kitchen/windows

btm:
@GolubevV do you have a copy of any of the debugging/error output from a failure on 2003? 

I'm kind of curious if we're removing recursive behavior here.

lib/kitchen/transport/winrm_file_transfer/remote_zip_file.rb
```diff
@@ -92,6 +93,7 @@ def extract_zip_command
             $shellApplication = new-object -com shell.application
 
             $zipPackage = $shellApplication.NameSpace('#{remote_path}')
+            Remove-Item -Recurse -Force $dest | Out-Null
```
mwrock: I think we should have either a `-ErrorAction SilentlyContinue` or wrap in a `Test-Path` since I believe this will result in an error if the directory does not exist. Pretty sure its a "non terminating" error but it may cause unnecessary confusion to see the error text.

GolubevV: Hello!
Looks like Windows 2003 R2 Server API does not handle correctly zip files, where threre are dedicated zip entries, created for the folders.
Initially, the TK was just hanging during the stage of transfering files to the guest. 
With the  '-l debug' option added to TK command like, 
I narrowed down the problem to the stage of unpacking the transfered 'kitchen.zip' archive with all cookbooks and chef staff. The unpacking was hanging on creating some subfolders, while the files in the root folder were unpacked correctly.

I wanted to verify the correctness of the archive and opened it in the  Windows explorer archive viewever, resulting in the following content (i took screenshot of apache cookbook subfolder, packed in kitchen.zip):
screenshot from 2015-02-12 11 02 14
So the directory entries in the zip files were interpreted by Windows Explorer API as files, though the archive was correctly unpacked, using unzip, 7zip or any other third-party tools.

Afterwards, I took the powershell commands from the 'remote_zip_file' script and tried to perform them manually in the powershell, resulting in the following error, which was causing the TK to hang:
screenshot from 2015-02-12 11 10 08
Tthe archived was extracted correctly by TK on Windows 2003 after removing the directory entries from the zip file, using code:
next if File.directory?(file)

GolubevV:
Hello!
Looks like Windows 2003 R2 Server API does not handle correctly zip files, where there are dedicated zip entries, created for the folders.
Initially, the TK was just hanging during the stage of transferring files to the guest. 
With the '-l debug' option added to TK command like, 
I narrowed down the problem to the stage of unpacking the transferred 'kitchen.zip' archive with all cookbooks and chef staff. The unpacking was hanging on creating some subfolders, while the files in the root folder were unpacked correctly.

I wanted to verify the correctness of the archive and opened it in the Windows explorer archive viewer, resulting in the following content (i took screenshot of apache cookbook subfolder, packed in kitchen.zip):
screenshot from 2015-02-12 11 02 14

So the directory entries in the zip file were interpreted by Windows Explorer API as files, though the archive was correctly unpacked, using unzip, 7zip or any other third-party tools.

Afterwards, I took the powershell commands from the 'remote_zip_file' script and tried to perform them manually in the powershell, resulting in the following error, which was causing the TK to hang:
screenshot from 2015-02-12 11 10 08

The archived was extracted correctly by TK on Windows 2003 after removing the directory entries from the zip file, using code:
next if File.directory?(file)

mwrock:
This looks good to me and I think it is a good catch! Not only does it fix 2k3 but this addresses a scenario where we currently simply copy new bits over the old and leave around unneeded files that may have been deleted and could potentially cause problems. I like the fact that this deletes the former bits before copying.

PR #612, opened 2015-02-12
Pinning winrm to newer version to support latest httpclient

tyler-ball:
Per bug https://github.com/test-kitchen/test-kitchen/commit/00470e9a49472fb23120f69965c8e760700e28de#commitcomment-9699914 reported by @mwrock 

\cc @test-kitchen/windows


mwrock:
I think we need to also constrain HTTP client to at least 2.5.3 since winrm 1.3 constrains to 2.2.0.2 which would continue to cause this error.

PR #620, opened 2015-02-24
Populate Chef Zero provisioner cookbook cache

pleary:
The Chef Zero provisioner can run the test client in local
mode, but when the test client is first created and invoked
it still needs to download cookbooks from the Chef Zero
server. I have found this initial "Synchronizing Cookbooks"
to be slow when there are many cookbooks in the test run list.

This patch copies the cookbooks used to populate the Chef
Zero server into the kitchen cache directory, which is the
cache directory of the test client. When the client runs
its initial "Synchronizing Cookbooks", the cookbooks will
already be in the cache and the initial synchronization
will be faster.

I should add that doing this will increase the size of the
sandbox needing to be transferred to the remote node.
Because of this I would recommend that something like
#466 is a prerequisite to this, otherwise uploading the

sandbox to remote provisioners can be prohibitively slow
(using local provisioners like Vagrant and Docker will
be faster straight away without needing packaging / 
compression).

### DIFF
--- a/lib/kitchen/provisioner/chef_zero.rb
+++ b/lib/kitchen/provisioner/chef_zero.rb
@@ -41,6 +41,7 @@ module Kitchen
       def create_sandbox
         super
         prepare_chef_client_zero_rb
+        prepare_chef_zero_cache
         prepare_validation_pem
         prepare_client_rb
       end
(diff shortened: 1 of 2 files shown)


cheeseplus:
The right way to fix this is fixing ChefFS which has seen significant improvements since this was opened. Also making it the default behaviour, especially in light of how common cloud drivers are in the ecosystem is far from ideal. Given these reasons and the age, closing.

PR #629, opened 2015-03-05
Allow custom sudo path. Also redact the password in the ssh opts

scotthain:
:construction: 

@fnichol - this is a possible way to do this, I'm totally open to other suggestions.

### DIFF
--- a/lib/kitchen/busser.rb
+++ b/lib/kitchen/busser.rb
@@ -38,6 +38,8 @@ module Kitchen
     #   containing the Ruby binary on the remote instance
     # @option opts [TrueClass, FalseClass] :sudo whether or not to invoke
     #   sudo before commands requiring root access (default: `true`)
+    # @option opts [String] :sudo_command sudo command for the system
+    #   (default: 'sudo -E')
     def initialize(suite_name, opts = {})
       validate_options(suite_name)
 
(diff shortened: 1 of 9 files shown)


btm:
Does this replace https://github.com/test-kitchen/test-kitchen/pull/592 ?

scotthain:
@btm it could actually - I didn't remove the -E flag as located here: https://github.com/test-kitchen/test-kitchen/blob/master/lib/kitchen/shell_out.rb#L59 but I could update to default to 'sudo -E' and allow for what is in #592.

scotthain:
@btm I think I've added all the things that #592 was looking to do. @aaron-lane didn't mean to steal your thunder! does this work for you as well?

aaron-lane:
@scotthain One of the raised issues was that `-E` is not supported on all platform implementations of sudo, so I'm not sure if this covers all of the original concerns.

scotthain:
@aaron-lane this defaults to 'sudo -E', but will allow you to provide (in .kitchen.yml) a different executable, be it 'sudo su -m -c' or '/my/custom/path/sudo'

That being said, there is a bug in this PR which I am fixing right now regarding busser.

aaron-lane:
@scotthain I'm guilty of not reading thoroughly :sad:. This sounds like an excellent solution!

scotthain:
@aaron-lane no worries! I'll try and have this updated a fixed fully either today or tomorrow. - Ran into some oddities with busser, as one does.

Issue #362, opened 2014-02-16
use a default path rubygems, ruby and busser


lmayorga1980:
Is it possible to use a default gem, ruby and busser in the path instead of specific paths?. I believe this is forcing me to install chef on my vagrant box (something i don't want to do). Otherwise, can i install an ubuntu chef package that make this error go away?

 ```      
sudo: /opt/chef/embedded/bin/gem: command not found
       sudo: /opt/chef/embedded/bin/gem: command not found
sh: 1:        /opt/chef/embedded/bin/ruby: not found
       sudo: /busser: command not found
       sudo: /tmp/busser/bin/busser: command not found

>>>>>> Setup failed on instance <default-puppet-ubuntu-precise64>.
>>>>>> Please see .kitchen/logs/default-puppet-ubuntu-precise64.log for more details
>>>>>> ------Exception-------
>>>>>> Class: Kitchen::ActionFailed
>>>>>> Message: SSH exited (1) for command: [sh -c 'BUSSER_ROOT="/tmp/busser" GEM_HOME="/tmp/busser/gems" GEM_PATH="/tmp/busser/gems" GEM_CACHE="/tmp/busser/gems/cache" ; export BUSSER_ROOT GEM_HOME GEM_PATH GEM_CACHE; if ! sudo -E /opt/chef/embedded/bin/gem list busser -i >/dev/null; then sudo -E /opt/chef/embedded/bin/gem install busser --no-rdoc --no-ri; fi; gem_bindir=`/opt/chef/embedded/bin/ruby -rrubygems -e "puts Gem.bindir"`; sudo -E ${gem_bindir}/busser setup; sudo -E /tmp/busser/bin/busser plugin install busser-serverspec']
>>>>>> ----------------------

```

cheeseplus:
This is the same root issue as - https://github.com/test-kitchen/test-kitchen/issues/347. We're working on abstracting Chef specific stuff out and will merge this into the other issue.

Issue #347, opened 2014-02-05
Busser depends on Ruby (ChefDK) being available on target VM


lmayorga1980:
I have a vagrant box with ubuntu precise 64 server with nothing on it. It seems that there are some requirements on the virtual machine in order to get it to work.

```
-----> Starting Kitchen (v1.1.2.dev)
-----> Cleaning up any prior instances of <default-puppet-ubuntu-precise64>
-----> Destroying <default-puppet-ubuntu-precise64>...
       [default] Forcing shutdown of VM...
       [default] Destroying VM and associated drives...
       Vagrant instance <default-puppet-ubuntu-precise64> destroyed.
       Finished destroying <default-puppet-ubuntu-precise64> (0m12.94s).
-----> Testing <default-puppet-ubuntu-precise64>
-----> Creating <default-puppet-ubuntu-precise64>...
       Bringing machine 'default' up with 'virtualbox' provider...
       [default] Importing base box 'ubuntu_master'...
       [default] Matching MAC address for NAT networking...
       [default] Setting the name of the VM...
       [default] Clearing any previously set forwarded ports...
       [default] Clearing any previously set network interfaces...
       [default] Preparing network interfaces based on configuration...
       [default] Forwarding ports...
       [default] -- 22 => 2222 (adapter 1)
       [default] Running 'pre-boot' VM customizations...
       [default] Booting VM...
       [default] Waiting for machine to boot. This may take a few minutes...
       [default] Machine booted and ready!
       [default] Setting hostname...
       Vagrant instance <default-puppet-ubuntu-precise64> created.
       Finished creating <default-puppet-ubuntu-precise64> (0m47.97s).
-----> Converging <default-puppet-ubuntu-precise64>...
       Preparing files for transfer
       Preparing data
       Preparing script
       Transfering files to <default-puppet-ubuntu-precise64>
Reading package lists... Done
Building dependency tree       
Reading state information... 0%
Reading state information... Done
       vim is already the newest version.
       0 upgraded, 0 newly installed, 0 to remove and 62 not upgraded.
       Finished converging <default-puppet-ubuntu-precise64> (0m3.19s).
-----> Setting up <default-puppet-ubuntu-precise64>...
       sudo: /opt/chef/embedded/bin/gem: command not found
       sudo: /opt/chef/embedded/bin/gem: command not found
sh: 1:        /opt/chef/embedded/bin/ruby: not found
       sudo: /busser: command not found
[...]

PR #632, opened 2015-03-09
install omnibus-chef before install busser if ruby missing for avoid `gem: command not found`

sawanoboly:
Hi, 

This PR is written for problem of https://github.com/test-kitchen/test-kitchen/issues/347. (and maybe https://github.com/test-kitchen/test-kitchen/issues/362 too?)
In rare cases, there is that I would like to run test on VM which was provisioned by shell with bats .

I think that this is a temporary implement until policy of yours will be determined.

### DIFF
--- a/lib/kitchen/busser.rb
+++ b/lib/kitchen/busser.rb
@@ -18,7 +18,8 @@
 
 require "base64"
 require "digest"
-require "Kitchen/provisioner/chef_base"
+require "kitchen"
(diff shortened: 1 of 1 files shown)


cheeseplus:
We want to address the root issue but in the meantime we aren't going to install Chef as part of the Busser run as a temporary fix. We should fix this the right way and we plan to.

PR #633, opened 2015-03-10
Fixing bad default setting - if ENV is not set we are accidently setting log_level to nil for whole run

tyler-ball:
This is a follow up to https://github.com/test-kitchen/test-kitchen/pull/600/files, which introduced a bug.  Not setting ENV[:log_level] causes the converge step to fail because we create the provisioner string `chef-client --log-level` with no supplied log level.

\cc @fnichol @schisamo

### DIFF
--- a/lib/kitchen/cli.rb
+++ b/lib/kitchen/cli.rb
@@ -64,6 +64,8 @@ module Kitchen
     # high
     MAX_CONCURRENCY = 9999
 
+    attr_reader :config
+
     # Constructs a new instance.
     def initialize(*args)
       super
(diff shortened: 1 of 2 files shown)


schisamo:
@tyler-ball This fixes the issue for me. It would be nice to add some test coverage to this PR so we ensure the issue doesn't bite us again in the future!

scotthain:
+1 to testing these defaults but yay! Thanks y'all!

tyler-ball:
Well I am really glad I added coverage - because I found a bug with my original code!  I'll merge this if the test looks good to yall

schisamo:
@tyler-ball once Travis is happy you should :shipit:

scotthain:
:+1: rock on

fnichol:
Yippie, agreed!

PR #644, opened 2015-03-21
open zip file in binary mode to avoid corrupting zip files on  windows

### LINKED ISSUE #643: RubyZip is corrupting zip files on windows hosts
This is related to #635 that @webframp, @fnichol and I were discussing earlier. Its the same cookbook, but i dont think the cookbook matters. After specifying a communicator and getting `kitchen create` to work, individual files were copied to `temp/kitchen` but not the cookbook directories. After troubleshooting, I found that the initial zip file created was corrupted. I couldn't open it. The same zip generated on a linux host is ~1k smaller and is just fine. I have a strong suspicion that this is related to line endings. In 2015 the type writer and its carriage return still haunts us.

I'll spend some time this weekend looking further into this.

mwrock:
fixes #643 by opening the zip to create in binary mode. This prevents the output stream from getting corrupted by extra carriage returns.

### DIFF
--- a/lib/kitchen/transport/winrm/tmp_zip.rb
+++ b/lib/kitchen/transport/winrm/tmp_zip.rb
@@ -79,7 +79,7 @@ module Kitchen
           @dir = Pathname.new(dir)
           @method = ::Zip::Entry::DEFLATED
           @compression = Zlib::BEST_COMPRESSION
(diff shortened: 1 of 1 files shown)


tyler-ball:
I tested this on my windows VM and got uncorrupted ZIPs with your change - merging!

PR #647, opened 2015-03-24
If bundler is defined, require any items defined within Gemfile

chrisroberts:
ohai! I would like to be able to have some gems be loaded automatically based on the Gemfile. This change allows that if Bundler is around.

### DIFF
--- a/bin/kitchen
+++ b/bin/kitchen
@@ -10,4 +10,6 @@ require "rubygems"
 require "kitchen/cli"
 require "kitchen/errors"
 
+Bundler.require if defined?(Bundler)
+
 Kitchen.with_friendly_errors { Kitchen::CLI.start }


tyler-ball:
This seems like a pretty wide-open change.  Is there a specific scenario you want to load items before Kitchen gets loaded?

The fear is that having a Gemfile with `gem 'kitchen-vagrant'` and using this change could cause load order issues.  There would definitely have to be testing around that.

luckymike:
@tyler-ball Given that the Chef-prescribed approach is not to use Bundler, it seems like the error you're concerned with would only occur in "power-user/operator-beware" scenarios. We discussed this with @fnichol at Chef Conf as an immediate solution to the dep-solver limitations mentioned in #686.

cheeseplus:
Given this was inspired around supporting Librarian and we've not seen much more interest since I'm closing this one.

PR #658, opened 2015-03-26
Updated for sh compatibility based on install.sh code

scotthain:
This PR fixes some of the issues that non updated bash shells have!


scotthain:
@fnichol we should probably change the hardcoded /tmp/stderr to $kitchen_tmp/stderr or something.
Also don't forget the error text! ;)

fnichol:
Sweet, probably a commit or rebase to clean this up, test for any regressions, and then we're in a good place!

PR #661, opened 2015-03-27
cleanup sh

curiositycasualty:
Bunch of niceties to help the code be a little more explicit and maintainable:
- escape and interpolate variables portably
- uniform if/then declarations
- uniform spacing
- prev. escaping not needed with  subshell syntax

Tested and working within Solaris 10u11's `sh`.

### DIFF
--- a/support/chef_base_install_command.sh
+++ b/support/chef_base_install_command.sh
@@ -8,31 +8,44 @@ exists() {
   fi
 }
 
-# do_wget URL FILENAME
 do_wget() {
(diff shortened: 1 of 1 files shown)


fnichol:
@curiositycasualty  I think we both worked on this at the same time! I ended up rebasing your work into #658, think it's okay?

fnichol:
After talking this through offline with @curiositycasualty, we should be okay to close this for #658, woo!

PR #666, opened 2015-03-27
re-adds PATH

curiositycasualty:
??


scotthain:
@fnichol Can we add this to the 'existing' path and add /usr/sfw/bin to the tacked-on path?

fnichol:
Seems like the absolutely perfect PR number

fnichol:
@scotthain I'll see what that looks like.

@curiositycasualty Were you able to narrow down what required these lines? We weren't able to determine why they were there in the first place…

fnichol:
Done!

PR #675, opened 2015-03-29
[Transport::Ssh] Add default :compression & :compression_level attrs.

fnichol:
This commit introduces 2 new configuration attributes for all SSH
Transports:

* `:compression` - can be set to `"zlib"` or `"none"`, default is `"zlib"`
* `:compression_level` - can be set to a number between 0 and 9 where 0
  is uncompressed and 9 is maximum compression, default is `9`

Closes #424


lamont-granquist:
just a note that i flipped the default to false/"none" in #1023 based on perf testing in travis to ec2

Issue #671, opened 2015-03-29
Wrong permissions in /tmp/verifier/gems/[bin/cache/gems]?


tknerr:
Up until test-kitchen 1.3.1 I could use [vagrant-cachier](https://github.com/fgrehm/vagrant-cachier) to cache the whole `/tmp/busser/gems` directory by adding this to my global `~/.vagrant.d/Vagrantfile`:
```
# global Vagrant configuration
Vagrant.configure("2") do |config|

  # enable cachier globally
  if Vagrant.has_plugin?("vagrant-cachier")
    config.cache.scope = :box

    # cache bussers for test-kitchen
    config.cache.enable :generic, {
      "busser" => { cache_dir: "/tmp/busser/gems" }
    }
  end
end
```

This gave me a noteable speedup for running the kitchenci tests. Without caching, it would download and install the bussers for every kitchenci run again.

I have noticed that the path changed to `/tmp/verifier/gems` now, but once I try to cache this path with 1.4.0.beta.2 the kitchen run will fail with some write permission error.

I'm still investigating the exact cause for it as it worked without problems with 1.3.1. One thing that looks suspicious to me is the ownership of the files in `/tmp/verifier/gems/[bin/cache/gems]` directories:
```
vagrant@default-ubuntu-1204:~$ ll /tmp/verifier/gems/gems/
total 72
drwxrwxr-x 18 vagrant vagrant 4096 Mar 29 14:12 ./
drwxrwxr-x  9 vagrant vagrant 4096 Mar 29 14:12 ../
drwxr-xr-x  5 root    root    4096 Mar 29 14:12 bundler-1.9.1/
drwxrwxr-x  7 vagrant vagrant 4096 Mar 29 14:12 busser-0.7.0/
drwxr-xr-x  4 root    root    4096 Mar 29 14:12 busser-serverspec-0.5.3/
drwxr-xr-x  7 root    root    4096 Mar 29 14:12 diff-lcs-1.2.5/
drwxr-xr-x  3 root    root    4096 Mar 29 14:12 multi_json-1.11.0/
drwxr-xr-x  4 root    root    4096 Mar 29 14:12 net-scp-1.2.1/
drwxr-xr-x  5 root    root    4096 Mar 29 14:12 net-ssh-2.9.2/
drwxr-xr-x  3 root    root    4096 Mar 29 14:12 rspec-3.2.0/
[...]

PR #677, opened 2015-03-31
ensure that $BUSSER_ROOT is owned by the current user... 

tknerr:
...to compensate for permission issues with vagrant synched folders

This PR is meant to work around the [permission issue mentioned here](https://github.com/test-kitchen/test-kitchen/issues/671#issuecomment-87449673) which occurs if parts of `$BUSSER_ROOT` (e.g. `/tmp/verifier/gems/cache`) are mounted via vagrant-cachier / vagrant synched folders.

I really don't like this workaround and would be happy if you decline it. Really. I mainly created the PR to get feedback on better ways / places to fix it.

**Considerations:**
- Since vagrant-cachier does not support windows guests I have only considered the `busser_install_command.sh`, not the .bat variant
- Initially wanted to place that fix somewhere in kitchen-vagrant (since it fixes a problem that occurs in a vagrant specific setup), but I decided against it because kitchen-vagrant does not know anything about `$BUSSER_ROOT`
- :warning: this might break things if the `$USER`'s group is not `$USER`!

### DIFF
--- a/support/busser_install_command.sh
+++ b/support/busser_install_command.sh
@@ -1,3 +1,9 @@
+
+# Ensure that $BUSSER_ROOT is owned by the current user, which migh not be the
+# case if parts of $BUSSER_ROOT are mounted via vagrant synched folders.
+# See https://github.com/test-kitchen/test-kitchen/issues/671 
+sudo chown -R $USER:$USER $BUSSER_ROOT
+
 $gem list busser -i 2>&1 >/dev/null
 if test $? -ne 0; then
   echo "-----> Installing Busser ($version)"


fnichol:
@tknerr Thanks for your feedback on this. I'll see what we can do this week to make the setup more cachier-friendly. The issue that we're dealing with is that now the SSH Transport SCPs the files over as the non-root user and thus wouldn't have permission to create the `$BUSSER_ROOT/suites`. Does cachier have to have root own all directories going back up to root on the synced folders?

tknerr:
@fnichol it's actually how vagrant synced folders work, not vagrant-cachier itself:

> The permissions you specify on shared folders will only apply to the deepest child. Otherwise, typical umask takes place

(from here: https://github.com/mitchellh/vagrant/issues/2257#issuecomment-24906376)

Since I wanted to make vagrant-cachier both optional and transparent (i.e. without changing .kitchen.yml), I added caching via my global `~/.vagrant.d/Vagrantfile`. I also added a shell provisioner there to fix the permissions, but it does not run since kitchen-vagrant creates the VM with the `--no-provision` flag. 

Hmmm... I just saw there is a `provision: false` config option in kitchen-vagrant. 
Oh. And I just saw there is a global `~/.kitchen/config.yml` as well!

I could probably combine the above to make it work just with the global vagrant & kitchen config. That way it would be completely external and we wouldn't need to change anything in test-kitchen or the kitchen-vagrant driver. Let me try...

tknerr:
Looks good! (I should have rubber-ducked this scenario before, sorry ;-))

So now I have in my `~/.vagrant.d/Vagrantfile`:
```ruby
Vagrant.configure("2") do |config|

  # enable cachier globally
  if Vagrant.has_plugin?("vagrant-cachier")
    config.cache.scope = :box

    # cache bussers for test-kitchen
    config.cache.enable :generic, {
      "verifier-gems" => { cache_dir: "/tmp/verifier/gems/cache" }
    }
    # correct permissions
    config.vm.provision "shell", inline: "chown -R vagrant:vagrant /tmp/verifier"
  end
end
```

And in `~/.kitchen/config.yml`:
```yaml
---
driver:
  name: vagrant
  # run provisioners (e.g. from global Vagrantfile)
  provision: true
```

However, this creates an easily overlooked soft dependency between the two files. Ideally I'd drive that specific kitchen-vagrant configuration from within the global Vagrantfile.

I have seen no env var like `KITCHEN_VAGRANT_PROVISION` or such that I could set from within the Vagrantfile, is there? (But it probably wouldn't help anyway as it's likely too late then already...)

Or any other ideas how to drive the kitchen-vagrant config from within the Vagrantfile?

tknerr:
Also: is it correct to cache only `/tmp/verifier/gems/cache`?

I was about to cache the whole `/tmp/verifier/gems` directory, but that would probably opt me out from any busser updates (as [per this](https://github.com/test-kitchen/test-kitchen/blob/4f2e2189937cb2e83ef173e3b48aef46a05ef1fa/support/busser_install_command.sh#L5-7)) once it's in the cache, I suspect.

PR #697, opened 2015-04-17
Set default log level even if you forget to add it to command line arg

scotthain:
Fixes a silly issue for people (like me) who forget how to do use the command line.

This solves the silly use case of:

'kitchen converge --log_level' (with nothing else specified)

If you'd prefer to have it blow up on this error, that's cool, you can probably use the same validation logic and just raise instead of defaulting.

### DIFF
--- a/lib/kitchen/cli.rb
+++ b/lib/kitchen/cli.rb
@@ -344,17 +344,8 @@ module Kitchen
     #
     # @api private
     def update_config!
-      if options[:log_level]
-        # validate log level specified on command line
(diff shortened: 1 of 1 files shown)


adamleff:
@scotthain yeah, so... bit of a backlog on this project, huh? :smile: I refactored this a bit to pass tests, and I think it's good to go now. I'm :+1:.

@cheeseplus take a looky-look, please?

cheeseplus:
+1 - love me some UX fixes

PR #702, opened 2015-04-24
don't prompt for passwords when using public keys

caboteria:
Sometimes when an ssh server is starting up it will accept connections
but fail any authentication attempts, even ones that should be valid.
By default, the ssh client will try to prompt for a password (the
"keyboard-interactive" authentication method) if the "publickey" method
fails, but in our case that's not desired.  We'd prefer that the
connection attempt fail immediately, which will cause a delay and retry.

This commit sets the ssh :auth_methods option to ['publickey'] if a
public key is provided.  This prevents ssh from prompting for a password
which will cause the connection attempt to fail and trigger a retry.


caboteria:
Sorry about the Rubocop failures, I'll fix them and open another PR.

PR #704, opened 2015-04-24
don't prompt for passwords when using public keys

caboteria:
Sometimes when an ssh server is starting up it will accept connections
but fail any authentication attempts, even ones that should be valid.
By default, the ssh client will try to prompt for a password (the
"keyboard-interactive" authentication method) if the "publickey" method
fails, but in our case that's not desired.  We'd prefer that the
connection attempt fail immediately, which will cause a delay and retry.

This commit sets the ssh :auth_methods option to ['publickey'] if a
public key is provided.  This prevents ssh from prompting for a password
which will cause the connection attempt to fail and trigger a retry.

### DIFF
--- a/lib/kitchen/transport/ssh.rb
+++ b/lib/kitchen/transport/ssh.rb
@@ -310,6 +310,7 @@ module Kitchen
 
         opts[:keys_only] = true                     if data[:ssh_key]
         opts[:keys] = Array(data[:ssh_key])         if data[:ssh_key]
+        opts[:auth_methods] = ["publickey"]         if data[:ssh_key]
         opts[:password] = data[:password]           if data.key?(:password)
         opts[:forward_agent] = data[:forward_agent] if data.key?(:forward_agent)
 
--- a/spec/kitchen/transport/ssh_spec.rb
+++ b/spec/kitchen/transport/ssh_spec.rb
@@ -483,6 +483,16 @@ describe Kitchen::Transport::Ssh do
         make_connection
       end
 
+      it "sets :auth_methods to only publickey if :ssh_key is set in config" do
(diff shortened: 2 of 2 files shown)


caboteria:
Not sure what to do here.  Rubocop passes on Ruby 2.2.0-p0 on OSX so I don't think my change caused the segfault.

tyler-ball:
Hey @caboteria - thanks for the PR and explanation!  Can you update [ssh_spec.rb](https://github.com/caboteria/test-kitchen/blob/no-password-prompt-2/spec/kitchen/transport/ssh_spec.rb#L518) with a spec?  Around line 518 you can see the existing specs so it should be pretty easy to copy from those.  After that, this should be ready for merge!

caboteria:
Thanks for the pointer to similar tests.  I added a test for the new code, and again Ruby crashes on one of the 2.2.0 runs.  As before I don't see how what I did could cause Ruby to crash.  Meh.

caboteria:
Looks like the build ran again this afternoon and passed this time so we should be good to go.

fnichol:
Looking good here, :+1: 

Thanks @caboteria for your follow up!

mfischer-zd:
This change didn't actually result in a fix for me. 

Does test-kitchen use Net::SSH or command-line ssh when logging into the host?  If the latter, we also need to add `-o PasswordAuthentication no` to the command line.

tyler-ball:
@mfischer-zd it uses `Net::SSH`

patcon:
Hm. Odd. I'm seeing this too when running DigitalOcean instances via travis, and i worry it will run up bills. I'll try to investigate

PR #706, opened 2015-04-28
Add ssl_ca_path default configuration entry, so we can override it, fixes #594

### LINKED ISSUE #594: ssl_ca_path cannot be set in kitchen client.rb
When i use chef 12.0.3, i need to pass a ssl_ca_path to client.rb to access some of my servers. It works on production... But when using kitchen, it's not working.

If i go into the vagrant box, to modify the client.rb located in /tmp/kitchen, it works.

 I verified that the configuration is loaded into dna.json, but it seems that it's not read, or don't override anything.

What's the good way to add ssl_ca_path into client.rb ? Do i need to propose a PR ?

Update: i test under debian.

sinfomicien:
Fixes https://github.com/test-kitchen/test-kitchen/issues/594

Without that, th default curl tool is unable to mange https self generated certificates.

### DIFF
--- a/lib/kitchen/provisioner/chef_base.rb
+++ b/lib/kitchen/provisioner/chef_base.rb
@@ -189,6 +189,7 @@ module Kitchen
           :validation_key   => remote_path_join(root, "validation.pem"),
           :client_key       => remote_path_join(root, "client.pem"),
           :chef_server_url  => "http://127.0.0.1:8889",
+          :ssl_ca_path      => "/etc/ssl/certs",
           :encrypted_data_bag_secret => remote_path_join(
             root, "encrypted_data_bag_secret"
           )


odolbeau:
:+1:

tyler-ball:
Hey @sinfomicien - the travis failure is coming from the lint tool.  Run `bundle exec rake` locally to repro it.

I'm unsure of the original issue too.  How does modifying the Chef config affect curl?  Do you have an example or a stacktrace of the original error?

sinfomicien:
Ok... My local debian mirror repository is on a local https. The certificate of this https server is signed by our own Certficate Authority. But to add this repository at the beginning of our vagrant up, the curl used by test-kitchen do not accept our system added CA. So the first curl to grab the gpg key used to sign package, from a url secure (https) with a certificate signed by our own CA, fail... The only way to be able to use such a configuration, is to force the ssl_ca_path. And it's possible only if we add it the default configuration hash of kitchen. That's my PR :)

Will correct simple quote, to double quote, so Travis check will pass.

tyler-ball:
Hey @sinfomicien, I think we were confused a little the first time.  Reading http://docs.chef.io/config_rb_client.html it looks like `ssl_ca_path` is already defaulted in the Chef Config without having to specify it in the client.rb.  So I'm concerned that overriding the default in TK will break more people than it will help.  It will also not work for Windows users.

Can you just set `client_rb` in your .kitchen.yml as described in https://docs.chef.io/config_yml_kitchen.html ?  This setting should have the same effect as your current PR.

Am I understanding your use case?  If you try that setting instead of this PR and are still getting a failure, will you post the full log output from your `.kitchen/logs` folder as well as the output from `kitchen diagnose`?

tyler-ball:
There hasn't been movement on this in 3 weeks, so I'm going to close it.  I'm hoping that our example of how to add custom client.rb code has worked for you and thats why there hasn't been a reply!

Issue #717, opened 2015-05-18
kitchen exec fails to show text content without linebreak


nogayama:
When I tried to see a file content in vagrant box, I faced this issue.

To reproduce, first we just make a file which end with ASCII character not line break.

```
$ kitchen login default
   > echo -n "hello" > /tmp/bar
   > exit
```

Then print it with "exec" command.
```
$ kitchen exec default -c "cat /tmp/bar"
-----> Execute command on default-centos65.
$
```

After several trial and error, I finally found that the following command works.
```
$ kitchen exec default -c "cat /tmp/bar; echo"
-----> Execute command on default-centos65.
hello
$
```

That is, the kitchen exec does not show an ASCII character only String to console. I guess this is not intended behavior.

sawanoboly:
I don't think that data transferring between host and vm via stdout is better way.

Having said that, it seems kitchen-logger's bug that ignoring last line of output without line break. I'll try fix it.

nogayama:
Thanks @sawanoboly.

I also agree with that using this to transfer file is not the correct way. But we are usually recording stdout to save the result of command execution. So I hope this will be fixed near future because `kitchen exec` is very useful and adds significant flexibility to test-kitchen.

PR #718, opened 2015-05-18
flush last line without linebrake to logger

sawanoboly:
This PR for #717 .

```
$ kitchen exec centos -c 'echo -n foo'
-----> Execute command on default-centos-66.
       foo

$ kitchen exec centos -c 'echo foo ; echo -n var'
-----> Execute command on default-centos-66.
       foo
       var

$ kitchen exec centos -c 'echo -n foo ; echo -n var'
-----> Execute command on default-centos-66.
       foovar
```

### DIFF
--- a/lib/kitchen/logger.rb
+++ b/lib/kitchen/logger.rb
@@ -38,6 +38,9 @@ module Kitchen
     #   overwriting
     attr_reader :log_overwrite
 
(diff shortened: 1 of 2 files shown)


lamont-granquist:
#824 may have fixed this more correctly.  should open a new issue if there's still a problem here.

PR #726, opened 2015-05-24
Use a relative name for the connection class.

coderanger:
This makes life easier for subclassing.

### DIFF
--- a/lib/kitchen/transport/ssh.rb
+++ b/lib/kitchen/transport/ssh.rb
@@ -330,7 +330,7 @@ module Kitchen
         end
 
         @connection_options = options
-        @connection = Kitchen::Transport::Ssh::Connection.new(options, &block)
+        @connection = self.class::Connection.new(options, &block)
       end
 
       # Return the last saved SSH connection instance.


metadave:
+1

tyler-ball:
@coderanger How does this make it easier for subclassing?  Wouldn't you still need `< Kitchen::Transport::Ssh` on a subclass?

coderanger:
@tyler-ball Right now you have to copy-pasta the whole method for any subclass. https://github.com/coderanger/kitchen-sync/blob/master/lib/kitchen/transport/sftp.rb#L29-L40

coderanger:
The problem is you need two classes for a transport plugin and the glue between them is buried inside a method with other logic. The other fix for this would be to make a method that returns the connection class and use it in `create_new_connection`, so subclasses can override just that.

tyler-ball:
@coderanger Gotcha - it makes it invoke the embedded `Connection` class rather than always invoking the `SSH::Connection` class.  Thanks for the explanation!

coderanger:
That was the intent at least, relative constant lookups in Ruby are a weird minefield so I very well might have gotten that wrong :-)

coderanger:
@cheeseplus At some point this got reverted. Can we revisit it?

PR #730, opened 2015-05-29
Update kitchen.yml.erb

jjasghar:
We should update this so it talks to atlas and queries correctly. `centos-7.1` does return anything.

### DIFF
--- a/templates/init/kitchen.yml.erb
+++ b/templates/init/kitchen.yml.erb
@@ -6,8 +6,8 @@ provisioner:
   name: <%= config[:provisioner] %>
 
 platforms:
-  - name: ubuntu-14.04
-  - name: centos-7.1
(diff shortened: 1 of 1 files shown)


jtimberman:
Actual retrieval of boxes is deferred to the [kitchen-vagrant driver](https://github.com/test-kitchen/kitchen-vagrant/blob/0983ba50239df96576b5b2bc23d5ff5e8b35ccc6/lib/kitchen/driver/vagrant.rb#L104-L114) - it automatically retrieves from the S3 bucket where we publish bento boxes, not from Atlas.

cheeseplus:
https://github.com/test-kitchen/kitchen-vagrant/issues/193

cheeseplus:
Closing this as mentioned above it occurs at driver levels and there is an open issue on kitchen-vagrant for it

PR #731, opened 2015-06-02
Revert "Use a relative name for the connection class."

metadave:
Reverts test-kitchen/test-kitchen#726

Travis was all green, but should have had it's tests rerun.

### DIFF
--- a/lib/kitchen/transport/ssh.rb
+++ b/lib/kitchen/transport/ssh.rb
@@ -330,7 +330,7 @@ module Kitchen
         end
 
         @connection_options = options
-        @connection = self.class::Connection.new(options, &block)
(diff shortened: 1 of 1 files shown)


randomcamel:
:+1:

metadave:
merging even though Travis is failing

tyler-ball:
@metadave @coderanger Do you want to re-open the original PR?  What was the issue that caused this to be reverted?

coderanger:
Yeah, would like to know what went wrong with this. The original issue still needs to be addressed one way or another.

coderanger:
Can we get some kind of explanation on this? This change or something like it is required to move forward with a decent transport API.

PR #734, opened 2015-06-04
Fix failing feature in `kitchen drvier discover` due to too many gems.

fnichol:
The code for `kitchen driver discover` was written when there weren't
many RubyGem Drivers published, however this is no longer the case.
Worse, the gems that are returned could be Provisioners, Transports,
etc.

In an attempt to make the output readable, the number of results was
arbitrarily cut off at 49 before adding "...". Guess what? There are now
many more than 49 results!

The easiest fix for this was to remove "kitchen-vagrant" (which fell off
the results due to its sorting), but the usefulness of this subcommand
might not be that high. This is a potential opportunity for deprecation
in the future.

### DIFF
--- a/features/kitchen_driver_discover_command.feature
+++ b/features/kitchen_driver_discover_command.feature
@@ -16,5 +16,4 @@ Feature: Search RubyGems to discover new Test Kitchen Driver gems
   Scenario: Running driver discover returns live results
     When I run `kitchen driver discover`
     Then the exit status should be 0
-    And the output should contain "kitchen-vagrant"
     And the output should contain "kitchen-bluebox"


metadave:
The PR resolves the gem issue, but now I'm getting a:

```
(::) failed steps (::)

Exiting RubyGems with exit_code 1 (Gem::SystemExitException)
./lib/kitchen/generator/init.rb:251:in `block in install_gem'
./lib/kitchen/generator/init.rb:283:in `unbundlerize'
./lib/kitchen/generator/init.rb:251:in `install_gem'
./lib/kitchen/generator/init.rb:232:in `block in add_drivers'
./lib/kitchen/generator/init.rb:228:in `each'
./lib/kitchen/generator/init.rb:228:in `add_drivers'
./lib/kitchen/generator/init.rb:69:in `init'
./lib/kitchen/cli.rb:308:in `invoke_task'
./features/support/env.rb:21:in `execute!'
features/kitchen_init_command.feature:174:in `When I successfully run `kitchen init`'

Failing Scenarios:
cucumber features/kitchen_init_command.feature:173 # Scenario: Running without git doesn't make a .gitignore
```

This results were produced using a fresh clone + checkout of `fnichol/fix-kitchen-driver-discover-failure`

metadave:
hm, if the tests pass for Travis and not for me, I'm guessing it's something I'm doing.

PR #737, opened 2015-06-08
generate a chefignore during init, fixes #732

### LINKED ISSUE #732: 'kitchen init' should create a chefignore file
Reference: https://github.com/test-kitchen/test-kitchen/issues/699

Windows users run into issues when Berkshelf tries to inspect files in the `.kitchen` folder and it has suuuuuuper long paths.  This is fixed by adding a `chefignore` file that tells it to ignore the `.kitchen` folder because berkshelf will respect that.

Also update the `chef generate cookbook` command to add `.kitchen` to the list of ignored files.

There is a bigger issue here that we're going to have to continue updating the `chefignore` file as more utilities add extra folders to the cookbook directory.  As bonus points, create an RFC to start managing cookbook contents either 1) the same way the server does or 2) the way a gemspec file does.

metadave:
as the tests on master won't pass locally for me, let's see what Travis has to say.

fixes https://github.com/test-kitchen/test-kitchen/issues/732


tyler-ball:
I updated the title to include `fixes #732` because then waffle will link the issue and the PR.  I also updated the description the same way so that Github will close the issue when the PR is merged.

templates/init/chefignore.erb
```diff
@@ -0,0 +1 @@
+*/.kitchen/*
```
tyler-ball: I think we decided that `.kitchen` was the appropriate entry.

tyler-ball:
LGTM! :+1:

PR #738, opened 2015-06-09
Adds number_of_password_prompts as a SSH option

wonko:
This allows this option of Net::SSH to be changed from within a driver. This
is needed when a VM template spins up an ssh daemon when the cloud-provided
password isn't set yet. This would prompt for a password, while this should
just fail.

The cloudstack driver needs this option to allow wait_for_sshd to wait for
an actual working sshd, and the possibility to login with the user/pass combo.

I assume other drivers might benefit from this too.

I actually need this to fix the "wait_for_sshd" in the cloudstack driver, on templates which have an sshd started before they set the password through the cloud-tools, and thus prompt for a password when the server is hit in between these two actions. 

I added some basic (copy-paste) tests, but I have no idea how i should actually test the correct functioning of this specific option. I have tested this locally on a working setup (with the needed changes in the cloudstack driver added, verifying the fix to my problem).

### DIFF
--- a/lib/kitchen/transport/ssh.rb
+++ b/lib/kitchen/transport/ssh.rb
@@ -48,6 +48,7 @@ module Kitchen
       default_config :connection_retries, 5
       default_config :connection_retry_sleep, 1
       default_config :max_wait_until_ready, 600
+      default_config :number_of_password_prompts, 3
 
       default_config :ssh_key, nil
       expand_path_for :ssh_key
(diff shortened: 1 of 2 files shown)


cheeseplus:
Given the age of the PR and lack of further use case validation we're closing this one. If the feature is added to kitchen-cloudstack and we need to accomodate please feel free to open a new PR/issue.

PR #740, opened 2015-06-10
add --fail-fast option

peterzhao:
When there are many test suites running in concurrency mode, sometimes we need to break the kitchen tests immediately if any of them fails without waiting for others to finish. Using --fail_fast option with --concurrency will make the failure fast to save time.


cheeseplus:
If you could rebase this I'd be happy to give it another look

lamont-granquist:
doesn't look like a bad idea but needs a rebase now.

tas50:
@peterzhao Thanks for this contribution. We've rebased it and merged it.

PR #743, opened 2015-06-12
Omnibus script cleanup

glasschef:
Files left over by the Omnibus installation script can cause ownership-related problems for subsequent runs if they were initially created as another user.


lamont-granquist:
i thought we converted to mixlib-install and chef_base_install_command.sh should no longer be used?

cheeseplus:
Closing this one as mixlib-install should now be used and we can probably kill the file in question with fire.

PR #758, opened 2015-06-25
Test fails under ChefDK

tfitch:
When running `rake default` a test in cucumber was failing to match the string `kc(Kitchen::Config)>`.  Thankfully the test does successfully match the computed values of `["default-flebian", "full-flebian"]` which is great, because it ensures the creation of those values is working.

So to get the `rake default` command to run without error, I've removed the line with the attempt to string match `kc(Kitchen::Config)>` and everything ran to completion without error.

### DIFF
--- a/features/kitchen_console_command.feature
+++ b/features/kitchen_console_command.feature
@@ -26,7 +26,6 @@ Feature: Running a console command
     When I run `kitchen console` interactively
     And I type "instances.map { |i| i.name }"
     And I type "exit"
-    Then the output should contain "kc(Kitchen::Config)> "
     Then the output should contain:
     """
     ["default-flebian", "full-flebian"]


cheeseplus:
I can't repro this anymore - is this still an issue?

tfitch:
@cheeseplus well, I just checked out the repo in to a clean directory with latest from master.  Ran `bundle install` and then `bundle exec rake default` and got the output

```
Finished in 6.065131s, 358.2775 runs/s, 548.0508 assertions/s.

2173 runs, 3324 assertions, 0 failures, 0 errors, 0 skips

/opt/chefdk/embedded/bin/ruby -S bundle exec cucumber features -x --format progress --no-color
........................................................................................................F--

(::) failed steps (::)

expected "\e[0G=> [\"default-flebian\", \"full-flebian\"]" to string includes: "kc(Kitchen::Config)> " (RSpec::Expectations::ExpectationNotMetError)
features/kitchen_console_command.feature:29:in `Then the output should contain "kc(Kitchen::Config)> "'

Failing Scenarios:
cucumber features/kitchen_console_command.feature:25 # Scenario: Launching a session

13 scenarios (1 failed, 12 passed)
107 steps (1 failed, 2 skipped, 104 passed)
0m29.765s
```

Unit tests were fine, still the same issue with Cucumber.  I'm using the Ruby from the Chef DK version 0.10.0.  I'm thinking that's the root of the difference if the failure doesn't happen for you.

cheeseplus:
Ah! That is something we've seen with ChefDK Ruby.

PR #768, opened 2015-07-08
Fix monkey patching of IO.read

375gnu:
Ruby 2.0 supports one more optional parameter of IO.read, which is hash of read modes.


fnichol:
@375gnu This looks good--think you could fix up the Rubocop failures from the Travis results? You can run `bundle exec rake style` to get these in development. Thanks!

375gnu:
I've pushed fixed patch

fnichol:
Excellent, thank you!

PR #689, opened 2015-04-15
Adding appveyor config

tyler-ball:
\cc @jdmundrawala @fnichol @schisamo

PR #776, opened 2015-07-13
Add more options for WinRM

smurawski:
Uses winrm-s to allow sspinegotiate auth from Windows hosts to Windows guests.

Adds winrm_transport to the configuration options for the winrm transport with defaults to plaintext (the current behavior) and can be set to sspinegotiate.

The effect is that we don't have to have AllowUnencypted set to $true for the WinRM service on the target machines.

This won't help drivers that have their own reliance on WinRM as a communication option (like kitchen-vagrant), but any driver that uses the transport abstraction can take advantage of this.

### DIFF
--- a/lib/kitchen/transport/winrm.rb
+++ b/lib/kitchen/transport/winrm.rb
@@ -24,9 +24,7 @@ require "uri"
 require "kitchen"
 
 module Kitchen
-
   module Transport
-
     # Wrapped exception for any internally raised WinRM-related errors.
     #
     # @author Fletcher Nichol <fnichol@nichol.ca>
@@ -38,19 +36,43 @@ module Kitchen
     # @author Salim Afiune <salim@afiunemaya.com.mx>
     # @author Fletcher Nichol <fnichol@nichol.ca>
     class Winrm < Kitchen::Transport::Base
-
       kitchen_transport_api_version 1
 
       plugin_version Kitchen::VERSION
 
-      default_config :port, 5985
       default_config :username, "administrator"
       default_config :password, nil
-      default_config :endpoint_template, "http://%{hostname}:%{port}/wsman"
       default_config :rdp_port, 3389
       default_config :connection_retries, 5
       default_config :connection_retry_sleep, 1
       default_config :max_wait_until_ready, 600
+      default_config :winrm_transport, "plaintext"
+      default_config :port do |transport|
+        transport[:winrm_transport] == :ssl ? 5986 : 5985
+      end
+      default_config :endpoint_template do |transport|
+        scheme = transport[:winrm_transport] == :ssl ? "https" : "http"
+        "#{scheme}://%{hostname}:%{port}/wsman"
+      end
+
+      def finalize_config!(instance)
+        super
+
+        transport = config[:winrm_transport].to_sym
+        config[:winrm_transport] =
+         case transport
+         when :sspinegotiate
+           if host_os_windows?
+             transport
+           else
+             :plaintext
+           end
+         else
+           transport
+         end
+
+        self
+      end
 
       # (see Base#connection)
       def connection(state, &block)
@@ -70,7 +92,6 @@ module Kitchen
       #
       # @author Fletcher Nichol <fnichol@nichol.ca>
       class Connection < Kitchen::Transport::Base::Connection
-
         # (see Base::Connection#close)
         def close
           return if @session.nil?
@@ -112,7 +133,7 @@ module Kitchen
           when /linux/
             login_command_for_linux
           else
-            raise ActionFailed, "Remote login not supported in #{self.class} " \
+            fail ActionFailed, "Remote login not supported in #{self.class} " \
               "from host OS '#{RbConfig::CONFIG["host_os"]}'."
           end
         end
@@ -126,9 +147,9 @@ module Kitchen
         def wait_until_ready
           delay = 3
           session(
-            :retries  => max_wait_until_ready / delay,
-            :delay    => delay,
-            :message  => "Waiting for WinRM service on #{endpoint}, " \
+            :retries => max_wait_until_ready / delay,
+            :delay => delay,
+            :message => "Waiting for WinRM service on #{endpoint}, " \
               "retrying in #{delay} seconds"
           )
           execute(PING_COMMAND.dup)
@@ -299,9 +320,9 @@ module Kitchen
         # @return [LoginCommand] a login command
         # @api private
         def login_command_for_linux
-          args  = %W[ -u #{options[:user]} ]
-          args += %W[ -p #{options[:pass]} ] if options.key?(:pass)
-          args += %W[ #{URI.parse(endpoint).host}:#{rdp_port} ]
+          args  = %W[-u #{options[:user]}]
+          args += %W[-p #{options[:pass]}] if options.key?(:pass)
+          args += %W[#{URI.parse(endpoint).host}:#{rdp_port}]
 
           LoginCommand.new("rdesktop", args)
         end
@@ -373,7 +394,7 @@ module Kitchen
         def session(retry_options = {})
           @session ||= establish_shell({
             :retries => connection_retries.to_i,
-            :delay   => connection_retry_sleep.to_i
+            :delay => connection_retry_sleep.to_i
           }.merge(retry_options))
         end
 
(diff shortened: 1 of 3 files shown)


smurawski:
@tyler-ball @fnichol test failure looks unrelated to my change.  Thoughts?

ksubrama:
Hey, I don't see a winrm-s dependency anywhere in the gemspec (and I only see a winrm development dependency there).  How does winrm-s get pulled in?

zl4bv:
It looks like winrm-s is bundled with chefdk, but that's no help if you're running test-kitchen some other way.

lib/kitchen/transport/winrm.rb
```diff
@@ -93,14 +104,14 @@ def execute(command)
             log_stderr_on_warn(stderr)
           elsif exit_code != 0
             log_stderr_on_warn(stderr)
-            raise Transport::WinrmFailed,
-              "WinRM exited (#{exit_code}) for command: [#{command}]"
+            fail Transport::WinrmFailed,
```
tyler-ball: Why this change?

lib/kitchen/transport/winrm.rb
```diff
@@ -122,9 +133,9 @@ def upload(locals, remote)
         def wait_until_ready
           delay = 3
           session(
-            :retries  => max_wait_until_ready / delay,
-            :delay    => delay,
-            :message  => "Waiting for WinRM service on #{endpoint}, " \
+            retries: max_wait_until_ready / delay,
```
tyler-ball: I actually don't think we'll be able to make this change - some people still use TK with Ruby 1.9

tyler-ball:
I'm generally :+1: on this.

I'm not sure what the test failures are...  Can you rebase off the latest master?

smurawski:
@tyler-ball Yeah, the test failures is what has sent me down the route of making the tests run on Windows, so this is blocked on https://github.com/test-kitchen/test-kitchen/pull/689

I'm also adding SSL support, since we are enabling ssl for winrm in knife and provisioning.

lib/kitchen/transport/winrm.rb
smurawski: I don't remember, I think rubocop made that change or suggested it.. I'll revert it.

lib/kitchen/transport/winrm.rb
smurawski: I can flip the syntax back, but due to the aruba update, we can only test on 2.x and newer.

lib/kitchen/transport/winrm.rb
```diff
@@ -38,22 +36,36 @@ class WinrmFailed < TransportFailed; end
...
       # (see Base#connection)
       def connection(state, &block)
+        case config[:winrm_transport]
+        when "ssl"
+          config[:winrm_transport] = :ssl
+        when "sspinegotiate" && host_os_windows?
+          config[:winrm_transport] = :sspinegotiate
+        else
+          config[:winrm_transport] = :plaintext
+        end
```
fnichol: This block would work better in a `def finalize_config!(instance)` method, similar to https://github.com/test-kitchen/test-kitchen/blob/621cd45784c3aed0e893d390c2758125b250151a/lib/kitchen/transport/ssh.rb#L62-L75 (note the `super` call there is pretty important :smile:)

The reason is that `kitchen diagnose` returns its configuration after the `#finalize_config!` has done any last minute tweaks, meaning that the final computed value of `config[:winrm_transport]` would show in the diagnostic output.

lib/kitchen/transport/winrm.rb
```diff
@@ -452,8 +477,29 @@ def load_needed_dependencies! # rubocop:disable Metrics/MethodLength
...
+        case RbConfig::CONFIG["host_os"]
+        when /darwin/
+          false
+        when /mswin|msys|mingw|cygwin|bccwin|wince|emc/
+          true
+        when /linux/
+          false
+        else
+          false
+        end
```
fnichol: I'd say you're okay with the super simple case of:

```ruby
case RbConfig::CONFIG["host_os"]
when /mswin|msys|mingw|cygwin|bccwin|wince|emc/
  true
else
  false
end
```

That way we're focused on the whilelisting tokens and dropping anything else by default.

fnichol: Now I see the code block this code was taken from :)

lib/kitchen/transport/winrm.rb
```diff
@@ -452,8 +477,29 @@ def load_needed_dependencies! # rubocop:disable Metrics/MethodLength
...
+      def sspi_enabled?
+        enabled = host_os_windows? &&
+          config[:winrm_transport] == :sspinegotiate
+        logger.debug("Checking for Negotiate Auth: Enabled = #{enabled}")
+        enabled
+      end
+
       def load_winrm_transport!
         silence_warnings { require "winrm" }
+        silence_warnings { require "winrm-s" } if sspi_enabled?
```
fnichol: I think we're going to need a little load/rescue code for this gem, similar in spirit to [#load_needed_dependencies!](https://github.com/test-kitchen/test-kitchen/blob/b73b8fbccaf84242efb3c05502e36557c963c83a/lib/kitchen/transport/winrm.rb#L426-L450) to handle the case where we want the gem, try to require it, but the gem can't be activated (primarily this for outside ChefDK, but still a valid failure case). I was trying to see where best to handle this, and am currently thinking a short method with the require and `rescue LoadError` blocks? Happy to chat more about this one--I think it's the first case we have of augmenting behavior with an addition gem. Can't think of another place to look for stealing ideas!

fnichol:
Other the inline comments above, any reason we can't add some tests to cover the data-in/config out behavior? Happy to help with ideas, direction, pairing, or code. I'm pretty sure we (finally) have enough examples to draw on this seems doable, woo!

smurawski:
@fnichol @tyler-ball @randomcamel It took me a bit to get back to this, but I've got the code refactored and restructured to address the concerns mentioned above.  I've also added tests to cover the configuration options and conditional loading of winrm-s.

andrewelizondo:
Any update on getting this merged? @fnichol @tyler-ball @randomcamel

lib/kitchen/transport/winrm.rb
```diff
@@ -38,19 +36,43 @@ class WinrmFailed < TransportFailed; end
...
 
-      default_config :port, 5985
       default_config :username, "administrator"
       default_config :password, nil
-      default_config :endpoint_template, "http://%{hostname}:%{port}/wsman"
       default_config :rdp_port, 3389
       default_config :connection_retries, 5
       default_config :connection_retry_sleep, 1
       default_config :max_wait_until_ready, 600
+      default_config :winrm_transport, "plaintext"
```
mwrock: default to sspinegotiate on windows hosts

lib/kitchen/transport/winrm.rb
```diff
@@ -389,24 +410,38 @@ def to_s
...
+          :winrm_transport => data[:winrm_transport]
         }
-
+        opts.merge!(additional_transport_args(opts[:winrm_transport]))
         opts
       end
 
+      def additional_transport_args(transport_type)
+        case transport_type
+        when :ssl, :sspinegotiate
```
mwrock: I suggest we match the logic in knife-windows 1.x:
`:disable_sspi` if not on windows, using `ssl, or not not authenticating with `negotiate`
`:basic_auth_only` only from linux when ssl is not specified

lib/kitchen/transport/winrm.rb
```diff
@@ -424,41 +459,90 @@ def create_new_connection(options, &block)
...
-        end
-      rescue LoadError => e
-        logger.fatal("The `winrm-transport' gem is missing and must" \
-          " be installed or cannot be properly activated. Run" \
-          " `gem install winrm-transport --version '#{spec_version}'`" \
-          " or add the following to your Gemfile if you are using Bundler:" \
-          " `gem 'winrm-transport', '#{spec_version}'`.")
-        raise UserError,
-          "Could not load or activate WinRM::Transport (#{e.message})"
+        load_winrm_s! if host_os_windows?
```
mwrock: We probably should not load winrm-s if using `ssl`

mwrock:
I'm totally :+1: on adding these winrm options but we should try to mach the defaults currently in use by knife-windows

smurawski:
@mwrock I'm open to matching the defaults in knife-windows, but that'll be a breaking change. This PR keeps the same default behavior but supports additional options which I think is better for a minor release.

mwrock:
Yeah I agree we don't want this to be a breaking change. I wasn't thinking that would be breaking though. Which scenario does this break?

smurawski:
If the defaults were changed to the same as knife-windows, any cross-platform connections would fail out of the box (since all previous kitchen connections would have been basic/unencrypted).  From Windows, we'd be ok because winrm-s.

mwrock:
Ah right we throw an error now asking the user to be explicit about basic auth. So I'd say keep the knife-windows strategy EXCEPT for linux hosts default to `basic_auth`

smurawski:
@tyler-ball This PR should be good to go.  The test failures are the same as the test failures against master.

Issue #779, opened 2015-07-15
create_sandbox attempts to use cyclical path when /tmp does not have correct permissions


rhass:
When /tmp has incorrect permissions, (eg: missing temp sticky bit in mode 0777 vs 1777) the mktmpdir call attempts to create tempdir which exists within the `kitchen_root` path resulting in a cyclical path when the during the "Prepairing cookbooks" phase.

This is an example of such issue:

```
-----> Starting Kitchen (v1.4.0)
D      [kitchen::driver::docker command] BEGIN (sudo -E docker >> /dev/null 2>&1)
D      [kitchen::driver::docker command] END (0m0.03s)
-----> Converging <default-ubuntu-1404>...
$$$$$$ Running legacy converge for 'Docker' Driver
       Preparing files for transfer
D      Creating local sandbox in /home/rhass/src/cookbooks/ratchet/default-ubuntu-1404-sandbox-20150714-13752-a7vopg
       Preparing dna.json
D      Creating dna.json from {:run_list=>["recipe[ratchet]"]}
       Preparing current project directory as a cookbook
D      Using metadata.rb from /home/rhass/src/cookbooks/ratchet/metadata.rb
D      Cleaning up local sandbox in /home/rhass/src/cookbooks/ratchet/default-ubuntu-1404-sandbox-20150714-13752-a7vopg
>>>>>> ------Exception-------
>>>>>> Class: Kitchen::ActionFailed
>>>>>> Message: Failed to complete #converge action: [cannot copy directory /home/rhass/src/cookbooks/ratchet/default-ubuntu-1404-sandbox-20150714-13752-a7vopg to itself /home/rhass/src/cookbooks/ratchet/default-ubuntu-1404-sandbox-20150714-13752-a7vopg/cookbooks/ratchet/default-ubuntu-1404-sandbox-20150714-13752-a7vopg]
>>>>>> ----------------------
>>>>>> Please see .kitchen/logs/kitchen.log for more details
>>>>>> Also try running `kitchen diagnose --all` for configuration

D      ------Exception-------
D      Class: Kitchen::ActionFailed
[...]

PR #781, opened 2015-07-15
Ensure sandbox_path is not within kitchen_root.

rhass:
# Summary of Change

In rare cases the system temp directory becomes unusable, which results
in the mktmpdir trying to use the current working directory. This
becomes an issue as the default value for kitchen_root is also cwd
which results in a confusing cyclical error in which it attempts to copy
the cookbooks into a temp directory within the path which is is
copying. This should avoid most instances of this error, and present a
possible solution in simple terms.
# Related Issues

Resolves Issue #779
# Validation
- [ ] `sudo chmod 0777 /tmp`
- [ ] Attempt to run `kitchen converge` on a cookbook, it should raise a ClientError with the message, "There was a problem attempting to write to your default system temp directory. Check the permissions, and on non-windows hosts ensure the tmp sticky bit is set on the temp directory."
- [ ] `sudo chmod 1777 /tmp'
- [ ] Attempt to run `kitchen converge` on a cookbook, everything should work normally.

### DIFF
--- a/lib/kitchen/provisioner/base.rb
+++ b/lib/kitchen/provisioner/base.rb
@@ -141,7 +141,18 @@ module Kitchen
       # @return [String] the absolute path to the sandbox directory
       # @raise [ClientError] if the sandbox directory has no yet been created
       #   by calling `#create_sandbox`
+      # @raise [ClientError] if the system temp directory has incorrect
+      #   permissions causing the mktempdir call to try to use config[:kitchen_root]
+      #   instead.
       def sandbox_path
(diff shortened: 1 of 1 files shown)


rhass:
Looks like I broke a unit test with this change. Unless I move the logic somewhere else, I am not entirely sure how to fix the test as it is intentionally trying to validate sandbox_path is nil, where my code expects it to be defined.

lamont-granquist:
this is fine, but needs a rebase, and its pretty edge case

PR #782, opened 2015-07-15
Use mixlib-install

thommay:
This is essentially a refactor with zero functional changes, moving
functionality from test-kitchen to a separate gem, mixlib-install. This
allows us to DRY up a bunch of projects.

### DIFF
--- a/lib/kitchen/provisioner/chef_base.rb
+++ b/lib/kitchen/provisioner/chef_base.rb
@@ -25,6 +25,7 @@ require "kitchen/provisioner/chef/berkshelf"
 require "kitchen/provisioner/chef/common_sandbox"
 require "kitchen/provisioner/chef/librarian"
 require "kitchen/util"
+require "mixlib/install"
 
 module Kitchen
 
@@ -46,18 +47,6 @@ module Kitchen
         attributes/**/* definitions/**/* files/**/* libraries/**/*
         providers/**/* recipes/**/* resources/**/* templates/**/*
       ].join(",")
-
-      default_config :chef_metadata_url do |provisioner|
-        provisioner.default_windows_chef_metadata_url if provisioner.windows_os?
-      end
-
-      default_config :chef_omnibus_root do |provisioner|
-        if provisioner.windows_os?
-          "$env:systemdrive\\opscode\\chef"
-        else
-          "/opt/chef"
-        end
-      end
 
       default_config :data_path do |provisioner|
         provisioner.calculate_path("data")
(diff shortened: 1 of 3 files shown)


thommay:
@fnichol @mcquin @tyler-ball anything I need to do for this?
The travis errors are all 1.9.3; do we want to continue to support 1.9.3 in t-k, in which case i'll change the ruby version requirement in `mixlib-install`, or should we just drop those tests?

tyler-ball:
I think you should be mocking the mixlib commands in the spec.  Something like

```ruby
let(:installer) { "bash install blah..." }
expect(Mixlib::Install).to receive(:new).with(version, powershell_shell?, install_options).and_return(installer)
```

tyler-ball:
@thommay I think you can do
```ruby
let(:installer) { "bash install blah..." }
Mixlib::Install.expects(:new).with(version, powershell_shell?, install_options).returns(installer)
```

thommay:
@tyler-ball seems you can `stub` but not `expect` like that in minitest, but stubbing these tests makes the whole exercise seem rather pointless to me. I'll `mochito` them up.

thommay:
@tyler-ball besides the nasty warning about initialize, which there doesn't seem to be much way to avoid, these tests are all mocked up and good to go. I've also bumped the version of mixlib-install to 0.6.0, which installs on 1.9.3

lib/kitchen/provisioner/chef_base.rb
```diff
@@ -47,18 +48,6 @@ class ChefBase < Base
         providers/**/* recipes/**/* resources/**/* templates/**/*
       ].join(",")
 
-      default_config :chef_metadata_url do |provisioner|
```
tyler-ball: Is the `chef_metadata_url` now handled entirely in mixlib-install?  I wouldn't think this would need to be configurable....

lib/kitchen/provisioner/chef_base.rb
```diff
@@ -47,18 +48,6 @@ class ChefBase < Base
...
-        provisioner.default_windows_chef_metadata_url if provisioner.windows_os?
-      end
-
-      default_config :chef_omnibus_root do |provisioner|
```
tyler-ball: If a user needs to override their omnibus install location, how do they do it now?  Is this supported anymore?

lib/kitchen/provisioner/chef_base.rb
thommay: yeah, although it's generated from `chef_omnibus_url`, so one can set that and both windows and unixes will do the right thing.

lib/kitchen/provisioner/chef_base.rb
thommay: It's supported in mixlib-install but not implemented in test-kitchen right now; I'll make it so.

thommay:
@tyler-ball added the ability to set the install root.

tyler-ball:
Thanks for all the work on this @thommay !

PR #741, opened 2015-06-10
Add shell verifier

sawanoboly:
Hi, I've created shell verifier. It's useful for run [infrataster](http://infrataster.net) by kitchen.

Features:

- exec shell command locally by shellout.
- fail when its exit_code `!==0`
- pass state via env, such as KITCHEN_HOSTNAME, KITCHEN_PORT, etc...
- option `:remote_exec`: run command on instance.

Configration example

```
platforms:
  - name: ubuntu-12.04
  - name: ubuntu-14.04

verifier:
  name: shell
  command: bundle exec rake
  sleep: 5
```

And sample rake

```
task :default do
  puts "Hello #{ENV['KITCHEN_HOSTNAME']}:#{ENV['KITCHEN_PORT']}"
end
```

Results

```
$ kitchen verify 
-----> Starting Kitchen (v1.4.1.dev)
-----> Verifying <default-ubuntu-1204>...
       [Shell] Verify on instance=#<Kitchen::Instance:0x007fd80e111ac8> with state={:hostname=>"127.0.0.1", :port=>"2200", :username=>"vagrant", :ssh_key=>"...", :last_action=>"verify"}
.....
Hello 127.0.0.1:2200
       Finished verifying <default-ubuntu-1204> (0m5.66s).
-----> Verifying <default-ubuntu-1404>...
       [Shell] Verify on instance=#<Kitchen::Instance:0x007fd80c41c148> with state={:hostname=>"127.0.0.1", :port=>"2222", :username=>"vagrant", :ssh_key=>"...", :last_action=>"verify"}
.....
Hello 127.0.0.1:2222
       Finished verifying <default-ubuntu-1404> (0m5.67s).
-----> Kitchen is finished. (0m11.59s)
```

PR #783, opened 2015-07-16
Ability to SCP download a file or directory from instance to host

ytsarev:
Closest use case is to get junit compatible report from test suite
and feed it to Jenkins

### DIFF
--- a/lib/kitchen/driver/ssh_base.rb
+++ b/lib/kitchen/driver/ssh_base.rb
@@ -102,7 +102,7 @@ module Kitchen
       end
 
       # (see Base#verify)
-      def verify(state) # rubocop:disable Metrics/AbcSize
+      def verify(state) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
         verifier = instance.verifier
         verifier.create_sandbox
         sandbox_dirs = Dir.glob(File.join(verifier.sandbox_path, "*"))
@@ -118,6 +118,17 @@ module Kitchen
       rescue Kitchen::Transport::TransportFailed => ex
         raise ActionFailed, ex.message
       ensure
+        if verifier[:scp_file]
+          instance.transport.connection(backcompat_merged_state(state)) do |conn|
+            info "Transferring #{verifier[:scp_file][:source]} from instance" \
+              " to #{verifier[:scp_file][:dest]} on host"
+            conn.download(
+              verifier[:scp_file][:source],
+              verifier[:scp_file][:dest],
+              verifier[:scp_file][:recursive])
+            debug("Transfer complete")
+          end
+        end
         instance.verifier.cleanup_sandbox
       end
 
--- a/lib/kitchen/transport/base.rb
+++ b/lib/kitchen/transport/base.rb
@@ -160,6 +160,16 @@ module Kitchen
           raise ClientError, "#{self.class}#upload must be implemented"
         end
 
+        # Dowload files form intstance under test to local host
+        #
+        # @param remote [String] path to remote files or directories
+        # @param local [String] path to local destination
+        # @param recursive [Boolean] if the remote parameter refers to a remote
+        #   directory copy it recursively to local
+        def download(remote, local, recursive = nil) # rubocop:disable Lint/UnusedMethodArgument
+          raise ClientError, "#{self.class}#download must be implemented"
+        end
+
         # Block and return only when the remote host is prepared and ready to
         # execute command and upload files. The semantics and details will
         # vary by implementation, but a round trip through the hosted
--- a/lib/kitchen/transport/ssh.rb
+++ b/lib/kitchen/transport/ssh.rb
@@ -170,6 +170,14 @@ module Kitchen
           raise SshFailed, "SCP upload failed (#{ex.message})"
         end
 
+        def download(remote, local, recursive = nil)
+          opts = recursive ? { :recursive => true } : {}
+          session.scp.download!(remote, local, opts)
+          logger.debug("Downloaded #{remote} to #{local}")
+        rescue Net::SSH::Exception => ex
+          raise SshFailed, "SCP download failed (#{ex.message})"
+        end
+
         # (see Base::Connection#wait_until_ready)
         def wait_until_ready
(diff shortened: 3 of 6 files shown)


ytsarev:
@fnichol  @tyler-ball Review anybody ?

lib/kitchen/driver/ssh_base.rb
```diff
@@ -115,6 +115,14 @@ def verify(state) # rubocop:disable Metrics/AbcSize
       rescue Kitchen::Transport::TransportFailed => ex
         raise ActionFailed, ex.message
       ensure
+        if verifier[:scp_file]
```
tyler-ball: Why did you decide to put this in the `ensure` block?  I would think you would want it to go in the existing connection block.

tyler-ball: There should also be some debug logging after the download finishes

tyler-ball:
I think `download` and its specs look good.  Can you update https://github.com/test-kitchen/test-kitchen/blob/master/spec/kitchen/driver/ssh_base_spec.rb to add coverage to the changes you made in `ssh_base.rb`?

lib/kitchen/driver/ssh_base.rb
ytsarev: In fact I want to download junit.xml with detailed report of serverspec failures.

ytsarev:
@tyler-ball i've addressed the issues you pointed out, could you review again ?

lib/kitchen/driver/ssh_base.rb
tyler-ball: I still don't think this needs to be in an `ensure` block.  If your serverspec command fails, does it raise an exception?  If so, which one?

There are already 4 commands being run on the remote host in the `begin`/`rescue` block.  If any of those 4 fail except the serverspec command then the `download` command below will also fail.  That could mask a real failure by raising, instead, a failure from the `download` command.

Does my concern make sense?

tyler-ball:
@ytsarev Thanks for the update specs.  I only had 1 other comment that I will wait for your feedback on

lib/kitchen/driver/ssh_base.rb
ytsarev: Concern are very valid.

However I'm not sure how to proceed. My use case is the use of shell verifier( https://github.com/test-kitchen/test-kitchen/pull/741) which runs wrapper around serverspec test suite. 

If the wrapper returns non-zero exit status it raises ssh TransportFailed ...
This behavior aslo relates to ssh transport design.

So in fact I'm not sure how to proceed with moving it out of `ensure` and avoid shotgun surgery of ssh transport.

Do you have any idea ?

tyler-ball: Paging @fnichol to get his opinion

ytsarev: @fnichol I will be grateful for review here

cliles:
So is this good to merge? +1 on the use case, just what I am trying to do.

ytsarev:
@tyler-ball @fnichol please decide on this one

tas50:
@fnichol This would be hugely useful in some work I'm noodling on with automated Ohai testing and I know @lamont-granquist had mentioned the need for this same sort of feature for omnibus work.

thommay:
supercedes #130 i guess :) be good to see this fixed.

yyuu:
:+1:

Now I'm doing dirty `instance_eval` to do this kind of trick....

ytsarev:
@tyler-ball @fnichol so Christmas present anybody ?

cheeseplus:
So we totally want to get this one in and given that WinRM needs more work we don't have to gate on that. That said could we get two things?

1)  raise NotImplementedError in the base transport to if you are on WinRM
2)  have a command, ex `kitchen download foo.txt` wired up

ytsarev:
@cheeseplus alright, thanks, sounds like a plan. Will ping you back when implementation ready.

cheeseplus:
@ytsarev awesome and if you need any help or get stuck feel free to holler

ytsarev:
@cheeseplus Sorry, it's been a while, finally returned to this one
* I've implemented 1), please review. 
* I believe that 2) here is irrelevant and should be implemented as separate additional feature as we have no symmetric upload command
What do you think ?

ytsarev:
@cheeseplus ping

KierranM:
Given that `winrm-fs` already has the functionality to download files it would be great to add this to winrm transport as well, I have essentially the same use case as @ytsarev of getting Pester results into TeamCity.

From what I can gather it would only involve adding:
```ruby
# (see Base::Connection#download)
def download(remote, local)
  file_transporter.download(remote, local)
end
```
to the WinRM transport.

Is there a reason, @cheeseplus, that you believe it would require enough work to not be included?

mwrock:
adding winrm compatibility should be trivial but it could certainly be added in a separate PR. I'd hate to see it block this. From what I can tell the only open item here appears to be implementation details around the `ensure` block. Those concerns do look valid.

My biggest concern there is that if an error occurs in one of the previous commands and then the ensure block also fails, we lose the original error.

mwrock:
Given we want download to be transport agnostic, we should use different setting names that don't contain `scp`

lib/kitchen/driver/ssh_base.rb
AgarFu: I'd love tho have this feature merged, looks like the the discussion about how to solve the rescue thing is not going anywhere. I do understand the concerns there is no an easy way to solve this issue.

Maybe to add as much as detail to the ActionFailed and SshFailed exceptions as possible to help the user understand the reason why the process failed is the only solution without change the ssh transport design.

IMHO this feature shouldn't be blocked any more time by this issue, I think this is super important to help identify errors in executions of more tan 3 or 4 tests suites in parallel. With the current implementation try to find the error in the log when you're running 15 or 16 parallel suites is a real pain.

👍

cheeseplus:
Closing in favor of https://github.com/test-kitchen/test-kitchen/pull/1306

PR #793, opened 2015-07-27
Nightlies

jaym:
This allows specifying nightlies in the following manner:
```yaml
provisioner:
  require_chef_omnibus: nightly
```
or
```yaml
provisioner:
  nightly: true
```
or
```yaml
provisioner:
  require_chef_omnibus: 12.4
  nightly: true
```

### DIFF
--- a/lib/kitchen/provisioner/chef_base.rb
+++ b/lib/kitchen/provisioner/chef_base.rb
@@ -262,6 +262,9 @@ module Kitchen
       # @api private
       def install_command_vars_for_bourne(version)
         install_flags = %w[latest true].include?(version) ? "" : "-v #{CGI.escape(version)}"
+        if config[:nightly]
+          install_flags << " " << "-n"
+        end
         if config[:chef_omnibus_install_options]
           install_flags << " " << config[:chef_omnibus_install_options]
         end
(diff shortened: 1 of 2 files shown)


smurawski:
:+1:

tyler-ball:
@jaym can you rebase this off https://github.com/test-kitchen/test-kitchen/pull/782 ? Its going to change how we install chef

jaym:
i think #782 will need to rebase against master first.

thommay:
@jaym #782 ought to be up to date (and if you just pass `opts['nightlies']=true` it should just work with this branch)

ksubrama:
I and @smurawski just walked through this.  You're right.  Using chef_omnibus_install_options: "-n" should now handle this scenario.

jaym:
I think this is much nicer than than `chef_omnibus_install_options: "-n"`

tyler-ball:
I like having 1 install library (MixLib::Install) and I worry that adding multiple, different APIs will lead to complexity for users and us (as authors).  That said, I could easily be swayed to support `nightly: true`.  Or perhaps we could introduce something like

```yaml
provisioner:
  chef_install:
    version: 12.4.1
    nightly: true
    root: /var/chef
```

Where all the keys we pass to `chef_install` are the options we pass to `MixLib::Install`.  Thoughts?

ksubrama:
+1 If we made a separate element that just acted as a pass through, that would indeed look quite good.  The corresponding options in chef_omnibus_* or the root or url can then be deprecated.

PR #801, opened 2015-08-03
Fix for net-ssh 2.10.0.

### LINKED ISSUE #729: Possible bug in Getting Started Guide: 'could not settle on compression_client algorithm'
I'm following the instructions in [Getting Started Guide: Creating a Cookbook](http://kitchen.ci/docs/getting-started/creating-cookbook) and am getting errors while running ```kitchen create default-ubuntu-1204```, I receive the following error:

```
-----> Starting Kitchen (v1.4.0)
-----> Creating <default-ubuntu-1204>...
       Bringing machine 'default' up with 'virtualbox' provider...
       ==> default: VirtualBox VM is already running.
>>>>>> ------Exception-------
>>>>>> Message: Failed to complete #create action: [could not settle on compression_client algorithm]
>>>>>> ----------------------
```

I'm not sure what's happening exactly, but it seems that the guest machine isn't ready to use zlib compression on incoming ssh connections out of the box. When I changed the default compression from "zlib" to "none" in kitchen/transport/ssh.rb:55, the problem went away.

Since this is so early in the tutorial, and test-kitchen manages the setup of the vagrant guest, I'm assuming this is either a default configuration error, or there's a missing step in the guide.

coderanger:
This sets the default compression to auto-negotiate the best available. This is broken currently because net-ssh no longer silently allows falling back to no compression and OpenSSH uses `"zlib@openssh.com"`, not `"zlib"`. I've removed the validation as it is incorrect and in the future it would be nice to not need an ASAP release for this kind of issue. If a user misconfigured the compression setting, they will now notice immediately because it will refuse to connect.

Fixes #729.

### DIFF
--- a/lib/kitchen/transport/ssh.rb
+++ b/lib/kitchen/transport/ssh.rb
@@ -52,13 +52,8 @@ module Kitchen
       default_config :ssh_key, nil
       expand_path_for :ssh_key
 
-      default_config :compression, "zlib"
-      required_config :compression do |attr, value, transport|
-        if !%W[zlib none].include?(value)
-          raise UserError, "#{transport} :#{attr} value may only " \
-            "be set to `none' or `zlib'."
-        end
-      end
+      default_config :compression, true
+      required_config :compression
 
       default_config :compression_level do |transport|
         transport[:compression] == "none" ? 0 : 6
--- a/spec/kitchen/transport/ssh_spec.rb
+++ b/spec/kitchen/transport/ssh_spec.rb
@@ -145,21 +145,14 @@ describe Kitchen::Transport::Ssh do
       transport[:username].must_equal "root"
     end
 
-    it "sets :compression to zlib by default" do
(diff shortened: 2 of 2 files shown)


coderanger:
Side effect of this is that something is funky with the use of zlib in net-ssh. `"zlib(finalizer): the stream was freed prematurely."` gets printed as the Ruby process exits. Guessing we might not be correctly closing the SSH connection? Regardless it is only a warning and doesn't cause any problems since the process exits afterwards.

tyler-ball:
@coderanger I added some backwards compatibility casts so we can release this as a patch version.  Thanks for this PR!

patcon:
Any pointers on what issue to track on the exit error? Would I look in zlib's bug tracker, or here? Thanks! :)

xmik:
Related issue about `"zlib(finalizer): the stream was freed prematurely."`: https://github.com/test-kitchen/test-kitchen/issues/822

PR #804, opened 2015-08-03
Drop Ruby 1.9.3 from TravisCI build matrix

### LINKED ISSUE #806: Drop Ruby 1.9 support
Then we could merge https://github.com/test-kitchen/test-kitchen/pull/804 and we could use the latest version of net-ssh

Tentatively setting this with milestone 1.5 but it is arguably a 2.0 breaking change.

Opening this so we can discuss it.

thommay:
Fixes #806


tyler-ball:
TK still supports Ruby 1.9.  I think it may only be because it was never deprecated, but @fnichol will have to weigh in here

ytsarev:
Well, we are running kitchen on scientific linux 6. 
I would not vote for dropping 1.9 support as it will be a pain for users of enterprise linux distributions

Peuserik:
I think ytrsarev is right. We are running multiple enterprise linux distributions and ruby repo version is 1.9.. Installing a newer version is as said, not a trivial task cause of missing dependencies etc..

tyler-ball:
@ytsarev and @Peuserik is it possible for you to use the ChefDK (which comes bundled with its own Ruby version) version of Test Kitchen?

fnichol:
I'd be okay moving to a "best effort" support of Ruby versions less than 2.0.0 as it wouldn't affect the gemspec. Bumping this hard requirement would be a major version bump, but that time is coming soon, I wager. Also, we're talking about a Ruby version present on your "workstation", not the remote instances, so I think it's an easier case to make.

fnichol:
@ytsarev, @Peuserik This should only affect systems which `gem install test-kitchen` (i.e. your workstation) which usually have a better chance of running a newer version of Ruby. Having said that, we wouldn't be able to drop 1.9 support explicitly until a major version bump of Test Kitchen which I suspect is a ways off. Sound reasonable?

ytsarev:
@fnichol my workstation is fine, hundreds of jenkis slaves are not :(

ytsarev:
@tyler-ball ChefDK can be suitable for workstation but not for CI infrastructure...

thommay:
@ytsarev can you explain more about why you think chefdk isn't suitable for CI?

ytsarev:
@thommay well, maybe suitable for people who testingh chef, but I'm using kitchen for puppet testing and don't want to bring bunch of unrelated stuff into dozens of machines

Peuserik:
@fnichol @thommay 
We are not testing chef as well... we are testing ansible roles and puppet manifest in jenkins ci for infrastructure testing.

Am i getting this right? Even if we are dropping ruby 1.9 support, i don't have to have a different version of ruby or something like chefdk on my test servers, right? I have to update or install a higher version of ruby on my jenkins servers only, right?

If this is true. I'm not exactly happy, but i think we could manage installing a different version of ruby or chefdk on our base images for spawning jenkins slaves. We would have to test this scenario .. but we have to test everything anytime anyway. ;)

if this is not true, dropping ruby 1.9 support would break the usability of test-kitchen for us. i have to test production like servers and environments. Installing software (not gems) not installed in production would make the tests useless.

tyler-ball:
Even if you are not testing chef, the ChefDK still gives you a Test Kitchen instances with an embedded Ruby that has been tested.  Its a larger package but nothing says you have to use the other features of the ChefDK (like chef, or berkshelf, etc.).  Are you worried that the additional code would interfere with some other running process?

If installing a higher version of ruby or installing the ChefDK aren't viable options, we could also create an omnibus package of Test Kitchen which has an embedded Ruby.  But I would rather not do that (and I don't think it would get on a roadmap anytime soon since the ChefDK is basically that).

We're also _not_ dropping Ruby support as part of this.  We're just dropping it from the test matrix and moving it to a `best effort` level of support.  It should continue working fine for a while, but the first time a 1.9 bug pops up there is a chance we won't fix it.

PR #805, opened 2015-08-03
Pinning net-ssh to 2.9

tyler-ball:
Because 2.10 only works with Ruby >= 2.0 and TK is still compatible with Ruby 1.9.3.

Going to release this as 1.4.2

\cc @coderanger @Annih @fnichol 

net-ssh bug - https://github.com/net-ssh/net-ssh/issues/260

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,12 @@
-## UNRELEASED
+## 1.4.2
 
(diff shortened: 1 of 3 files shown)


coderanger:
:+1: Can sort out if/when to do a 2.0 and drop this later.

PR #808, opened 2015-08-04
Add config site_cookbooks_path

hirocaster:
can chage `site_cookbooks` directory path.

in `.kitchen.yml`
```
  site_cookbooks_path: ./config/site-cookbooks
```

default is under the kitchen root

### DIFF
--- a/lib/kitchen/provisioner/chef/common_sandbox.rb
+++ b/lib/kitchen/provisioner/chef/common_sandbox.rb
@@ -297,10 +297,14 @@ module Kitchen
         end
 
         # @return [String] an absolute path to a site-cookbooks/ directory,
-        #   relative to the kitchen root
+        #   configured site_cookbooks_path(Default: relative to the kitchen root)
         # @api private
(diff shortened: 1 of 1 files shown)


cheeseplus:
I don't have any opposition to this one, will get someone else to review.

adamleff:
I'm OK with this concept, but I'd like to see the implementation cleaned up a small bit:

 * `@config` has an attr_reader setup, so let's just call `config` instead of `@config`
 * Many users rely on the `default_config` and `required_config` statements in these classes as the documentation as to what options exist.  Since you are introducing a new option, please add it to `chef_base.rb` with a default value of `nil` so users know it's an available option.

cheeseplus:
I feel like this is covered/superseded by https://github.com/test-kitchen/test-kitchen/pull/393

adamleff:
@cheeseplus I agree and think #393 is the right place to work this out. I'd recommend closing this in favor of #393.

PR #813, opened 2015-08-10
Honor proxy env vars.

mcquin:
Resolves [EASY-8](https://chefio.atlassian.net/browse/EASY-8): Test Kitchen should honor local environment proxy variables.

### DIFF
--- a/lib/kitchen/configurable.rb
+++ b/lib/kitchen/configurable.rb
@@ -295,15 +295,22 @@ module Kitchen
     # @param code [String] the shell code to be wrapped
     # @return [String] wrapped shell code
     # @api private
+    # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
     def wrap_shell_code(code)
       env = []
       if config[:http_proxy]
         env << shell_env_var("http_proxy", config[:http_proxy])
         env << shell_env_var("HTTP_PROXY", config[:http_proxy])
+      else
+        env << shell_env_var("http_proxy", ENV["http_proxy"]) if ENV["http_proxy"]
+        env << shell_env_var("HTTP_PROXY", ENV["HTTP_PROXY"]) if ENV["HTTP_PROXY"]
       end
       if config[:https_proxy]
         env << shell_env_var("https_proxy", config[:https_proxy])
         env << shell_env_var("HTTPS_PROXY", config[:https_proxy])
+      else
+        env << shell_env_var("https_proxy", ENV["https_proxy"]) if ENV["https_proxy"]
+        env << shell_env_var("HTTPS_PROXY", ENV["HTTPS_PROXY"]) if ENV["HTTPS_PROXY"]
       end
       if powershell_shell?
         env.join("\n").concat("\n").concat(code)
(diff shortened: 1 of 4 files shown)


tyler-ball:
I'm totally :+1: on this.  @jamescott can you update https://docs.chef.io/config_yml_kitchen.html#provisioner-settings to have the proxy settings say `The proxy server for HTTP connections.  This value will overwrite ENV['http_proxy']` and similarly for the https_proxy?

MarkGibbons:
Could there be some way to disable local proxy settings in .kitchen.yml?  My development environments usually have proxies set, my converge environments don't.   

Thanks,
   Mark

tyler-ball:
@MarkGibbons You could specify `http_proxy: ""` in your .kitchen.yml - because we favor the config file over the environment variables it will end up setting `HTTP_PROXY=""` in the environment.

I just did a quick test with curl and it didn't try and use an empty proxy, but I think that behavior is probably tool dependent.  Would that work for you?

MarkGibbons:
@tyler-ball That worked on a couple tests. Thanks

tyler-ball:
Actually, I thought of something we should merge - any time we add `http_proxy` to the environment we also need to add `no_proxy` if it is present.

mcquin:
Yep! I can do that.

mcquin:
@tyler-ball I've added support for including `no_proxy` settings from workstation environment, if `http_proxy` or `https_proxy` are set from the workstation environment.

tyler-ball:
Thanks @mcquin !

tacchino:
@tyler-ball Setting to empty strings fails in some scenarios. Maybe it would be a better solution to evaluate the value and if empty don't add to the environment at all?

Example failure:

```
Recipe: sensu::_linux
         * apt_repository[sensu] action add
           * remote_file[/tmp/kitchen/cache/pubkey.gpg] action create

             ================================================================================
             Error executing action `create` on resource 'remote_file[/tmp/kitchen/cache/pubkey.gpg]'
             ================================================================================

             NoMethodError
             -------------
             undefined method `host' for "":String

             Resource Declaration:
             ---------------------
             # In /tmp/kitchen/cache/cookbooks/apt/providers/repository.rb

       90:     remote_file cached_keyfile do
       91:       source new_resource.key
       92:       mode 00644
       93:       sensitive new_resource.sensitive if respond_to?(:sensitive)
       94:       action :create
       95:     end
       96:   else

             Compiled Resource:
             ------------------
             # Declared in /tmp/kitchen/cache/cookbooks/apt/providers/repository.rb:90:in `install_key_from_uri'

             remote_file("/tmp/kitchen/cache/pubkey.gpg") do
        provider Chef::Provider::RemoteFile
        action [:create]
        retries 0
        retry_delay 2
        default_guard_interpreter :default
        path "/tmp/kitchen/cache/pubkey.gpg"
        backup 5
        atomic_update true
        source ["http://repositories.sensuapp.org/apt/pubkey.gpg"]
        use_etag true
        use_last_modified true
        declared_type :remote_file
        cookbook_name "sensu"
        mode 420
             end
```

tyler-ball:
@tacchino Will you open a new issue for that?

PR #816, opened 2015-08-12
fixed SuSe OS busser install

Peuserik:
In SuSe-based OS the rubygem binaries are installed with their version number in their name. This behavior can be disabled with the option --no-format-executable. Without this little change the busser installation will fail on SuSe-based OS like OpenSUSE or SLES

testet with 
OS: openSUSE 13.2, SLES 12 Ubuntu 14.04
Provisioner: kitchen-ansible
Driver: kitchen-vagrant
            kitchen-docker

with verifier setting:
verifier:
  ruby_bindir: '/usr/bin'

### DIFF
--- a/lib/kitchen/verifier/busser.rb
+++ b/lib/kitchen/verifier/busser.rb
@@ -175,7 +175,7 @@ module Kitchen
 
         args = gem
         args += " --version #{version}" if version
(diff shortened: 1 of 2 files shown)


fnichol:
Looks good to me, thanks!

PR #835, opened 2015-09-22
WinRM connect (with retry) is failing on Windows

Stift:
I just found out that when test kitchen tries to connect to an ec2 instance via winRM it fails because the exception type is not filtered and therefore it fails directly. This only happened on windows machines on linux everything was fine.
The added exception type should now prevent the instant failing, when the ec2 instance is not ready yet.

A related issue is: test-kitchen/kitchen-ec2#175


smurawski:
:+1:

mwrock:
:+1:

adamleff:
:+1:

Stift:
very nice. now I can finally remove my custom version :grinning:

PR #837, opened 2015-09-23
include hidden files when copying over from sandbox to target

hslee16:
Currently, the upload from sandbox_dir doesn't include hidden files (.e.g [.filename])

This PR aims to include hidden files.

### DIFF
--- a/lib/kitchen/provisioner/base.rb
+++ b/lib/kitchen/provisioner/base.rb
@@ -59,7 +59,7 @@ module Kitchen
       # @raise [ActionFailed] if the action could not be completed
       def call(state)
         create_sandbox
-        sandbox_dirs = Dir.glob(File.join(sandbox_path, "*"))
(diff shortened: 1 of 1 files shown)


cheeseplus:
My initial feeling is that this behaviour should be something optional and not the default.

coderanger:
:-1: I don't think this is a good idea overall, and this patch doesn't implement the feature correctly anyway.

coderanger:
(stuff like `.~whatever` wouldn't be transferred, should use `Dir.entries` or similar to actually get all files if that's what we want to do)

cheeseplus:
For the reasons above as well as not wanting to make this the default behaviour I'm going to close the PR. That said we have some other PRs and Issues related to issues around including/excluding files that will likely address this concern.

Thanks again for the PR!

hslee16:
@cheeseplus @coderanger thanks for the reply!

PR #838, opened 2015-09-24
Config class should be initialized with env variables if they are set

nimeshsubramanian:
Preferring an environment variable if set over the default value when initialize an instance of config class
Also removed a redundant line from rake_tasks. Config class creates a logger instance.


nimeshsubramanian:
Closing this in favor of https://github.com/test-kitchen/test-kitchen/pull/839

PR #839, opened 2015-09-24
Config class should be initialized with env variables if they are set

nimeshsubramanian:
Preferring environment variable if set over the default values when an instance of config class is initialized.
Also removed a redundant line from rake_tasks. Config class creates a logger instance.


cheeseplus:
+1

cheeseplus:
Closing out due to age, need for rebased, and lack of additional interest.

PR #841, opened 2015-09-29
option to remove chef repo from server after chef run

neillturner:
It useful to delete the chef repo and the all the /tmp/kitchen files after a chef run so the server is clean and can be use for other purposes after the chef run. This way the server will not contain the chef repo and maybe configuration info on the server which you might want to keep secret.


cheeseplus:
Would be ok with a generic cleanup command or a chef specific one but this introduces puppet into core as well. We're already planning to remove chef from core so closing this one on the same principle.

Issue #657, opened 2015-03-26
Transfer files more efficiently.


danieljimenez:
When I use -l debug, it seems like test kitchen is transferring every file one at a time, very slowly, rather than taring them before transfer. I'm not sure if it uses something under the hood, but this step takes over 2 minutes on some of my cookbooks.

jessecrossen:
I'm seeing this as well. It seems like an even better solution would be to use rsync or similar.

danieljimenez:
Yeah, +1 for rsync.

jessecrossen:
@danieljimenez Looking at the source code it seems like in principle it would be relatively easy to use something like rsync, however the current code seems to use https://github.com/net-ssh/net-scp, and I'm not aware of any similarly abstracted (i.e. pure Ruby) equivalent for rsync. If anyone is able to find such a module, I'd be willing to take a stab at forking and writing in this feature if I can find the time.

Of course, yet another alternative would be to use modification times to skip files that are unmodified since the last run.

pelletiermaxime:
https://github.com/coderanger/kitchen-sync

jessecrossen:
@pelletiermaxime Thanks!

jessecrossen:
But it does beg the question: why is this functionality in a separate repo that monkey-patches its way into test-kitchen? Would anyone like to see this implemented as a pull request to this repo?

danieljimenez:
I couldn't get it to work, just... nothing happened.

jessecrossen:
@danieljimenez I couldn't get it to work either. I think it might be because the monkey patch was made for an earlier version and no longer maps onto anything useful, but I don't know for sure. That's why I'm suggesting making that functionality a part of test kitchen itself. It seems strange to have a modular transport system with no built-in alternatives.

pelletiermaxime:
Ah, you're right, kitchen-sync seems broken since test-kitchen 1.4. But in my tests, the new version is way faster, probably thanks to the added compression in the ssh transport.

danieljimenez:
I've been using 1.4 for a bit now, and still see very slow transfers when using `kitchen-ec2`.

iiro:
I see very slow transfers with 1.4.0 too with kitchen-ec2 - making Test-Kitchen almost useless.

Not sure where the true problem lies:

https://github.com/test-kitchen/kitchen-ec2/issues/93

https://github.com/coderanger/kitchen-sync/issues/8

iiro:
[...]

PR #845, opened 2015-10-02
Add ssh_tgz transport mechanism that compresses files (with tar-gzip)

northvankiwiguy:
This drastically improves performance of uploading cookbooks, when working over high-latency network links. This fixes Issue #657. I understand that there's another PR open for this same problem, but it seems to have stalled.

To make my solution work, you need to add:

```
transport:
  name: ssh_tgz
```

I added the tar-gzip feature to a separate transport, so that it's optional, rather than being integrated into the `ssh` transport. It also falls-back to using regular ssh (file-by-file upload) if anything goes wrong.

### DIFF
--- a/lib/kitchen/tgz.rb
+++ b/lib/kitchen/tgz.rb
@@ -1,0 +1,176 @@
+# -*- encoding: utf-8 -*-
+#
+# Author:: Peter Smith (<peter@petersmith.net>)
+#
+# Copyright (C) 2015, Peter Smith
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require "tempfile"
+require "zlib"
+require "rubygems/package"
+
+module Kitchen
+  #
+  # Error thrown if the Tgz has an invalid format.
+  #
+  class GzipFormatError < StandardError; end
+
+  #
+  # Represents a Tar-Gzip file, allowing multiple files to be combined
+  # into a single file. This is useful for transmitting a large number
+  # of files across high-latency networks.
+  #
+  class Tgz
+    #
+    # @return [String] the file system path of the generated tar-gzip file.
+    #
+    attr_reader :path
+
+    #
+    # Create a new Tgz object, in preparation for adding files to the
+    # tar-gzip archive.
+    #
+    # @param tgz_file_name [String, nil] The output file name, in
+    #   tar-gzipped format. If not specified, a temporary file name will be generated.
+    #
+    def initialize(tgz_file_name = nil)
+      if tgz_file_name
+        # user-provided output file name
+        @tgz_file = File.open(tgz_file_name, "wb+")
+        @path = tgz_file_name
+      else
+        # auto-generated output file name
+        @tgz_file = Tempfile.new("tgz")
+        @tgz_file.binmode
+        @path = @tgz_file.path
+      end
+
+      #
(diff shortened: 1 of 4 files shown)


iiro:
:+1:

SIGUSR2:
:+1:

joshzitting:
This is a great feature! makes kitchen runs SOOO much faster with bigger cookbooks etc!

hulu1522:
:+1:

fractil:
Huge performance boost :+1:

coderanger:
There is no need to put this in Kitchen. Make it a gem, and if it gets adopted we can look at adding it to core. Also obligatory plug for kitchen-sync which sets compression modes more aggressively and should probably be faster than this anyway.

northvankiwiguy:
Sure, it could be made a gem, but I'd prefer not to worry about whether the gem is getting out of sync with the main test-kitchen code. This is what often happens when a plug-in is created, and the API is not 100% stable.

coderanger:
Okay, have you benchmarked this against kitchen-sync?

northvankiwiguy:
no I haven't, but my 8 minute upload now takes 8 seconds, so that's fast enough for me. Optimizing it even further didn't seem worth the effort, given that other parts of the test-kitchen run are now much slower.

coderanger:
Okay, well if you aren't interested in actually checking if this is a good idea or not I'll repeat that you should make a gem and we can see if the community uses it or not :)

joshzitting:
@coderanger he put in an optional transport method. It isn't like he is forcing ppl to use this? Why go thru the hassle of changing this and making it a gem thus making it a bigger pain to maintain and use?

coderanger:
Because adding stuff to core (especially hacky stuff like this) creates a maintenance burden and we should only take that on in core once there is a clear signal from the community that they think it is worth the downsides (ex. this requires tar on the remote side so can fail with some minimal container images). This is also why kitchen-sync is still a gem.

northvankiwiguy:
To be technically correct, this hacky solution will actually fall back to using the regular upload mechanism if the tar command fails. And... as @joshzitting pointed out, it's purely optional to use this feature.

I think I'll let the maintainers of test-kitchen decide on whether they want to include this. If they don't, then I'll create a gem. There already seems to be 5 members of the community giving their :+1: on this change.

iiro:
The slowness of the vanilla Test-Kitchen is an issue which has been around quite a long time. What would be the non-hacky solution to get rid of slow syncs - or can there be one? I understand you don't want kludges around TK core - but is this one really?

Considering how quickly this one got :+1: 's, there might be quite many devs silently suffering this problem also. They haven't just found these issue tickets / PRs...

PS. Unibet has something released as well: https://github.com/unibet/kitchen-transport-sshtar and https://github.com/unibet/kitchen-transport-rsync...

neillturner:
I implement ssh_gzip in my kitchen-ssh driver to solve this problem as well.

xaniasd:
+1

EdHurtig:
Would love to see this in TK!

cheeseplus:
So speaking generally the plugin approach that @coderanger mentions is the way we want to do things. This abstraction works well so we can have lots of drivers/transports without trying to do everything in core.

That said we are aware that folks want TK transport to be faster but right now the extensibility of external gems works and we're trying to batten down core before adding anything.

northvankiwiguy:
OK, if that's the official decision, then I'll live with it. However, I'm not in a position to support this as an external plugin, which means it'll just end up becoming unstable and incompatible with the core over time. Unfortunately, that's why I wrote this code in the first place, since none of the existing plugins seemed to be compatible with the test-kitchen core, indicating to me that their maintainers also didn't have time to keep them up to date :-(

cheeseplus:
@northvankiwiguy well "official" may be a bit much but we've got a new group of maintainers in so we're trying to balance new features vs debt. We've got some good plugins now for this that the community are using and will definitely look at adding core.

As to maintainers not having time, that was totally a thing hence the new resurgence of activity with new maintainers. What we want to do is finish out lots of the planned work in abstracting TK cleanly at every interface level so it's easily extensible - don't fear, we'll make it clear how to use the ecoystem plugins to highlight and if things keep momentum we may even adopt some plugins into core or our maintainership.

PR #848, opened 2015-10-08
Pin mixlib-install more strictly

sersut:
We are doing some work on mixlib-install to extend its functionality and we will be making some compat breaking changes. E.g. https://github.com/chef/mixlib-install/pull/11

We would like to pin it more strictly here until we get the mixlib-install to 1.0.0 stage.


schisamo:
Do we think the new mixlib-install version released will be 1.0.0? If so won't the existing pinning protect things?

sersut:
I think eventually yes. For now @thommay want to use 0.8.0 as the version (see the conversation in the linked PR) so we want to make this a bit more strict.

cheeseplus:
Is this still needed?

mwrock:
closing this since we are now consuming 0.7.1 and dont want 0.7.0. We can open a new pr if we feel we need to hard pin to that.

PR #859, opened 2015-10-20
Rsync transport for rsync+ssh file transfers

mattlqx:
This is a hack to override the upload method on the Ssh transport to use an external rsync binary to transfer needed files between local and remote. It's not glamorous and I have no idea how to actually write a good spec file for it, but here it is in case it helps anybody.

To use it, in your kitchen.yml specify:
```
transport:
  name: rsync
```

And if you want to preserve files between converges (which will speed up the rsync):
```
provisioner:
  clean_files_first: false
```

### DIFF
--- a/lib/kitchen/provisioner/chef_base.rb
+++ b/lib/kitchen/provisioner/chef_base.rb
@@ -42,6 +42,7 @@ module Kitchen
       default_config :run_list, []
       default_config :attributes, {}
       default_config :log_file, nil
+      default_config :clean_files_first, true
       default_config :cookbook_files_glob, %w[
         README.* metadata.{json,rb}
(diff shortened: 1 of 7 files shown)


mattlqx:
Derp, I see https://github.com/unibet/kitchen-transport-rsync already exists, which is mostly the same thing. And it's made as a gem, which is a better call, but I'm not sure it would be possible to integrate the "preserve files between runs" functionality through a gem.

coderanger:
@mattlqx See the kitchen-sync gem for an external implementation.

cheeseplus:
Definitely would use the kitchen-sync implementation (or similar) internally but at this time we'll close this given the dependency on a local binary and existing alternatives that better fit the implementation.

PR #860, opened 2015-10-21
clean command for remote sandbox

neillturner:
Add clean command for remote sandbox on an instance. This is useful to run before creating an image of a  server instance after it has been provisioned as we don't want to leave the chef repo on the server. The provisioner plugins can override the base clean method to add additional cleanup as necessary.


cheeseplus:
This can be handled with the lifecycle hooks: https://github.com/test-kitchen/test-kitchen/pull/1428

PR #861, opened 2015-10-21
package driver command

neillturner:
Add a new command package that runs a package driver command on the instance. 
The vision is that test-kitchen will allow packaging of the provisioner server. 
kitchen-ec2 driver can implement code to create an image on AWS.
kitchen-docker could write code to implement creating a docker image in a docker registry. 
This works in conjunction with the clean command which deletes the remote sandbox containing the chef repo so when a package command is run then the images does not contain the chef repo.

### DIFF
--- a/lib/kitchen/cli.rb
+++ b/lib/kitchen/cli.rb
@@ -224,6 +224,13 @@ module Kitchen
       perform("login", "login", args)
     end
 
+    desc "package INSTANCE|REGEXP", "package an instance"
+    log_options
+    def package(*args)
+      update_config!
+      perform("package", "package", args)
+    end
+
     desc "exec INSTANCE|REGEXP -c REMOTE_COMMAND",
       "Execute command on one or more instance"
     method_option :command,
--- a/lib/kitchen/command/package.rb
+++ b/lib/kitchen/command/package.rb
@@ -1,0 +1,36 @@
+# -*- encoding: utf-8 -*-
+#
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require "kitchen/command"
+
+module Kitchen
+
+  module Command
+
+    # Execute command on remote instance.
+    #
+    class Package < Kitchen::Command::Base
+
+      # Invoke the command.
+      def call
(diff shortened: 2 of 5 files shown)


mwrock:
Interesting. It seems by this PR and #860 you are treating test-kitchen similar to how one uses packer - a means of creating base images to be used for other VMs/Containers.

On the one hand I'd be cautious to have test-kitchen fill that role because that's not the set of concerns I think it is aiming to address. On the other hand, as a cookbook author, test-kitchen just makes this stuff so easy and in the end, you are SOOO close to what you get with packer without changing workflow. So I definitely see the motivation.

neillturner:
With the rise of docker i expect we will move away from the current approach of: 
* test-kitchen to development and test cookbook(s)
* checkin to git
* some sort of pipeline to production with end result deployed to chef server.
to 
* test-kitchen to development and test cookbook(s)
* checkin to git
* package in docker
* some sort of pipeline to production via docker containers.
ofcouse package to docker doesn't have to go through test-kitchen. 
it could just be a docker command run in the ci steps. 
(i expect that is how to is going to work)

neillturner:
test-kitchen in the future will need to compete with packer i believe.

 packer and test-kitchen are quite similar really. 
 if packer added the ability to verify servers with tests
and 
 test-kitchen added the ability to create packages 
they would be very similar.

cheeseplus:
Where is the package command itself or is that something you have to provide? I feel like this may better belong as part of each of the drivers (vagrant, aws, etc)

Not opposed to allowing it but want to make sure it's put in the right place.

ccope:
Looks to me like this just stubs out the base class, and the drivers are supposed to override the base package method with a real implementation. @neillturner do you have a driver with an example implementation?

neillturner:
No i don't as it just allows the possibility of "real implementations" particularly for docker. That's how you encourage innovation. One the best decision fletcher did was designing a plugin architecture with separate gems. Hashicorp didn't do this so much wit vagrant which has reduced their speed of  innovation -). 

I guess I have been much more aggressive adding functionality from pull requests in my kitchen-puppet and kitchen-ansible plugins which would not have been possible if they were not separate gems. test-kitchen has been slow to process all these pulls and issues. Generally i think that the core of test-kitchen should be lots of these options which allow people to write code that is external. 

 Also i see we are moving to more of a build container approach and test-kitchen getting more "packer" type functionality. I think that is inevitable but i guess it just depends if you guys want test-kitchen to be part of that. -)

ccope:
I'd definitely be interested in functionality like this. I currently have vagrant and testkitchen setups using the vagrant-lxc plugin, but to package snapshots of containers my team went with vagrant provisioning plus executing some commands over vagrant ssh to clean up the image and do smoketests. I'd rather do it with testkitchen though, since the testing could be much more thorough.

cheeseplus:
@neillturner any chance we could get a rebase?

neillturner:
I have rebased and pushed this as 
package driver command #1074
any chance of merging this?

neillturner:
closing this as Add package driver command #1074 is merged.

Issue #811, opened 2015-08-06
WinRM commandline limits can break Windows bootstrapping


jf647:
On TK 1.4.1:

```
-----> Starting Kitchen (v1.4.1)
-----> Creating <default-win2012r2>...
       Bringing machine 'default' up with 'virtualbox' provider...
       ==> default: Importing base box 'win2012r2'...
==> default: Matching MAC address for NAT networking...
       ==> default: Setting the name of the VM: kitchen-example-default-win2012r2_default_1438881656026_31809
       ==> default: Clearing any previously set network interfaces...
       ==> default: Preparing network interfaces based on configuration...
           default: Adapter 1: nat
       ==> default: Forwarding ports...
           default: 3389 => 3389 (adapter 1)
           default: 5985 => 5985 (adapter 1)
           default: 22 => 2222 (adapter 1)
       ==> default: Running 'pre-boot' VM customizations...
       ==> default: Booting VM...
       ==> default: Waiting for machine to boot. This may take a few minutes...
       ==> default: Machine booted and ready!
       ==> default: Checking for guest additions in VM...
       ==> default: Machine not provisioning because `--no-provision` is specified.
       [WinRM] Established
       Vagrant instance <default-win2012r2> created.
       Finished creating <default-win2012r2> (1m18.77s).
-----> Converging <default-win2012r2>...
       Preparing files for transfer
       Preparing dna.json
       Resolving cookbook dependencies with Berkshelf 3.3.0...
       Removing non-cookbook files before transfer
       Preparing validation.pem
       Preparing client.rb
-----> Installing Chef Omnibus (install only if missing)
       Downloading package from https://opscode-omnibus-packages.s3.amazonaws.com/windows/2008r2/x86_64/chef-client-12.4.1-1.msi
       Download complete.
```

On TK 1.4.2:

```
-----> Starting Kitchen (v1.4.2)
-----> Creating <default-win2012r2>...
       Bringing machine 'default' up with 'virtualbox' provider...
       ==> default: Importing base box 'win2012r2'...
==> default: Matching MAC address for NAT networking...
       ==> default: Setting the name of the VM: kitchen-example-default-win2012r2_default_1438881568509_73724
       ==> default: Fixed port collision for 3389 => 3389. Now on port 2200.
       ==> default: Fixed port collision for 5985 => 5985. Now on port 2201.
       ==> default: Fixed port collision for 22 => 2222. Now on port 2202.
[...]

PR #864, opened 2015-10-26
Fix Windows Omnibus Install #811

dissonanz:
Updated `select-object` syntax and added a split so that `$chef_version` returns the chef version from version-manifest.txt


jaym:
:+1: thought it looks like this file is in mixlib-install now. cc @thommay

dissonanz:
Looks like it's the same. Submitted PR to chef/mixlib-install.

PR #802, opened 2015-08-03
Rescue Errno::ETIMEDOUT instead of Timeout::Error on Establish

Annih:
Fix #800 
net-ssh 2.10 does not use 'timeout' anymore, it use the Socket `connect_timeout` instead.
See: https://github.com/net-ssh/net-ssh/commit/58b379787b19b47e3b17f73ac05bda208772bb3c and https://github.com/net-ssh/net-ssh/commit/9bfbdc98e8d36e8f62792a6ecdd258bc232b9d9a

PR #866, opened 2015-10-27
Fixes constant error when using recent net-ssh

komuta:
Following the changes in #802, depending on the ruby / gems environnement the ssh transport doesn't work anymore. Here is the exact error, as seen in a Ruby 2.1.2 environment with Kitchen 1.4.2 and net-ssh 2.10.0beta2:

```
-----> Starting Kitchen (v1.4.2)
------Exception-------
Class: Kitchen::ClientError
Message: Could not load the 'ssh' transport from the load path. Please ensure that your transport is installed as a gem or included in your Gemfile if using Bundler.
---Nested Exception---
Class: NameError
Message: uninitialized constant Kitchen::Transport::Ssh::Connection::Timeout
```

This quick fix ensures the *Timeout* constant is known by the system.

### DIFF
--- a/lib/kitchen/transport/ssh.rb
+++ b/lib/kitchen/transport/ssh.rb
@@ -18,6 +18,7 @@
 
 require "kitchen"
 
+require "timeout"
 require "net/ssh"
 require "net/scp"
 


cheeseplus:
Is this still necessary with the latest net-ssh updates?

cheeseplus:
Closing as this should have been fixed by other net-ssh updates and fixes.

PR #872, opened 2015-11-03
Added try/catch around main and set error action to stop

mcallb:
Added a try/catch around main and set the ErrorActionPreference to stop from the default value continue. It's confusing to troubleshoot an exception when a script continues to run through non terminating errors. In my case I had a malformed chef_metadata_url text file but the script continued to run until it tried to install an MSI that didn’t exist. See following gist for an example of the difference in error output https://gist.github.com/mcallb/6aa219b8ebcbb536edc0.


mwrock:
:+1: this seems like a good move.

cheeseplus:
:+1:

PR #875, opened 2015-11-05
Run chef-client twice in chef-zero provisioner

### LINKED ISSUE #874: Enforce suite idempotency
Discussions in chef summit in London lead us to the need to automatically test for idempotency and convergence in one chef-client run.
This is usually checked by hand, running a second kitchen converge and ensuring no resource is updated.

I'd like to allow suites to make this check automatically.

This relates to discussions in #289.

kamaradclimber:
chef-zero provisioner runs chef-client twice. Second time expect no
resource to be updated. This is a way to enforce idempotency.

This is a proof of concept to fix #874 but it lacks:
- testing
- windows support
- documentation

Could you tell me what you think of the need, this proof of concept ? And also guide me to write relevant testing.

### DIFF
--- a/lib/kitchen/provisioner/chef_base.rb
+++ b/lib/kitchen/provisioner/chef_base.rb
@@ -151,6 +151,10 @@ module Kitchen
 
       private
 
+      def last_exit_code
+        "; exit $LastExitCode" if powershell_shell?
+      end
+
       # @return [Hash] an option hash for the install commands
       # @api private
       def install_options
--- a/lib/kitchen/provisioner/chef_base.rb
+++ b/lib/kitchen/provisioner/chef_base.rb
@@ -393,7 +393,6 @@ module Kitchen
         end
 
         prepare_config_idempotency_check if config[:enforce_idempotency]
-
       end
 
       # Writes a configuration file to the sandbox directory
--- a/lib/kitchen/provisioner/chef_base.rb
+++ b/lib/kitchen/provisioner/chef_base.rb
@@ -68,6 +68,8 @@ module Kitchen
         providers/**/* recipes/**/* resources/**/* templates/**/*
       ].join(",")
 
+      default_config :multiple_converge, 1
+
       default_config :enforce_idempotency, false
 
       default_config :data_path do |provisioner|
@@ -432,20 +434,35 @@ module Kitchen
       # Gives the command used to run chef
       # @api private
       def chef_cmd(base_cmd)
+        if windows_os?
+          separator = [
+            "; if ($LastExitCode -ne 0) { ",
+            "throw \"Command failed with exit code $LastExitCode.\" } ;"
+          ].join
+        else
+          separator = " && "
+        end
+        chef_cmds(base_cmd).join(separator)
+      end
+
+      # Gives an array of command
+      # @api private
+      def chef_cmds(base_cmd)
         cmd = prefix_command(wrap_shell_code(
           [base_cmd, *chef_args(config_filename), last_exit_code].join(" ").
           tap { |str| str.insert(0, reload_ps1_path) if windows_os? }
         ))
+
+        cmds = [cmd].cycle(config[:multiple_converge].to_i).to_a
 
         if config[:enforce_idempotency]
           idempotent_cmd = prefix_command(wrap_shell_code(
             [base_cmd, *chef_args("client_no_updated_resources.rb"), last_exit_code].join(" ").
             tap { |str| str.insert(0, reload_ps1_path) if windows_os? }
           ))
-          separator = windows_os? ? " ; " : " && "
-          cmd = [cmd, idempotent_cmd].join(separator)
+          cmds[-1] = idempotent_cmd
         end
-        cmd
+        cmds
       end
     end
   end
--- a/lib/kitchen/provisioner/chef_base.rb
+++ b/lib/kitchen/provisioner/chef_base.rb
@@ -68,6 +68,8 @@ module Kitchen
         providers/**/* recipes/**/* resources/**/* templates/**/*
       ].join(",")
 
+      default_config :enforce_idempotency, false
+
       default_config :data_path do |provisioner|
         provisioner.calculate_path("data")
       end
@@ -375,6 +377,76 @@ module Kitchen
             "Either use a different provisioner, or delete/rename " \
             "#{policyfile}"
         end
+      end
+
+      # Writes a configuration file to the sandbox directory.
+      # @api private
+      def prepare_config_rb
+        data = default_config_rb.merge(config[config_filename.gsub(".", "_").to_sym])
+        data = data.merge(:named_run_list => config[:named_run_list]) if config[:named_run_list]
+
+        info("Preparing #{config_filename}")
+        debug("Creating #{config_filename} from #{data.inspect}")
+
+        File.open(File.join(sandbox_path, config_filename), "wb") do |file|
+          file.write(format_config_file(data))
+        end
+
+        prepare_config_idempotency_check if config[:enforce_idempotency]
+
+      end
+
+      # Writes a configuration file to the sandbox directory
+      # to check for idempotency of the run.
+      # @api private
+      def prepare_config_idempotency_check
+        handler_filename = "chef-client-fail-if-update-handler.rb"
+        source = File.join(
+          File.dirname(__FILE__), %w[.. .. .. support ], handler_filename
+        )
+        FileUtils.cp(source, File.join(sandbox_path, handler_filename))
+        File.open(File.join(sandbox_path, "client_no_updated_resources.rb"), "wb") do |file|
+          file.write(format_config_file(data))
+          file.write("\n\n")
+          file.write("handler_file = File.join(File.dirname(__FILE__), '#{handler_filename}')\n")
+          file.write "Chef::Config.from_file(handler_file)\n"
+        end
+      end
+
+      # Returns an Array of command line arguments for the chef client.
+      #
+      # @return [Array<String>] an array of command line arguments
+      # @api private
+      def chef_args(_config_filename)
+        fail "You must override in sub classes!"
+      end
+
+      # Returns a filename for the configuration file
+      # defaults to client.rb
+      #
+      # @return [String] a filename
+      # @api private
+      def config_filename
+        "client.rb"
+      end
+
+      # Gives the command used to run chef
+      # @api private
+      def chef_cmd(base_cmd)
+        cmd = prefix_command(wrap_shell_code(
+          [base_cmd, *chef_args(config_filename), last_exit_code].join(" ").
+          tap { |str| str.insert(0, reload_ps1_path) if windows_os? }
+        ))
+
+        if config[:enforce_idempotency]
+          idempotent_cmd = prefix_command(wrap_shell_code(
+            [base_cmd, *chef_args("client_no_updated_resources.rb"), last_exit_code].join(" ").
+            tap { |str| str.insert(0, reload_ps1_path) if windows_os? }
+          ))
+          separator = windows_os? ? " ; " : " && "
+          cmd = [cmd, idempotent_cmd].join(separator)
+        end
+        cmd
       end
     end
   end
--- a/lib/kitchen/provisioner/chef_zero.rb
+++ b/lib/kitchen/provisioner/chef_zero.rb
@@ -94,10 +94,6 @@ module Kitchen
 
       private
 
-      def last_exit_code
-        "; exit $LastExitCode" if powershell_shell?
-      end
-
       # Adds optional flags to a chef-client command, depending on
       # configuration data. Note that this method mutates the incoming Array.
       #
--- a/lib/kitchen/provisioner/chef_zero.rb
+++ b/lib/kitchen/provisioner/chef_zero.rb
@@ -43,6 +43,8 @@ module Kitchen
           tap { |path| path.concat(".bat") if provisioner.windows_os? }
       end
 
+      default_config :enforce_idempotency, false
+
       default_config :ruby_bindir do |provisioner|
         provisioner.
           remote_path_join(%W[#{provisioner[:chef_omnibus_root]} embedded bin])
@@ -74,12 +76,20 @@ module Kitchen
       def run_command
         cmd = modern? ? local_mode_command : shim_command
 
-        prefix_command(
-          wrap_shell_code(
-            [cmd, *chef_client_args, last_exit_code].join(" ").
+        chef_client_cmd = prefix_command(wrap_shell_code(
+          [cmd, *chef_client_args("client.rb"), last_exit_code].join(" ").
+          tap { |str| str.insert(0, reload_ps1_path) if windows_os? }
+        ))
+
+        if config[:enforce_idempotency]
+          idempotent_check_client_cmd = wrap_shell_code(
+            [cmd, *chef_client_args("client_no_updated_resources.rb"), last_exit_code].join(" ").
             tap { |str| str.insert(0, reload_ps1_path) if windows_os? }
           )
-        )
+          separator = windows_os? ? " ; " : " && "
+          chef_client_cmd = [chef_client_cmd, idempotent_check_client_cmd].join(separator)
+        end
+        chef_client_cmd
       end
 
       private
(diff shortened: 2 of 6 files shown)


kamaradclimber:
@tyler-ball would you have feedback on the change ?

tyler-ball:
Hey @kamaradclimber - sorry for the delay.  I'm working on non-TK projects recently, so I'll ping around and see if anyone has time to review this.  Thanks for the contribution!

kamaradclimber:
ok thanks for the heads up

mwrock:
In addition to my comments above, it would be cool to work this into chef_base so that the solo provisioner can leverage it as well. 

This will also need tests as well.

Overall I think this is a great idea!

lib/kitchen/provisioner/chef_zero.rb
```diff
@@ -195,6 +207,23 @@ def prepare_chef_client_zero_rb
...
+# Handler to kill the run if any resource is updated
+class UpdatedResources < ::Chef::Handler
+  def report
+    if updated_resources.size > 0
+      #exit 203 # if chef handler code was not catching Exception
+      puts "Chef run should reach a converged state. Resources updated in a second chef-client run:"
+      updated_resources.each do |r|
+        puts r
+      end
+      Process.kill('KILL', Process.pid)
```
kamaradclimber: handler code catches all exceptions (https://github.com/chef/chef/blob/master/lib/chef/handler.rb#L215) which prevent to use a classic exception raise to end the run.

I agree this is a hack, I'd be happy to find another solution.

lib/kitchen/provisioner/chef_zero.rb
```diff
@@ -195,6 +207,23 @@ def prepare_chef_client_zero_rb
         FileUtils.cp(source, File.join(sandbox_path, "chef-client-zero.rb"))
       end
 
+      HANDLER_CODE = <<-EOH
+# Handler to kill the run if any resource is updated
+class UpdatedResources < ::Chef::Handler
```
kamaradclimber: I am not sure to understand what you mean, the code for this handler is already written in the client_no_updated_resources.rb file which is copied from the sandbox.

Do you mean that you want the handler code to be written in a specific file and included in client.rb conf using an include such as:
```ruby
Dir.glob(File.join("/etc/chef", "client.d", "*.rb")).each do |conf|
  Chef::Config.from_file(conf)
end
```

kamaradclimber:
@mwrock thanks a lot for your feedback.

I have extracted the necessary code to make it work with chef-solo as well. Most of the code is now in chef_base class.

Could you give me some pointers to write proper tests for this feature?

kamaradclimber:
I have followed your suggestion to extract the handler code to support directory.

Specs are also there.

kamaradclimber:
Hi @mwrock,

I think we are now closer, would you have other comments to help this patch getting better?

mwrock:
I think if you added a `bin_file_name` method to the base class that the 2 provisioners could implement and fill in with `solo.rb` and `client.rb` and then not have to use those strings elsewhere, that could clean things up a bit. There may be a better name than `bin_file_name` though.

Also, should you add specs to `chef_solo_spec`?

kamaradclimber:
thanks @mwrock for your comments, I have implemented the required changes.

I have named the variable `config_filename` and make it default to client.rb

lib/kitchen/provisioner/chef_base.rb
```diff
@@ -265,6 +267,74 @@ def reload_ps1_path
...
+          file.write("handler_file = File.join(File.dirname(__FILE__), '#{handler_filename}')\n")
+          file.write "Chef::Config.from_file(handler_file)\n"
+        end
+      end
+
+      # Returns an Array of command line arguments for the chef client.
+      #
+      # @return [Array<String>] an array of command line arguments
+      # @api private
+      def chef_args(_config_filename)
```
mwrock: dont need the `_config_filename` param here anymore since the method can just use the new `config_filename` method.

kamaradclimber: actually I need it to access chef_args with the handler (see L330)

mwrock: ah right.

mwrock:
:+1: LGTM

kamaradclimber:
@mwrock  what should I do now to get this merged ?

mwrock:
First thing I'd do is address the rubocop violations and then maybe ping @tyler-ball

kamaradclimber:
@tyler-ball would you have time to review this PR ?

tyler-ball:
@kamaradclimber Unfortunately I do not - we're working internally on a plan for TK maintenance so hopefully we will have some news about that early next year

cheeseplus:
Can we get this rebased? We're happy to give it another look.

kamaradclimber:
@cheeseplus : just rebased.

However I don't understand why the test fail: https://travis-ci.org/test-kitchen/test-kitchen/jobs/102287127. It seems that some crendentials are missing.
Could you point me into the right direction?

kamaradclimber:
@cheeseplus would you give me a hint to solve the failing tests?

mwrock:
That test failure is unrelated to this PR so I think this is effectively passing.

cheeseplus:
The tests are likely fine but upon reviewing the code I feel like this option may not have the desired effect. For example: a chef run can still be idempotent and still update resources. This is not a mutually exclusive condition. I'm still thinking through the implications of this but right now I think this gives a false sense of security or at worst will misreport idempotency.

kamaradclimber:
@cheeseplus thanks for your feedback.
Yes resource can be "idempotent" and report as updated. But this is usually a bug than an intent from the resource maintainer.
The dual is also true: a resource can modify the system and not report to be updated.

However properly written resources do not have that kind of behavior.

The wish to have chef converging in one state is quite common (you can see it in foodcritic asking user to write gards on execute resources).

I would be glad to discuss the need for this feature, do  you want me to write a rationale for this feature?

cheeseplus:
So there are many cases that I have seen in the field where something like an `execute` or `ruby_block` have to run/update every run even if no action is taken - I agree that these could be better written as full on resources and certainly encourage that approach. I can't in good conscience call it a bug but it's certainly something I would want to improve. That said, the vast majority of Chef users are not that sophisticated and so I'm mostly concerned about their expectations of this feature.

I totally understand the rationale and I think we still may merge this since it's optional but it is something I want to give some more thought and discuss with others today.

thommay:
I think we should be thinking about this as a higher level option, not a chef specific one. In that case, I think it makes more sense to talk about running converge more than once, and if you want to be able to supply specific options to the second converge (such as your handler that enforces that there are no updated resources), then that is the choice of the user.
But that's not to say we shouldn't merge this in the short term.

cheeseplus:
I think @thommay is definitely on point with abstracting this out a bit more to make this generally useful. We can mark that as a future feature or improvement but I do want to ensure that we address the language issue and default case so that expectations are calibrated correctly.

kamaradclimber:
thanks for your comments @thommay and @cheeseplus.

I agree about the need to have a better name.
Also ok about abstracting but I think this should be done indeed in another patchset since I really don't know the other language test-kitchen supports.

In my opinion, `log` is exactly the same as `execute` you should have a guard to converge it only if makes sense. The common use case, for me, is to log if notified by another resource convergence. Otherwise, I usually use the `Chef::Log.info` method to add classical logging.

About the general need for the feature, I agree that most use case (users?) do not really care about the one-run-convergence. Even in a disciplined team, this is hard to achieve if your code has dependencies.
However being able to rely on chef "updated resource" list is really useful to understand what happens in your environment.

cheeseplus:
I entirely agree about your second point but it's the implementation approach that is very different for most of what I see in the field. In a production environments people are getting this data via the reporting API and can create rules/alerts in Chef Analytics that can be complex but also have a fully featured DSL for these rules. 

Moreover, we also have audit-mode which runs after the converge phase which can be used for much of this validation as well as the testing frameworks like ServerSpec/InSpec. Granted this doesn't address a multi-converge scenario but from my perspective relying on the updated resources can be dubious in terms of usefulness. I'd much rather assert the state, with specific checks, to tell me if things worked. Part of the trouble in my mind (and generally what we tell clients/customers) is that overloading chef to tell you if chef is working is a slippery slope - there isn't a check and balance as there is with something external or separate asserting state. What I am working around is that I while the "updated resource" list can definitely be useful I believe that there are other approaches that offer more value for answering that specific kind of issue. 

In summation:
- let's change the name of the config flag
- let's make the enforce_idempotentcy an option for the above flag
  - that way we get multiple converge but it's not bound to the above
  - we could also make the # of converges configurable
- we can look at abstracting this from being chef-specific in a separate patch in the future (assuming we address the above)

Sound good @kamaradclimber?

kamaradclimber:
good idea

I suggest:
- `multiple_converge` option whose value is the number of runs
- `enforce_idempotency` flag which fails the last run if it updates anything.

Tell me if that suits you

cheeseplus:
That seems very reasonable to me

kamaradclimber:
I've added a commit to add the multiple converge.

Two questions:
- for the windows case, I've used ";" (assuming the command is launched through powershell) but it does not have the same effect as "&&" on linux. I need to search how I could achieve the same intent, maybe you have input on this.
- I'd like to perform validate on the config since using `multiple_converge:1` and `enforce_idempotency:true` is not a valid conf. Is there a way to perform this validation?

mwrock:
* What issues are you having with `;` on windows?
* I think you just want to raise an error if you find that both are set `multiple_converge` is set to one and `enforce_idempotency` is `true`

kamaradclimber:
@mwrock : on linux && garantee that the command will succeed iff all chef-clients have succeeded.

It seems that `;` simply run command consecutively ignoring the intermediate exit codes.

Where can raise an error ? It seems weird to raise it before converging chef. It looks like it should be done when config is validated.

mwrock:
ah right @kamaradclimber because powershell really isn't natively exit code driven, it will happily keep running unless a terminating error is thrown or you check the `$LASTEXITCODE` variable and halt if not 0.

kamaradclimber:
For consistency I think checking $LASTEXITCODE is necessary, do you have a preference?

kamaradclimber:
@mwrock  I've added the lastexitcode check

mwrock:
looks like you have some rubocop violations

kamaradclimber:
ok I've fixed the issues (we still have the failing tests though)

kamaradclimber:
@mwrock @cheeseplus I think we are getting closer. Would you have more feedback ?

kamaradclimber:
@mwrock @cheeseplus Do you want to me to improve the continue to improve the code ?

kamaradclimber:
@thommay Could you help me to get this merged? I'll be happy to refactor the code if necessary. Thanks

kamaradclimber:
Closing this PR as I don't get the traction I hoped

artem-sidorenko:
:-(

kamaradclimber:
@artem-sidorenko feel free to ask to a maintainer to reopen it.

cheeseplus:
Re-opening this one as we're definitely open to it but most of us were slammed leading into ChefConf.

kamaradclimber:
Just rebased. Test are not passing yet because of inexisting branch winrm-v2 for train. Reverting it locally makes all tests but 1 pass.

Waiting for a stable master branch to investigate.

cheeseplus:
@kamaradclimber master is fixed - upstream deps nuked some branches post merge.

kamaradclimber:
thanks @cheeseplus for the heads-up, the test are now passing

axos88:
any chance to get this merged?

kamaradclimber:
@axos88 I'll fix the conflicts once I get some feedback that the PR will seriously be considered (nearly 11 months since last feedback about the content)

thommay:
I think if you have passing tests (and it looks like the last time travis ran you did), and the conflicts get fixed, we'd probably merge it.

cheeseplus:
@kamaradclimber definitely still open to this, I was just on the road for most of the end of last year and am now on a new team where I have more time for kitchen related tasks

kamaradclimber:
@cheeseplus I have the feeling that this discussion is going nowhere since the PR has been opened for >1y, rebased countless times and feedback taken into account whenever available.

I'll close it to avoid cluttering the PR list even more.

If you are still interested I'll let anyone reopen and fix the conflicts.

thommay:
@kamaradclimber please just fix the conflicts and we'll merge it, like we said last month.

thommay:
(or tick the button that lets maintainers push to your branch, and it might well just get done)

coderanger:
This might be better handled via the supersuite system rather than something one-off for this provisioner.

thommay:
@kamaradclimber thanks for your patience on this, and yay for a great feature. 
@coderanger we can revisit this as and when the super suite stuff becomes a thing :)

kamaradclimber:
thx a lot @thommay!

axos88:
Now... how to use it? :)

kamaradclimber:
@axos88 two config options have been added:`multiple_converge` and `enforce_idempotency`.  If you set the former to `2` and the latter to `true`, you should be able to get it.

axos88:
Why are there two? What is the use case of multiple converge withot enforce idempotency?

kamaradclimber:
@axos88 you're right having two options might seems over complex. The discussion in the PR will retrace why we came up to have two options.

I'll add that you can check several scenarios with the different values of those options:

|   | multiple_converge: 1  (default)| multiple_converge > 1 |
|---|---|---|
| enforce_idempotency: false (default) | default behavior of test-kitchen. One chef-client run succeeds | Simply test that several chef-client run works. This can detect a class of issues where one convergence sets the system into a non usable state|
| enforce_idempotency: true | I'm not sure that this has a practical use. | Ensure that the system is fully converged after `multiple_converge` runs. This is the core of the feature|

axos88:
Ah. gotcha. thanks.

neillturner:
You must take the chef provisioner into a separate code base and a separate plugin if it takes 1 year and 3 months to get such a simple change into the product. I'm innovating much faster in my ansible and puppet plugins (although i admit my test coverage is inadequate -)) which is crazy when test-kitchen is a chef product.

cheeseplus:
@neillturner, to be very clear, test-kitchen has never been a Chef 'Product' and nobody at the company Chef (who is also my employer) is dedicated/paid to work on test-kitchen.

I agree that the chef code needs to be broken out but right now it's largely just myself covering the bulk of kitchen-* things.

ryangribble:
Im running test kitchen 1.16 (on OSX) but these new options aren't working for me.  The `multiple_converge` one seems to have no effect, and when I have `enforce_idempotency` in my kitchen.yml file, I actually get an error when kitchen tries to converge.  Any ideas?

```
Chef Development Kit Version: 1.3.43
chef-client version: 12.19.36
delivery version: master (dd319aa632c2f550c92a2172b9d1226478fea997)
berks version: 5.6.4
kitchen version: 1.16.0
```

```
-----> Converging <default-windows2012r2-windowsonly>...
       Preparing files for transfer
       Preparing dna.json
       Resolving cookbook dependencies with Berkshelf 5.6.4...
       Preparing data_bags
       Preparing validation.pem
       Preparing client.rb
>>>>>> ------Exception-------
>>>>>> Message: 1 actions failed.
>>>>>>     Failed to complete #converge action: [undefined local variable or method `data' for #<Kitchen::Provisioner::ChefZero:0x007fc1e3997c60>] on default-windows2012r2-windowsonly
>>>>>> ----------------------
```

this is the relevant parts of the kitchen.yml
```
driver:
  name: vagrant
  gui: true
  provider: virtualbox
  linked_clone: true

driver_config:
  guest: windows
  communicator: 'winrm'

provisioner:
  name: chef_zero
  enforce_idempotency: true
  multiple_converge: 2

platforms:
  - name: windows2012r2-windowsonly
```

MarkGibbons:
Hi Ryan,

The enforce_idempotency error is likely the same thing fixed in #1068  in method prepare_config_idempotency_check.

multiple_converge worked for me on a version of test kitchen between 1.15 and 1.16. I see two chef-client runs using your settings. I run a forked version of test kitchen though.

Mark

ryangribble:
Yep that fix on that other (unrelated) PR looks like the problem I've got 

I'd suggest splitting it out into a stand alone PR since its not really related to #1068 and that way this feature would be fixed for the next release regardless of whether the no proxy thing gets in or not

DPD-30:
Added pull request with Just the needed change to correct the problem see PR 1252 above.

ghostsquad:
why is this at the provisioner level instead of the suite level? Only certain suites I have are idempotent. :(

PR #878, opened 2015-11-11
write install_command to file and invoke on the instance to avoid command too long on windows

### LINKED ISSUE #854: test-kitchen winrm w/proxies "The command line is too long."
I'm having a problem using test-kitchen with a windows host thats behind a proxy with authentication.  Without the proxy I'm able to run a kitchen create, but my converge will fail (because of the proxy).  That's expected.  But when I add the proxy strings to the provider section of test-kitchen, everything fails with a command too long error.

```
$ kitchen converge Windows2
-----> Starting Kitchen (v1.4.2)
-----> Converging <default-Windows2>...
       Preparing files for transfer
       Preparing dna.json
       Resolving cookbook dependencies with Berkshelf 4.0.1...
       Preparing validation.pem
       Preparing client.rb
$$$$$$ The command line is too long.
>>>>>> Converge failed on instance <default-Windows2>.
```

I'm passing the proxies (both http and https) as 
```
provisioner:
  name: chef_zero
  http_proxy: http://uuuuuuuuuuu:pppppppp@myproxyserver
  https_proxy: https://uuuuuuuuuuu:pppppppp@myproxyserver
```
Those are the actual lengths of the strings being passed, just cleaned.  The same settings work on a linux host.  Both the https and http connections are being proxied from port 80, I left it out trying to shorten the command some (along with the rest of the domain) but no dice.

Here's the full error and the return code:
```
$ kitchen converge Windows2
-----> Starting Kitchen (v1.4.2)
-----> Converging <default-Windows2>...
       Preparing files for transfer
       Preparing dna.json
       Resolving cookbook dependencies with Berkshelf 4.0.1...
       Preparing validation.pem
       Preparing client.rb
$$$$$$ The command line is too long.
>>>>>> Converge failed on instance <default-Windows2>.
>>>>>> Please see .kitchen/logs/default-Windows2.log for more details
>>>>>> ------Exception-------
>>>>>> Message: WinRM exited (1) for command: [$env:http_proxy = "http://uuuuuuuuuuu:pppppppp@myproxyserver"
$env:HTTP_PROXY = "http://uuuuuuuuuuu:pppppppp@myproxyserver"
$env:https_proxy = "https://uuuuuuuuuuu:pppppppp@myproxyserver"
$env:HTTPS_PROXY = "https://uuuuuuuuuuu:pppppppp@myproxyserver"
$chef_metadata_url = "https://www.chef.io/chef/metadata?p=windows&m=x86_64&pv=2008r2&v=latest"
$chef_omnibus_root = "$env:systemdrive\opscode\chef"
$msi = "$env:TEMP\chef-true.msi"
$pretty_version = "install only if missing"
$version = "true"

Function Check-UpdateChef($root, $version) {
  if (-Not (Test-Path $root)) { return $true }
  elseif ("$version" -eq "true") { return $false }
  elseif ("$version" -eq "latest") { return $true }
  Try { $chef_version = Get-Content $root\version-manifest.txt | select-object -1}
  Catch {
    Try { $chef_version = (& $root\bin\chef-solo.bat -v).split(" ", 2)[1] }
    Catch { $chef_version = "" }
  }

  if ($chef_version.StartsWith($version)) { return $false }
  else { return $true }
}

Function Get-ChefMetadata($url) {
  Try { $response = ($c = Make-WebClient).DownloadString($url) }
  Finally { if ($c -ne $null) { $c.Dispose() } }

  $md = ConvertFrom-StringData $response.Replace("`t", "=")
  return @($md.url, $md.md5)
}

Function Get-MD5Sum($src) {
  Try {
    $c = New-Object -TypeName System.Security.Cryptography.MD5CryptoServiceProvider
    $bytes = $c.ComputeHash(($in = (Get-Item $src).OpenRead()))
    return ([System.BitConverter]::ToString($bytes)).Replace("-", "").ToLower()
  } Finally { if (($c -ne $null) -and ($c.GetType().GetMethod("Dispose") -ne $null)) { $c.Dispose() }; if ($in -ne $null) { $in.Dispose() } }
}

Function Download-Chef($md_url, $dst) {
  $url, $md5 = Get-ChefMetadata $md_url

  Try {
    Log "Downloading package from $url"
    ($c = Make-WebClient).DownloadFile($url, $dst)
    Log "Download complete."
  } Finally { if ($c -ne $null) { $c.Dispose() } }

  if (($dmd5 = Get-MD5Sum $dst) -eq $md5) { Log "Successfully verified $dst" }
  else { throw "MD5 for $dst $dmd5 does not match $md5" }
}

Function Install-Chef($msi) {
  Log "Installing Chef Omnibus package $msi"
  $p = Start-Process -FilePath "msiexec.exe" -ArgumentList "/qn /i $msi" -Passthru -Wait

  if ($p.ExitCode -ne 0) { throw "msiexec was not successful. Received exit code $($p.ExitCode)" }

  Remove-Item $msi -Force
  Log "Installation complete"
}

Function Log($m) { Write-Host "       $m`n" }

Function Make-WebClient {
  $proxy = New-Object -TypeName System.Net.WebProxy
  $proxy.Address = $env:http_proxy
  $client = New-Object -TypeName System.Net.WebClient
  $client.Proxy = $proxy
  return $client
}

Function Unresolve-Path($p) {
  if ($p -eq $null) { return $null }
  else { return $ExecutionContext.SessionState.Path.GetUnresolvedProviderPathFromPSPath($p) }
}

$chef_omnibus_root = Unresolve-Path $chef_omnibus_root
$msi = Unresolve-Path $msi

if (Check-UpdateChef $chef_omnibus_root $version) {
  Write-Host "-----> Installing Chef Omnibus ($pretty_version)`n"
  Download-Chef "$chef_metadata_url" $msi
  Install-Chef $msi
} else {
  Write-Host "-----> Chef Omnibus installation detected ($pretty_version)`n"
}
]
>>>>>> ----------------------
$ echo $?
10
$
```

Anybody have any ideas how to work around this?

mwrock:
Fixes #854. The current `install_command` emmited by mixlib-install almost fills the max length of the command line on windows. So if you need to add a proxy, the command is too long and the install fails.

This saves the command to a file and then just executes the call to the file.

### DIFF
--- a/lib/kitchen/transport/winrm.rb
+++ b/lib/kitchen/transport/winrm.rb
@@ -87,6 +87,10 @@ module Kitchen
         def execute(command)
           return if command.nil?
           logger.debug("[WinRM] #{self} (#{command})")
+
+          if command.length > MAX_COMMAND_SIZE
+            command = run_from_file_command(command)
+          end
           exit_code, stderr = execute_with_exit_code(command)
 
           if logger.debug? && exit_code == 0
@@ -133,6 +137,11 @@ module Kitchen
         private
 
         PING_COMMAND = "Write-Host '[WinRM] Established\n'".freeze
+
+        # Maximum string to send to the transport to execute. WinRM has an 8000 character
+        # command line limit. The original command string is coverted to a base 64 encoded
+        # UTF-16 string which will double the string size.
+        MAX_COMMAND_SIZE = 3000
 
         RESCUE_EXCEPTIONS_ON_ESTABLISH = lambda do
           [
@@ -375,11 +384,36 @@ module Kitchen
         def to_s
           "#{winrm_transport}::#{endpoint}<#{options.inspect}>"
         end
+
+        # takes a long (greater than 3000 characters) command and saves it to a
+        # file and uploads it to the test instance.
+        #
+        # @param command [String] a long command to be saved and uploaded
+        # @return [String] a command that executes the uploaded script
+        # @api private
+        def run_from_file_command(command)
+          temp_dir = Dir.mktmpdir("kitchen-long-script")
+          begin
+            script_name = "#{instance_name}-long_script.ps1"
+            script_path = File.join(temp_dir, script_name)
+
+            File.open(script_path, "wb") do |file|
+              file.write(command)
+            end
+
+            target_path = File.join("$env:TEMP", "kitchen")
+            upload(script_path, target_path)
+
+            %{& "#{File.join(target_path, script_name)}"}
+          ensure
+            FileUtils.rmtree(temp_dir)
+          end
+        end
       end
 
       private
 
-      WINRM_TRANSPORT_SPEC_VERSION = "~> 1.0".freeze
+      WINRM_TRANSPORT_SPEC_VERSION = ["~> 1.0", ">= 1.0.3"].freeze
 
       # Builds the hash of options needed by the Connection object on
       # construction.
@@ -429,7 +463,7 @@ module Kitchen
         spec_version = WINRM_TRANSPORT_SPEC_VERSION.dup
         logger.debug("Winrm Transport requested," \
           " loading WinRM::Transport gem (#{spec_version})")
-        gem "winrm-transport", spec_version
+        gem "winrm-transport", *spec_version
         first_load = require "winrm/transport/version"
         load_winrm_transport!
 
--- a/test-kitchen.gemspec
+++ b/test-kitchen.gemspec
@@ -31,7 +31,7 @@ Gem::Specification.new do |gem|
   gem.add_dependency "mixlib-install",  "~> 0.7"
 
   gem.add_development_dependency "pry"
-  gem.add_development_dependency "winrm-transport", "~> 1.0"
+  gem.add_development_dependency "winrm-transport", "~> 1.0", ">= 1.0.3"
 
   gem.add_development_dependency "bundler",   "~> 1.3"
   gem.add_development_dependency "rake"
--- a/lib/kitchen/provisioner/chef_base.rb
+++ b/lib/kitchen/provisioner/chef_base.rb
@@ -111,6 +111,11 @@ module Kitchen
 
         version = config[:require_chef_omnibus].to_s.downcase
 
+        # Passing "true" to mixlib-install currently breaks the windows metadata_url
+        # TODO: remove this line once https://github.com/chef/mixlib-install/pull/22
+        # is accepted and released
+        version = "" if version == "true"
+
         installer = Mixlib::Install.new(version, powershell_shell?, install_options)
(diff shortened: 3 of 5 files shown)


btm:
@mwrock rubocop is offended.

mwrock:
Yeah. I think I just fixed that. I'm also looking at generalizing the "execute from file" in such a way that is not provisioner specific.

mwrock:
This fix can now be applied to all winrm commands (provisioner, verifiers, other) and also now leaves SSH commands alone. I'm deliberately not puting the generalization in winrm-transport because I want to refactor/consolidate winrm/winrm-fs/winrm-transport before adding functionality there.

lib/kitchen/transport/winrm.rb
```diff
@@ -87,6 +87,10 @@ def close
         def execute(command)
           return if command.nil?
           logger.debug("[WinRM] #{self} (#{command})")
+
+          if command.length > MAX_COMMAND_SIZE
+            command = run_from_file_command(command)
```
tyler-ball: Why don't we just always run from a file?

tyler-ball: Also, feel free to disagree with me and leave this the way it is - my comment is pretty nitpicky.

test-kitchen.gemspec
```diff
@@ -31,7 +31,7 @@ Gem::Specification.new do |gem|
   gem.add_dependency "mixlib-install",  "~> 0.7"
 
   gem.add_development_dependency "pry"
-  gem.add_development_dependency "winrm-transport", "~> 1.0"
+  gem.add_development_dependency "winrm-transport", "~> 1.0.3"
```
tyler-ball: Can we do `'~> 1.0', '>= 1.0.3'` instead, so we aren't pinning to a minor version?

lib/kitchen/transport/winrm.rb
mwrock: thats a valid point. if you have alot of individual commands you are going to run separately then there is overhead creating files for each command. However compared to the overhead of going over a wire, thats extremely negligible.

test-kitchen.gemspec
mwrock: ah right. and we'd need the same when loading the gem at runtime

lib/kitchen/transport/winrm.rb
tyler-ball: I leave the decision in your capable hands

mwrock: I'm leaning to stay with this as for this PR as it may have some unforseen side effects if applied globally and I'd want to add tests and do a bunch of validation.

mwrock:
updated the gem constraints

PR #883, opened 2015-11-30
Adding support for a command prefix

adamleff:
Some platforms, such as Cisco devices, require commands to be
prefixed with a command to force the execution to take place within
a VRF. This allows the command to be configured in the .kitchen.yml.

### DIFF
--- a/lib/kitchen/transport/ssh.rb
+++ b/lib/kitchen/transport/ssh.rb
@@ -108,7 +108,7 @@ module Kitchen
         def execute(command)
           return if command.nil?
 
-          command = [ command_prefix, command ].compact.join(' ')
+          command = [command_prefix, command].compact.join(" ")
           logger.debug("[SSH] #{self} (#{command})")
 
           exit_code = execute_with_exit_code(command)
--- a/lib/kitchen/transport/ssh.rb
+++ b/lib/kitchen/transport/ssh.rb
@@ -48,6 +48,7 @@ module Kitchen
       default_config :connection_retries, 5
       default_config :connection_retry_sleep, 1
       default_config :max_wait_until_ready, 600
+      default_config :command_prefix, nil
 
       default_config :ssh_key, nil
       expand_path_for :ssh_key
@@ -106,7 +107,10 @@ module Kitchen
         # (see Base::Connection#execute)
         def execute(command)
           return if command.nil?
+
+          command = [ command_prefix, command ].compact.join(' ')
           logger.debug("[SSH] #{self} (#{command})")
+
           exit_code = execute_with_exit_code(command)
 
           if exit_code != 0
@@ -197,6 +201,10 @@ module Kitchen
         # @api private
         attr_reader :port
 
+        # @return [String] prefix for each executed command
+        # @api private
+        attr_reader :command_prefix
+
         # Establish an SSH session on the remote host.
         #
         # @param opts [Hash] retry options
@@ -268,6 +276,7 @@ module Kitchen
           @connection_retries     = @options.delete(:connection_retries)
           @connection_retry_sleep = @options.delete(:connection_retry_sleep)
           @max_wait_until_ready   = @options.delete(:max_wait_until_ready)
+          @command_prefix         = @options.delete(:command_prefix)
         end
 
         # Returns a connection session, or establishes one when invoked the
@@ -315,7 +324,8 @@ module Kitchen
           :timeout                => data[:connection_timeout],
           :connection_retries     => data[:connection_retries],
           :connection_retry_sleep => data[:connection_retry_sleep],
-          :max_wait_until_ready   => data[:max_wait_until_ready]
+          :max_wait_until_ready   => data[:max_wait_until_ready],
+          :command_prefix         => data[:command_prefix]
         }
 
         opts[:keys_only] = true                     if data[:ssh_key]
--- a/lib/kitchen/transport/winrm.rb
+++ b/lib/kitchen/transport/winrm.rb
@@ -88,7 +88,7 @@ module Kitchen
         def execute(command)
           return if command.nil?
 
-          command = [ command_prefix, command ].compact.join(' ')
+          command = [command_prefix, command].compact.join(" ")
           logger.debug("[WinRM] #{self} (#{command})")
 
           if command.length > MAX_COMMAND_SIZE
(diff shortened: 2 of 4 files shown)


tyler-ball:
Slacked with @adamleff about this - we're going to see if the `sudo_command` config on the verifier and provisioner will work for him

adamleff:
Unfortunately, `sudo_command` doesn't fit the bill.  For the particular platform we're working on (Cisco Nexus), the particular command prefix must apply to all commands, and `sudo_command` is applied selectively.

For example, when installing the omnibus chef package on an instance:

https://github.com/chef/mixlib-install/blob/14611366684808e4db54dfaeee8ca49a8816a4e6/support/install_command.sh#L218-L219

The `install_command` is created via a Mixlib::Install instance:

https://github.com/chef/mixlib-install/blob/14611366684808e4db54dfaeee8ca49a8816a4e6/support/install_command.sh#L218-L219

And the `install_options` hash contains the `sudo_command` value:

https://github.com/test-kitchen/test-kitchen/blob/5e4fa6baff5d51dc983be3f7b67ffe1bcbf2a8d2/lib/kitchen/provisioner/chef_base.rb#L130

The `sudo_command` value is used selectively in the install_command:

https://github.com/chef/mixlib-install/blob/14611366684808e4db54dfaeee8ca49a8816a4e6/support/install_command.sh#L218-L219

On Cisco, without using the command prefix that runs this command in the management VRF, the `do_download` fails due to lack of DNS:

```
       Downloading https://www.chef.io/chef/install.sh to file /tmp/install.sh
       Trying wget...
       Trying curl...
       Trying python...
       Trying perl...
       Unable to download https://www.chef.io/chef/install.sh to /tmp/install.sh, aborting
       \nDEBUG OUTPUT FOLLOWS:\n\nSTDERR from wget:\n\n--2015-11-30 21:46:50--  https://www.chef.io/chef/install.sh
       Resolving www.chef.io... failed: Name or service not known.
       wget: unable to resolve host address `www.chef.io'\n\nSTDERR from curl:\n\n\n\nSTDERR from python:\n\nTraceback (most recent call last):
         File "<string>", line 1, in <module>
         File "/usr/lib64/python2.7/urllib2.py", line 126, in urlopen
           return _opener.open(url, data, timeout)
         File "/usr/lib64/python2.7/urllib2.py", line 394, in open
           response = self._open(req, data)
         File "/usr/lib64/python2.7/urllib2.py", line 412, in _open
           '_open', req)
         File "/usr/lib64/python2.7/urllib2.py", line 372, in _call_chain
           result = func(*args)
         File "/usr/lib64/python2.7/urllib2.py", line 1207, in https_open
           return self.do_open(httplib.HTTPSConnection, req)
         File "/usr/lib64/python2.7/urllib2.py", line 1174, in do_open
           raise URLError(err)
       .>\n\nSTDERR from perl:\n\nCan't locate LWP/Simple.pm in @INC (@INC contains: /etc/perl /usr/lib64/perl/site_perl/5.14.2/ /usr/lib64/perl/site_perl/5.14.2 /usr/lib64/perl/vendor_perl/5.14.2/ /usr/lib64/perl/vendor_perl/5.14.2 /usr/lib64/perl/5.14.2/ /usr/lib64/perl/5.14.2 /usr/local/lib/site_perl /usr/lib64/perl/5.14.2 .) at -e line 1.
       BEGIN failed--compilation aborted at -e line 1.\n
D      Cleaning up local sandbox in /var/folders/v5/z54gb76j2rs3wrn65hmtyf1r0000gp/T/default-nexus-sandbox-20151130-61305-1plv3kx
>>>>>> Converge failed on instance <default-nexus>.
```

This is why I made my change at the transport layer for those transport connections that won't be able to execute any command successfully without such a prefix added. While I think the `sudo_command` on the provisioner and verifier would work, we'd have to modify Mixlib::Install to prefix like 218 above with `$sudo_sh` as well so the curl/wget/perl fetching of the install.sh script happens with the `sudo_command` prefix too.  Not sure if that's the right call here or not.

tyler-ball:
Mixlib::Install is going to stop accepting a sudo-or-not-sudo argument when it releases 1.0 (paging @thommay or @sersut for confirmation).  We also don't need it - once we have the install command from Mixlib::Install we can chose to run it as sudo or not.

Im fully in agreement that we need the ability to prepend commands TK runs with some prefix.  But I think thats what the purpose of `sudo_command` is.

I think the proper refactor is to:
1.  Stop passing `sudo_command` to Mixlib::Install because it will stop accepting it eventually
2.  Instead, run the [`install_command`](https://github.com/test-kitchen/test-kitchen/blob/5e4fa6baff5d51dc983be3f7b67ffe1bcbf2a8d2/lib/kitchen/provisioner/chef_base.rb#L109) using the `sudo_command` argument.
3.  Further, all [commands](https://github.com/test-kitchen/test-kitchen/blob/5e4fa6baff5d51dc983be3f7b67ffe1bcbf2a8d2/lib/kitchen/provisioner/base.rb#L64-L72) executed by the provisioner (and verifier and busser) should be prefixed with the `sudo_command`.  
4.  The Transport class is the wrong place to do this.  Yes, the Transport class handles executing the commands and reading their response but it is not intended to modify the commands at all.  It simply handles executing whatever it is given.

I'm making an assumption here - that all commands should be prefixed with `sudo_command`.  @fnichol or someone else with more Test Kitchen experience will have to validate that.  I don't have time right now to dive in and do the research.

If all commands should not be prefixed with `sudo_command` (if it is to be used selectively) then I agree that we need to introduce another configuration similar to what you initially proposed.  Still, the wrong place to do that is the Transport class.  We would need to add this `command_prefix` configuration to the Provisioners and Verifiers (and any other classes I am forgetting which perform remote execution).

adamleff:
For my own edification, can you explain why the Transport class is the wrong place for this if the argument is that all commands should be prefixed with `sudo_command`?  That's the one place where all edges meet vs. having to implement the prefix logic in N places.

tyler-ball:
Just because it can does not mean that it should.  It comes down to a separation of concerns.  The Transport class _should not_ have any clue what it is transporting.  Re:



It is scope creep for the Transport class to start inspecting the items it is transporting, and thats bad OO design.

The best class to modify the messages is probably the Instance class before it invokes its Transport - the Instance knows things like 'I am a Cisco box, therefor I need to prefix *all* my commands'.  Seems like a good place to use the [interceptor pattern](https://en.wikipedia.org/wiki/Interceptor_pattern) or possibly the [decorator pattern](https://en.wikipedia.org/wiki/Decorator_pattern).

adamleff:
That's fair, thanks :)  Except we don't really have a way to decorate "instances" per se as the concept of an instance is contrived by Test Kitchen (platform + suite), not the human configuring it.  But given that you can specify config items globally or at a per-platform scope, I'm on board with your initially proposed refactor.

I'll get to work on that today and close this out for now.  Thanks, @tyler-ball!

PR #885, opened 2015-12-02
Running the chef_base provisioner install_command via sudo, and command_prefix support

adamleff:
Currently, whether or not to use sudo during the installation of
the omnibus chef package is left up to the install command
generated by Mixlib::Install based on the sudo flag and command
passed to the constructor. Future versions of Mixlib::Install
do not support the sudo option, and some platforms (such as Cisco
Nexus) require the entire process to run with escalated privileges
in order to obtain outbound internet access. This change tells
Mixlib::Install to not care about sudo and instead wrap the entire
install_command with sudo from within the provisioner.

On certain platforms, commands must be prefixed with another
command to work properly, such as on Cisco Nexus where the
user must switch into a management vrf to obtain outbound network
access. Using the sudo_command config parameter is not sufficient
as some commands, such as busser verifier commands and provisioner
init commands, need to run as the non-privileged user.

This change allows the user to configure a string that will be
prepended to provisioner and verifier commands, much like how
sudo_command works, but impacts all commands.

### DIFF
--- a/lib/kitchen/provisioner/chef_base.rb
+++ b/lib/kitchen/provisioner/chef_base.rb
@@ -102,7 +102,7 @@ module Kitchen
           init_command_vars_for_bourne(dirs)
         end
 
-        shell_code_from_file(vars, "chef_base_init_command")
+        prefix_command(shell_code_from_file(vars, "chef_base_init_command"))
       end
 
       # (see Base#install_command)
@@ -118,7 +118,7 @@ module Kitchen
 
         installer = Mixlib::Install.new(version, powershell_shell?, install_options)
         config[:chef_omnibus_root] = installer.root
-        installer.install_command
+        prefix_command(sudo(installer.install_command))
       end
 
       private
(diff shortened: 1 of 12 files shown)


lib/kitchen/provisioner/chef_base.rb
```diff
@@ -131,8 +131,8 @@ def install_options
           :omnibus_url => config[:chef_omnibus_url],
           :project => project.nil? ? nil : project[1],
           :install_flags => config[:chef_omnibus_install_options],
-          :use_sudo => config[:sudo],
-          :sudo_command => config[:sudo_command]
+          :use_sudo => false,
```
tyler-ball: Instead of setting `false` and `nil` here, lets just leave them out - Mixlib::Install will use the defaults we're already specifying and then we have less code to migrate later.

tyler-ball:
Other than 1 comment, :+1: 

Hopefully those travis failures go away after the next commit

lib/kitchen/provisioner/chef_base.rb
adamleff: `:use_sudo` actually defaults to `true` in Mixlib::Install which is why I was explicitly setting it to false, avoiding a double-sudo. This isn't really a problem unless someone is setting a wacky `sudo_command`.  I'll make the change, test it on a few platforms, and push it up.  I'll hold-off on squashing the commits in case there is any other feedback from the other maintainers.

adamleff:
Travis failures appear to be unrelated to this change. I intermittently received the same failures testing locally prior to my changes.

tyler-ball:
@adamleff Hmm, well, we will have to get those failures addressed before we can merge this.  Or we can roll the Travis dice a few times and see if it goes green :)

adamleff:
Travis has re-run and it's all green now: https://travis-ci.org/test-kitchen/test-kitchen/builds/94438660

PR #895, opened 2015-12-17
Adding in ChefConfig support to enable loading proxy config from chef config files

tyler-ball:
Also added support for pushing ftp_proxy to target machines (because this was a quick fix in the code I was at)

\cc @fnichol @randomcamel @smith @juliandunn @jkeiser


randomcamel:
:+1: Nice.

danielsdeleo:
:+1:

PR #898, opened 2015-12-22
Adding the CHANGELOG and version.rb update for 1.5.0.rc.1

tyler-ball:
\cc @fnichol @jkeiser @adamedx @randomcamel @danielsdeleo

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,29 @@
-## 1.4.2
+## 1.5.0
+
+### Potentially breaking changes
+
+### Bug fixes
+
+* PR [#816][]: Fix SuSe OS Busser install ([@Peuserik][])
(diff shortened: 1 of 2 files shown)


danielsdeleo:
The travis failure looks to me like a real thing, but I don't know what to make of it from just looking at the travis matrix, since the same ruby version is passing when used with a different rubygems version. If it's a real thing, probably worth fixing that, then rebasing this on top. But :+1: on the changes in this patch

smurawski:
Any chance of getting my winrm PR into this?  By now it needs a rebase but I can do that today if it'll get into the next release.  I'm blocked on a couple of customer scenarios since transport only does unencrypted winrm.

tyler-ball:
@smurawski This is just the RC that I'm trying to get out - if your PR doesn't make it into the RC it could still make it into the release.  Just ping me when its ready for review

PR #899, opened 2015-12-30
fix chef install on non-windows

mwrock:
It appears that my commit https://github.com/test-kitchen/test-kitchen/commit/7318fa5bb93319b68303b0f93302f1c35bb8e562 added a hack to fix windows chef installs due to an issue in mixlib-install but broke non-windows in the process. This PR rectifies breaks non-windows.

### DIFF
--- a/lib/kitchen/provisioner/chef_base.rb
+++ b/lib/kitchen/provisioner/chef_base.rb
@@ -132,7 +132,7 @@ module Kitchen
         # Passing "true" to mixlib-install currently breaks the windows metadata_url
         # TODO: remove this line once https://github.com/chef/mixlib-install/pull/22
         # is accepted and released
-        version = "" if version == "true"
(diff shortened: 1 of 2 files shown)


coderanger:
:+1:

scotthain:
:+1:

martinb3:
:+1: master is currently broken for my Linux workstation until this is fixed.

martinb3:
Using "-v <empty string>" still reads kind of oddly to me. I've tried `version = nil if version == "true"` and it also works for Linux (haven't tested Powershell/Windows), and feels more sane than passing an empty string to a version param on Windows.

mwrock:
Yeah its definitely weird. I think the `nil` breaks powershell but would need to double check. This line will be removed when a fixed mixlib-install is available.

PR #901, opened 2016-01-02
Add support for --profile-ruby

martinb3:
Add a special option to the chef provisioners (both solo and zero) for supporting the new --profile-ruby argument [added in 12.6.0](https://github.com/chef/chef/blob/master/CHANGELOG.md#1260) to the chef client. This is extremely useful for profiling Chef when doing development.


akemner:
this would be nice!

adamleff:
Great idea - thanks for the contribution.  :+1:

cheeseplus:
+1

martinb3:
Thanks!

PR #906, opened 2016-01-11
Adding Maintainers file

cheeseplus:
As per Rally discussions - adding new fleet of maintainers.

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -1,10 +1,9 @@
 # Maintainers
 
-This file lists how the Test-Kitchen project is maintained. When making changes
-to the system, this file tells you who needs to review your patch - you need a
-simple majority of maintainers for the relevant subsystems to provide a :+1: on
-your pull request. Additionally, you need to not receive a veto from a
-Lieutenant or the Project Lead.
+This file lists how the Test Kitchen project is maintained. When making changes
+to the system, this file tells you who needs to review your patch. You need a :+1: vote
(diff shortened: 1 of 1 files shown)


mwrock:
:+1:

smurawski:
:+1:

adamleff:
With a growing maintainers list, obtaining a "simple majority" may be time-consuming... we may want to reconsider that at some point.  :+1: regardless :)

mwrock:
Or do we go with the rule of 2 +1s. I think that's preferable.

adamleff:
That would be my preference as well. Even though it's not an "official Chef project," such a policy would put it more in line with RFC 030. We could even state here that we're following that model.

cheeseplus:
To be clear I was cargo culting RFC30 as the skeleton so I'm happy to acknowledge the influence - just wanted to make clear the non-official support. Suggestions for text/PR?

adamleff:
When RFC030 was accepted, we did not go back and modify many of the MAINTAINERS.md files in our projects, so they're kind of in conflict. :smile: 

Proposed:

```
This file lists how the Test Kitchen project is maintained. When making changes
to the system, this file tells you who needs to review your patch. You need at least
2 :+1: votes from the users listed below. Additionally, you need to not receive a veto
from the Lieutenant or the Project Lead.
```

cheeseplus:
Done

adamleff:
:rocket: :cake: :100:

smurawski:
:+1:

PR #907, opened 2016-01-11
Set version to 1.5.0

jkeiser:
Having a non-RC version is important to make bundle installs in omnibus work (in the chef-dk). We will not *release* 1.5.0 until we actually tag it.


tyler-ball:
:+1: As soon as travis is happy :shipit:

PR #909, opened 2016-01-11
Revert "Updating to the latest release of net-ssh to consume https://github.com/net-ssh/net-ssh/pull/280"

tyler-ball:
Reverts test-kitchen/test-kitchen#908


cheeseplus:
We fixed this so I guess we don't need it.

tyler-ball:
@cheeseplus Yup!  I thought for a minute we were going to, but then we rolled forward with the correct fix!

PR #911, opened 2016-01-13
Diet travis

cheeseplus:
Slimming down platforms to latest and something a bit more reasonable.


coderanger:
:+1: and we should turn on nightli.es for safety.

cheeseplus:
@coderanger http://nightli.es enabled on the repo \o/

tyler-ball:
So many :+1:

PR #924, opened 2016-01-25
Make chef-config an optional dependency.

coderanger:
This helps with working with older versions of Chef or using Test Kitchen in a non-Chef environment.

Ping @cheeseplus for review.


cheeseplus:
Still getting some RuboCop failures after kicking the build

coderanger:
The remaining failures don't seem to be related.

cheeseplus:
Yep, that remaining failure seems to to be from the fact it's another repo. +1

PR #925, opened 2016-01-27
bypass execution policy when running powershell script files

mwrock:
If a test instance is using the default execution policy, executing unsigned script files will fail. This bypasses that policy just for the files that test-kitchen pushes


mwrock:
Test failure seems unrelated to change

sean-horn:
Thanks for this, @mwrock !

A customer ran into it in real life the other day trying to bootstrap win 2012r2 systems in AWS and worked around the issue by falling back to TK 1.4.2

chris-rock:
:+1:

PR #927, opened 2016-01-27
Chef's default log level is warn, not info.

### LINKED ISSUE #529: Cannot set log level to "info"
As noticed in #478 & AFAIK, you cannot currently set log level to "info" as test-kitchen's chef solo and zero provisioners crowbar the setting to "auto" before forwarding to chef (see https://github.com/test-kitchen/test-kitchen/blob/a83398eba6d850971805447a42270925847acc08/lib/kitchen/provisioner/chef_solo.rb#L44), and chef interprets 'auto' as 'warn'.

This is currently more than "a bit annoying" as @danielsdeleo commented.  If you really wanted to trim down the log verbiage, change the default test-kitchen input from "info" to "warn", but please allow those of us who want it to get it rather than have to wade through debug.

Will happily submit a PR to remove the crow-baring in the line noted above & chef-zero if wanted.  Please advise on whether I should change the default TK log level to compensate at the same time or not.  Would suggest yes, as that would match Chef.

axos88:
Fixes #529

### DIFF
--- a/lib/kitchen/provisioner/chef_solo.rb
+++ b/lib/kitchen/provisioner/chef_solo.rb
@@ -47,7 +47,7 @@ module Kitchen
 
       # (see Base#run_command)
       def run_command # rubocop:disable Metrics/AbcSize
-        level = config[:log_level] == :info ? :auto : config[:log_level]
+        level = config[:log_level] == :warn ? :auto : config[:log_level]
 
         cmd = sudo(config[:chef_solo_path]).dup.
           tap { |str| str.insert(0, "& ") if powershell_shell? }


cheeseplus:
If you could alter the tests to check for the new level properly I can get this merged :confetti_ball:

alexpop:
Please also make the change to `provisioner/chef_zero.rb`

I've fixed the tests as well, you can see the entire diff here:
https://github.com/test-kitchen/test-kitchen/compare/master...alexpop:ap/chef_log

Let me know if you need more help here. I want to see this change merged. Cheers.

cheeseplus:
This only fixes for chef_solo and also doesn't fix the root issue which is that there need to be two separate log levels as described in some other issues. Going to close this in favor of the approach here https://github.com/test-kitchen/test-kitchen/pull/950

PR #294, opened 2013-12-19
SSH gateway support

coderanger:
Creating this for tracking, not actually ready for merge yet.

PR #933, opened 2016-02-04
SSH gateway support redux

andrewjamesbrown:
This PR is an rewrite/extension of https://github.com/test-kitchen/test-kitchen/pull/294, that brings it up to date with the current API.

It supports ssh gateways for converging, verifying, and login to TK instances.
I've tested with a legacy driver, which works, but I have limited access to other drivers.

### DIFF
--- a/lib/kitchen/driver/ssh_base.rb
+++ b/lib/kitchen/driver/ssh_base.rb
@@ -306,8 +306,6 @@ module Kitchen
         pseudo_state[:username] = username if username
         pseudo_state.merge!(options)
 
-        # TODO: 
-        # SSH.new(*build_ssh_args(options.merge(:hostname => hostname))).wait
         instance.transport.connection(backcompat_merged_state(pseudo_state)).
           wait_until_ready
       end
--- a/lib/kitchen/driver/ssh_base.rb
+++ b/lib/kitchen/driver/ssh_base.rb
@@ -229,6 +229,7 @@ module Kitchen
         opts[:port] = combined[:port] if combined[:port]
         opts[:keys] = Array(combined[:ssh_key]) if combined[:ssh_key]
         opts[:logger] = logger
+        opts[:gateway] = combined[:ssh_gateway] if combined[:ssh_gateway]
 
         [combined[:hostname], combined[:username], opts]
       end
@@ -305,6 +306,8 @@ module Kitchen
         pseudo_state[:username] = username if username
         pseudo_state.merge!(options)
 
+        # TODO: 
+        # SSH.new(*build_ssh_args(options.merge(:hostname => hostname))).wait
         instance.transport.connection(backcompat_merged_state(pseudo_state)).
           wait_until_ready
       end
--- a/lib/kitchen/driver/ssh_base.rb
+++ b/lib/kitchen/driver/ssh_base.rb
@@ -217,7 +217,7 @@ module Kitchen
       # @param state [Hash] state hash
       # @return [Array] SSH constructor arguments
       # @api private
-      def build_ssh_args(state)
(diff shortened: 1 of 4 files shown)


coderanger:
This has the same problem as last time, we need a way to know if the remote side is up.

andrewjamesbrown:
Can you help me understand where its missing?  This is from a machine where I deliberately disabled any networking:

```
-----> Converging <default-ubuntu-1204>...
       Preparing files for transfer
       Preparing dna.json
       Resolving cookbook dependencies with Berkshelf 4.0.1...
       Preparing validation.pem
       Preparing client.rb
       [SSH] connection failed, retrying in 1 seconds (#<Timeout::Error: execution expired>)
       [SSH] connection failed, retrying in 1 seconds (#<Timeout::Error: execution expired>)
       [SSH] connection failed, retrying in 1 seconds (#<Timeout::Error: execution expired>)
       [SSH] connection failed, retrying in 1 seconds (#<Timeout::Error: execution expired>)
$$$$$$ [SSH] connection failed, terminating (#<Timeout::Error: execution expired>)
```

coderanger:
https://github.com/test-kitchen/test-kitchen/pull/933/files#diff-8f274fcf245bb8ca455e8f272baa9149R288 specifically. I guess we could ignore that because the actual SSH transport doesn't use that code path anymore?

andrewjamesbrown:
Got it.  We can probably ignore it, though it does feel like a wart that wants fixing.
I could tunnel through the gateway and use `netcat` to check if the port is up, but that also seems ugly.

Thoughts on the approach?  Is it worthwhile pursuing?

andrewjamesbrown:
@coderanger I've added a rudimentary test for remote availability through an ssh gateway.  I'm not super happy with it - its ugly and relies on using netcat to test the remote end.  My attempts to do the same using ssh tunnels failed miserably, so I used this instead.

pburkholder:
Since these test are passing (except the ones that can't pass because Ec2 instances won't spin up) could someone comment on the quality/acceptability of this code?

andrewjamesbrown:
Is this stalled?  Anything I can do to move it along?

iennae:
@andrewjamesbrown, 

Thanks for the contribution here. I'll add this to our review list for our weekly test kitchen maintainers meeting on 3/30. Apologies for the delay in response. We've been trucking through catching up on PRs for all of the test-kitchen products, but we haven't been able to capture them all. I'll give an update next week. (or one of my co-maintainers will before then).

Thank you so much for your patience. We really appreciate all the community involvement and passion around test-kitchen.

@coderanger, 

Did you have further thoughts on this PR? Would love to get more of your feedback, especially as you had started this work with PR #294. 

Thanks!

cheeseplus:
Just wanted to follow up to say this didn't come up at our 3/30 meeting due to a several other priority issues / bugs.

While I'd love a cleaner and more portable test than using `netcat` I realize the value of the feature is huge but I'd also appreciate @coderanger weighing in here before we merge.

poliva83:
Any update on this PR?

coderanger:
I don't think depending on `nc` on the gateway machine is viable. We should do the same thing net-ssh-gateway does which is to use a port forward (https://github.com/net-ssh/net-ssh-gateway/blob/master/lib/net/ssh/gateway.rb#L109) and then use a TCP connection as before.

EYurchenko:
test_ssh is indirectly called only from spec tests. So why it exists at all? what am I missing?

coderanger:
Closing this because #1091 was merged. Whatever issues I was seeing must have gone away elsewhere (maybe in the SSH core rewrite).

PR #937, opened 2016-02-08
properly initialize attributes for new negotiate

chris-rock:
this `negotiate` was a not a symbol, properties have not been initialized properly. We figured this during our units tests in https://github.com/chef/kitchen-inspec


chris-rock:
cc @mwrock

mwrock:
:+1: 

https://github.com/test-kitchen/test-kitchen/blob/chris-rock/winrm-properties-fix/lib/kitchen/transport/winrm.rb#L61 convrts `to_sym` but it is definitely better to initialize as a symbol.

chris-rock:
we cannot be sure that users really use a symbol

Issue #428, opened 2014-05-05
Add Berksfile location config option


sonnysideup:
The provisioner is hard-coded to read the Berksfile from the kitchen root (https://github.com/test-kitchen/test-kitchen/blob/master/lib/kitchen/provisioner/chef_base.rb#L290-L291). It would be nice to be able to use a different path in the `.kitchen.yml` file.

chrisfarnham:
Without this feature we cannot use test-kitchen in our current environment

mstyne:
:+1:

gled-rs:
done here:
https://github.com/gled-rs/test-kitchen-ez/commit/514a43a483d227170a042c16f7b5d48f5b2e7e49

if you want a PR, just ask, the previous one on that subject was closed.

cheeseplus:
I'd happily review/merge a PR for this!

tas50:
This has already been shipped in with the berksfile_path config

PR #939, opened 2016-02-08
Added berksfile_path option to provisioner, to allow selecting a diff…

gled-rs:
…erent Berksfile, or one that is not in the current dir

As asked by #428

### DIFF
--- a/lib/kitchen/provisioner/chef_base.rb
+++ b/lib/kitchen/provisioner/chef_base.rb
@@ -86,6 +86,16 @@ module Kitchen
       end
       expand_path_for :clients_path
 
+      default_config :berksfile_path do |provisioner|
+        provisioner.calculate_path("berksfile")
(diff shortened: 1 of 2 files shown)


lib/kitchen/provisioner/chef_base.rb
```diff
@@ -86,6 +86,16 @@ class ChefBase < Base
...
 
+      default_config :berksfile_path do |provisioner|
+        provisioner.calculate_path("berksfile")
+      end
+      expand_path_for :berksfile_path
+
+      default_config :cookbooks_path do |provisioner|
+        provisioner.calculate_path("cookbooks")
+      end
+      expand_path_for :cookbooks_path
```
lamont-granquist: cookbooks_path seems to have nothing to do with this (and may be redundant?)

Issue #931, opened 2016-02-03
Shell provisioner fails on Windows host to Linux guest


b00ga:
I have a mixed environment (Linux, Mac, Windows). I've been doing a lot of work with Test Kitchen on Mac and Linux. I'm getting a lot of my developer's on the Test Kitchen bandwagon.

Today I tried testing a basic kitchen config with virtualbox/vagrant/shell provisioner/bats test on a Windows machine to document it for my Windows developers.

The shell provisioner fails with "sudo: /tmp/kitchen/bootstrap.sh: command not found"
Windows7 / VirtualBox 5.0.14 / vagrant 1.8.1 / ruby 2.2.3 / Test Kitchen 1.5.0 running in a Git Bash shell (git version 2.7.0.windows.2). CentOS and Ubuntu LTS platforms from internally maintained vagrant box images. Same .kitchen.yml works fine on RHEL/CentOS & Mac OS X.

After doing some debugging, I've found the issue, but not sure what the right way to fix it is. The underlying issue is that the bootstrap script that is copied into the guest VM has 0644 permissions (no executable bit set).

The code path I'm looking at seems to start in lib/kitchen/provisioner/shell.rb. The host sets up a sandbox of files to copy over to the guest. Looking at the prepare_script() function, the bootstrap script gets copied to the host's sandbox dir and then FileUtils.chmod is called to make sure it is executable (0755). IIUC, FileUtils.chmod doesn't do anything on Windows (my host), so when the script is copied into the Linux guest, there is no Unix executable bit for Net::SCP (who is ultimately doing the copy) to send along with the file. (ls -l /tmp/kitchen/bootstrap.sh inside the guest confirms 0644 permissions).

After a failed converge, if I 'kitchen login' and then 'sudo chmod 755 /tmp/kitchen/bootstrap.sh' and rerun the converge, it succeeds (the scp overwrite the bootstrap, but doesn't change the file mode for what it copies over).

Maybe the chmod can be moved to be executed in the guest? Maybe the bootstrap can explicitly be called as "sh bootstrap.sh"? 

-Shawn

b00ga:
Not sure if the solution is ideal, but I tried changing in provisioner/shell.rb in run_command,
`code = powershell_shell? ? %{& "#{script}"} : sudo(script)`
to
`code = powershell_shell? ? %{& "#{script}"} : sudo("sh #{script}")`
[...]

PR #942, opened 2016-02-12
Bootstrap with command interpreter

b00ga:
This is a stab at introducing a feature that would provide a fix to #931 and is an alternative idea to the fix proposed in PR #442 
I tested with a little test rig I setup and made available here: https://github.com/b00ga/kitchen-bootstrap-testrig
This is my first time making any remotely significant changes to ruby code, so please be kind. Suggestions, discussion appreciated.

-Shawn


b00ga:
PR #1381 now addresses this resolving issue #931

Issue #917, opened 2016-01-17
ability to pass arguments to shell provisioner


mmckinst:
It would be great if you could specify a list of arguments for the shell provisioner to pass to the script. Something like this config would run the script below:

``` yaml
provisioner:
  name: shell
  arguments:
    - -t foobar
    - -n 6
    - --assume-yes
    - build
```

``` shell
bootstrap.sh -t foobar -n 6 --assume-yes build
```

cheeseplus:
If you'd like to make a PR we're happy to review.

mmckinst:
I ended up doing it so its just a key:value instead of an array, so it looks more like this and will result in the below script

```yaml
provisioner:
  name: shell
  arguments: '--php 70 --mysql 57'
```

```bootstrap.sh --php 70 --mysql 57```

I don't know if this works with Powershell or not, nor do I have a way of testing.

Let me know if it looks good and I can document it too if you want.

**Comment by mmckinst** [comment]
regarding the failing tests, it appears all the pull requests are failing for that tests.

**Comment by mmckinst** [comment]
Also, here's a better example of how this might be used IRL if you have a script to install a certain version of PHP on a bunch of different platforms.

I don't think the fact you get set attributes for the provisioner inside the suite is well documented, but its a great feature.

```yaml
provisioner:
  name: shell
  script: install.sh

platforms:
  - name: centos-6
  - name: centos-7
  - name: ubuntu-14.04
  - name: debian-7
  - name: debian-8

suites:
  - name: php56
    provisioner:
      arguments: '--php 56'  
  - name: php70
    provisioner:
      arguments: '--php 70'
```

logicminds:
What ever happen to this?  I need to pass in arguments as well or at least some environment variables.

mmckinst:
PR is passing tests and ready to be merged at https://github.com/test-kitchen/test-kitchen/pull/943

cheeseplus:
Merged!

PR #943, opened 2016-02-13
allow command line arguments

mmckinst:
closes issue #917

### DIFF
--- a/lib/kitchen/provisioner/shell.rb
+++ b/lib/kitchen/provisioner/shell.rb
@@ -74,6 +74,11 @@ module Kitchen
           config[:root_path],
           File.basename(config[:script])
         )
+
+        if config[:arguments]
+          script.concat(" ").concat(config[:arguments])
+        end
+
         code = powershell_shell? ? %{& "#{script}"} : sudo(script)
 
         prefix_command(wrap_shell_code(code))
--- a/spec/kitchen/provisioner/shell_spec.rb
+++ b/spec/kitchen/provisioner/shell_spec.rb
@@ -314,6 +314,11 @@ describe Kitchen::Provisioner::Shell do
 
         cmd.must_match(/^TEST=yes/)
       end
+
+      it "accepts arguments when configured" do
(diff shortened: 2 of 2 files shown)


cheeseplus:
If you could rebased I'd be happy to get this merged in - that or we'll fix it up as it's a good contribution.

mmckinst:
@cheeseplus I rebased. travis is still failing but appears to be failing on the current master branch already.

logicminds:
@mmckinst hey can you rebase this?  The travis and Gemfile have likely changed causing the build to fail.

mmckinst:
all tests passing now. @cheeseplus can you merge this in?

bad-beat:
Great feature! And at exactly the right time :)

Only found one issue, the double quotes around the powershell code does not work, the script is not executed in my case. Without the double quotes around the script and the arguments works fine.

New to GitHub, should I register a new issue or can this be done under this issue?

mmckinst:
@bad-beat can you post an example what you mean? like post the yaml file without the double quotes that doesn't work and post the one with the double quotes that does work?

bad-beat:
@mmckinst when I have the arguments in my yaml file as shown below. I use the same arguments list for both linux and windows machines.

In bootstrap.sh the arguments are added after the script and they are passed along just fine.
In bootstrap.ps1 the arguments are added after the script, but the whole set of script and arguments are encapsulated by double quotes by this code `code = powershell_shell? ? %{& "#{script}"} : sudo(script)` (from lib/kitchen/provisioner/shell.rb on line 82).
By adding the arguments in my yaml file, the bootstrap.ps1 is no longer executed at all. When I change the shell.rb code so it does not encapsluate the script and arguments in double quotes, the arguments are passed along just fine.

`provisioner:
  name: shell
  privileged: true
  arguments: '-version 01020304'`

mmckinst:
@bad-beat I don't know enough about powershell to debug this. You should create a separate issue for this or submit a PR if your change fixes the problem and won't break anything else.

PR #944, opened 2016-02-15
allow non-busser verifier to work with legacy drivers

chris-rock:
This allows to `kitchen-inspec` to work with multiple legacy drivers that use ssh as a transport.

### DIFF
--- a/lib/kitchen/instance.rb
+++ b/lib/kitchen/instance.rb
@@ -400,12 +400,13 @@ module Kitchen
       banner "Verifying #{to_str}..."
       elapsed = action(:verify) do |state|
         # use special handling for busser
-        if (
-          legacy_ssh_base_driver? &&
-          !defined?(Kitchen::Verifier::Busser).nil? &&
-          verifier.is_a?(Kitchen::Verifier::Busser)
-        )
-           legacy_ssh_base_verify(state)
+        if legacy_ssh_base_driver? &&
+            (
+              (!defined?(Kitchen::Verifier::Busser).nil? &&
+                verifier.is_a?(Kitchen::Verifier::Busser)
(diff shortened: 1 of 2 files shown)


chris-rock:
@tyler-ball and @cheeseplus Do you see any issues with this approach?

lib/kitchen/instance.rb
```diff
@@ -396,11 +396,23 @@ def setup_action
     # @see Driver::Base#verify
     # @return [self] this instance, used to chain actions
     # @api private
-    def verify_action
+    def verify_action # rubocop:disable Metrics/PerceivedComplexity, Metrics/CyclomaticComplexity
       banner "Verifying #{to_str}..."
       elapsed = action(:verify) do |state|
-        if legacy_ssh_base_driver?
+        # use special handling for busser
```
adamleff: I think we could make this more readable by adding some helper methods.

```
def verifier_busser?(verifier)
   # check if is a Kitchen::Verifier::Vusser
end

def verifier_dummy?(verifier)
  # check if verifier is a Kitchen::Verifier::Dummy
end

def use_legacy_ssh_verifier?(verifier)
  verifier_busser? || verifier_dummy?
end

if legacy_ssh_base_driver? && use_legacy_ssh_verifier?(verifier)
  legacy_ssh_base_verify(state)
elsif legacy_ssh_base_driver?
  verifier.call(driver.legacy_state(state))
else
  verifier.call(state)
end
```

I guess you could classify this as a nitpick, but it reads much cleaner to me and is easier to test.

chris-rock: @adamleff love it, will add this. thanks

adamleff:
Seems like a fine change to me.  I'm :+1:.

PR #946, opened 2016-02-17
add support for inspec_tests in suites

chris-rock:
This adds a new flag to `inspec_tests` to suites and copies the value over to verifier. This allows us to reuse InSpec profiles for multiple cookbooks in kitchen-inspec

```
suites:
  - name: profile
    run_list:
      - recipe[os_prepare]
    inspec_tests:
      - https://github.com/nathenharvey/tmp_compliance_profile
```


chris-rock:
An alternative that may be more generic: We could base the complete `suite` attribute tree to the verifier as well. This allows us to not make additions for one specific verifier.

PR #950, opened 2016-02-26
Stop log_level being copied from base config into provisioner config

drrk:
This allows you to override it in the provisioner section of the configfile.  In the chef_base it is then set to default to 'auto'

The next stage to this would be to have a commandline switch to pass data to the provisioner config.

### DIFF
--- a/lib/kitchen/provisioner/chef_base.rb
+++ b/lib/kitchen/provisioner/chef_base.rb
@@ -51,6 +51,7 @@ module Kitchen
       default_config :attributes, {}
       default_config :config_path, nil
       default_config :log_file, nil
+      default_config :log_level, 'auto'
       default_config :profile_ruby, false
       default_config :cookbook_files_glob, %w[
         README.* metadata.{json,rb}
--- a/lib/kitchen/provisioner/chef_base.rb
+++ b/lib/kitchen/provisioner/chef_base.rb
@@ -51,7 +51,7 @@ module Kitchen
       default_config :attributes, {}
       default_config :config_path, nil
       default_config :log_file, nil
-      default_config :log_level, 'auto'
+      default_config :log_level, "auto"
       default_config :profile_ruby, false
       default_config :cookbook_files_glob, %w[
         README.* metadata.{json,rb}
--- a/lib/kitchen/data_munger.rb
+++ b/lib/kitchen/data_munger.rb
@@ -82,7 +82,6 @@ module Kitchen
       merged_data_for(:provisioner, suite, platform).tap do |pdata|
         set_kitchen_config_at!(pdata, :kitchen_root)
         set_kitchen_config_at!(pdata, :test_base_path)
-        set_kitchen_config_at!(pdata, :log_level)
         combine_arrays!(pdata, :run_list, :platform, :suite)
       end
     end
--- a/lib/kitchen/provisioner/chef_apply.rb
+++ b/lib/kitchen/provisioner/chef_apply.rb
@@ -99,8 +99,7 @@ module Kitchen
 
       # (see ChefSolo#run_command)
       def run_command
-        level = config[:log_level] == :info ? :auto : config[:log_level]
-
+        level = config[:log_level]
         lines = []
         config[:run_list].map do |recipe|
           cmd = sudo(config[:chef_apply_path]).dup.
--- a/lib/kitchen/provisioner/chef_solo.rb
+++ b/lib/kitchen/provisioner/chef_solo.rb
@@ -47,8 +47,7 @@ module Kitchen
 
       # (see Base#run_command)
       def run_command # rubocop:disable Metrics/AbcSize
-        level = config[:log_level] == :info ? :auto : config[:log_level]
-
+        level = config[:log_level]
         cmd = sudo(config[:chef_solo_path]).dup.
           tap { |str| str.insert(0, "& ") if powershell_shell? }
         args = [
--- a/lib/kitchen/provisioner/chef_zero.rb
+++ b/lib/kitchen/provisioner/chef_zero.rb
@@ -118,7 +118,7 @@ module Kitchen
       # @return [Array<String>] an array of command line arguments
       # @api private
       def chef_client_args
-        level = config[:log_level] == :info ? :warn : config[:log_level]
+        level = config[:log_level]
         args = [
           "--config #{remote_path_join(config[:root_path], "client.rb")}",
           "--log_level #{level}",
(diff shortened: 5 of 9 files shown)


alexpop:
Looks like a great first step. There are a few failing tests to update

lib/kitchen/provisioner/chef_base.rb
```diff
@@ -49,6 +49,7 @@ class ChefBase < Base
       default_config :attributes, {}
       default_config :config_path, nil
       default_config :log_file, nil
+      default_config :log_level, 'auto'
```
alexpop: Is it working with log level as a string? I see the previous defaults were symbols(`:info`, `:auto`)

drrk: It's passed to the command line, so will get converted to a string at that point anyway, but I can make it a symbol for consistancy

drrk:
Aware I need to updated the tests - will do that once I add the command line options as well, just wanted feedback on this idea :)

lamont-granquist:
@drrk this looks correct, please continue

mfischer-zd:
A command-line argument to control this is not needed, provided the value can be set from the YAML file.  

If varying behavior is needed, the user can a specific value via an environment variable; since the YAML file is run through ERB evaluation, this will be trivial to implement on the user's part.

vinyar:
as @mfischer-zd mention, can this option be specified via .kitchen.yml ?

mwrock:
:+1:

smurawski:
:+1:

coderanger:
So this got merged and released in 1.7 and no one followed up on the command line changes. How can we make sure we don't release partial features in the future? As it stands, the only way to set the inner chef-client to debug logging appears to be a marginally undocumented config section.

cheeseplus:
I don't consider it a partial feature, in that we addressed a longstanding behavioural bug, but you are correct that we do need to get it documented. We should also open up another PR for the CLI changes.

coderanger:
I guess I didn't think of it as a bug given that it was the explicit way to do test debugging :) It's cool to split `--chef-log-level` to be its own thang, but I feel like without that this was strictly a step backwards for users who care more about easily seeing chef debug than not seeing TK debug.

mfischer-zd:
Since all other aspects of Chef configuration (which version to use, etc) are all encoded in the .kitchen.yml file, I don't think it's a problem not to offer a command-line argument, as long as any new or modified parameters in the configuration file are properly documented.

coderanger:
@mfischer-zd The log level is different from most other things in terms of workflow. You wouldn't want to leave it in the config permanently as `debug`, more likely it would be a one-off for a few runs to debug a specific problem. The config file should describe the tests, log_level is a property of the specific run.

mfischer-zd:
Sure, that's what env vars are for.  

```
log_level: <%= ENV['CHEF_LOG_LEVEL'] || 'info' %>
```

coderanger:
That's super gross both to expect everyone to both figure that out and include it in their kitchen config, and then to use different mechanisms for things like `--concurrent` vs `$CHEF_LOG_LEVEL`.

drrk:
And I have created a pull request on the docs to explain that too, but it has not been merged yet: https://github.com/test-kitchen/kitchen-docs/pull/44

mfischer-zd:
> That's super gross

Let's stick to quantifiable criticisms, please.

lamont-granquist:
I agree with @coderanger.

I'm very :-1: on environment variables, and I would expect 'kitchen test 1404 -l debug' to give me debug output on the chef client run.

cheeseplus:
So let me back this up one second:

We (test-kitchen core) are definitely onboard with adding a CLI option here. I don't see any reason why we can't support ENV variables, command line options, and kitchen_yml configuration.

mfischer-zd:
@lamont-granquist I would like to give users the flexibility to specify separate log levels: one for test-kitchen itself, and one for the provisioner.  (We'd benefit from this distinction.)  So I think another command-line option would be fine (e.g. `--provisioner-log-level`), as long as we don't change the behavior of the existing `--log-level` option.

lamont-granquist:
i'd kind of like an option that was a bit less of a mouthful than that, but i can see the different use cases, yeah.  if i'm debugging something that TK is doing talking to vagrant i don't want to have to slog through ohai debug output.

the only problem i see is that when someone thinks 'how do i get debug output out of my chef-client in this TK run?' that the flag they will naturally try is going to be '-l debug' because we've trained them to do that.  for the user coming to TK looking out to get CCR debug output having the flag that they're used to typing do something else is going to surprise them.

mfischer-zd:
Today, if a user specifies `--log-level` they won't get what they want.  So adding a new option leaves them no worse off than they are today.  

Sure, you could reverse the meaning of the different options (where `--log-level` applies to the provisioner, and `--kitchen-log-level` applies to TK itself), but I fear that sort of change could be more disruptive.  I don't have any data that would be helpful in making such a decision, though.

cheeseplus:
I'm all for maintaining default behaviour so changing `--log-level` gives me the heebjeebies. I also am of the mind that we try to keep TK as agnostic as possible when presented a choice like this so even though it'll suck for the Chef faithful we'll just need to make sure to have some blink tags or something in the docs.

drrk:
I believe that Thor allows short options to be more than a single character, so -pl to reduce the verbosity of the flag?

lamont-granquist:
You could do it additively instead of reversing the meaning.  So '-l debug' would give both debug outputs.  Then you either introduce selective flags, or introduce flags to suppress one or the other debug outputs.  Which makes sense to me because if you're just telling kitchen to give you debug output without telling it what debug output you want, you should probably get back all the debug output.  Then that can become annoying, so there should be ways to tweak it.

mfischer-zd:
As a user, I think even more flags to override the default would make it more complicated than necessary to achieve my aim.  But I could live with it.  I'd prefer the addition of a new option, though.

drrk:
That's a fine idea for debug, but for other log levels we get back to the original issue of them not matching up in a sensible way between kitchen and the provisioner.  I suppose if it's not set at all it could use the current behavior, and then if set carry to both if there isn't a separate override in the provisioner, but I feel like this will get extremely complex very quickly

vinyar:
I don't know if this thread is still active, I just gotta say that below was not easy to find. Thanks drrk for adding it to the docs!!

provisioner:
  log_level: debug

jvogt:
@drrk is this documented anywhere?

vinyar:
@jvogt Found it in the TK docs - https://github.com/test-kitchen/kitchen-docs/pull/51/files

rarsan:
Is there a plan to doc this in https://docs.chef.io/kitchen.html#kitchen-yml ?

PR #956, opened 2016-02-28
make rubocop glücklich

chris-rock:


mwrock:
:+1:  yes! glücklich !!

cheeseplus:
+1

chris-rock:
This fixes style issues introduced in https://github.com/test-kitchen/test-kitchen/commit/31b34d40f003c20b5bfbfd38b4828ab3c210fc13#diff-52c976fc38ed2b4e3b1192f8a8e24cffR69

chris-rock:
cc @jkeiser

adamleff:
This changes "Exclude From Changelog" and "Upstream Bug" being 2 excluded labels to 5 labels excluded named "Exclude", "From", "Changelog", "Upstream" and "Bug", which I don't think is intended. 

We probably cannot use the %w decorator here and should just use `[]`.

chris-rock:
@adamleff you're right, great catch. Instead of changing the %w decorator I filtered all label names that are not available in tk

adamleff:
:+1:

thommay:
:+1:

PR #959, opened 2016-03-02
decouple verifier class loading from instantiation

chris-rock:
The idea behind this PR is to load the class before test-kitchen tries to load the verifier data. That allows the verifier to extend test-kitchen, if needed.

This implementation allows test-kitchen to be backwards compatible and 'kitchen-inspec' to add required features, since verifiers have not been fully treated as a first-class citizen in the past and have not access to all configuration parameters.

Would love to hear your feedback and other ideas

### DIFF
--- a/lib/kitchen/config.rb
+++ b/lib/kitchen/config.rb
@@ -315,7 +315,11 @@ module Kitchen
     # @return [Verifier] a new Verifier object
     # @api private
     def new_verifier(suite, platform)
+      # load verifier class
+      Verifier.load_plugin(data.verifier_name)
+      # fetch verifier configuration
       vdata = data.verifier_data_for(suite.name, platform.name)
+      # create verifier instance
       Verifier.for_plugin(vdata[:name], vdata)
     end
   end
--- a/lib/kitchen/data_munger.rb
+++ b/lib/kitchen/data_munger.rb
@@ -124,6 +124,12 @@ module Kitchen
       end
     end
 
+    def verifier_name
+      vdata = data.fetch(:verifier, {})
+      vdata = { :name => vdata } if vdata.is_a?(String)
+      vdata[:name]
+    end
+
     private
 
     # @return [Hash] the user data hash
--- a/lib/kitchen/verifier.rb
+++ b/lib/kitchen/verifier.rb
@@ -31,25 +31,28 @@ module Kitchen
     # Default verifier to use
     DEFAULT_PLUGIN = "busser".freeze
 
+    def self.load_plugin(verifier_name)
+      require("kitchen/verifier/#{verifier_name}")
+    rescue LoadError, NameError
+      raise ClientError,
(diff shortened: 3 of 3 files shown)


mwrock:
Seems reasonable. Could you describe a bit ore about how you see verifiers "extending test-kitchen?"

My only concern is that we dont encourage a "monkey patch" architecture.

adamleff:
This seems like a reasonable approach to me and follows suit with how drivers and transports are loaded. But I have the same concern as @mwrock and would love to hear more about future plans. :)

chris-rock:
@mwrock I agree with you. The alternative that I have in mind is, that we share the global options to all plugins (driver, provisioner, verifier) as well. This would allow them to use all parameters.

Another alternative is the focus on sharing suite data with verifiers as well. And we may introduce a hook for verifier or provisioner, to add method to fetch specific data. This would also mean, that run_list is never defined in core test-kitchen but only in the chef provisioner.

adamleff:
@chris-rock I have a similar need where a driver needs to get the username from the transport in order to do reset the windows password on Google Compute Engine.

I'm against sharing the entire "config" object between each TK module because I think it will blur the lines between the responsibilities of each module. But if there's a specific piece of data that should be shared, I think it's reasonable to add a public method to expose that to the instance. In my case, I plan on sending up a PR to add a public `#username` method so that can be exposed.

What is the specific use-case you have in mind?

chris-rock:
The use case I have in mind:
```
suites:
  - name: profile
    run_list:
      - recipe[os_prepare]
    inspec_tests:
      - https://github.com/nathenharvey/tmp_compliance_profile
```
This already works, but not for the first suite, since kitchen-inspec is not loaded to adapt the data-mungler. I need `inspect_tests` from the suite. 

The stuff for the chef-provisioner is built-in: https://github.com/test-kitchen/test-kitchen/blob/master/lib/kitchen/data_munger.rb#L86, but this should be moved out. We should find a way to offer custom elements to plugins

smurawski:
@chris-rock you should have access to the instance object inside the verifier and through their everything in the config for the suite.. not everything is exposed via public properties/methods, so I've had to call diagnose on occasion to get "private" data.  Had to do this to get password out of the state object before.

chris-rock:
@smurawski I tried to get this via instance, I get the provisioner and the verifier information, but the suite config is deleted, because it only has the suite instance.

smurawski:
@chris-rock 

with the use case above, why not add inspec_tests as a conf element to the verifier and do:
```
suites:
  - name: profile
    run_list:
      - recipe[os_prepare]
    verifier:
      inspec_tests:
        - https://github.com/nathenharvey/tmp_compliance_profile
```

That should work today with any recent test-kitchen.

chris-rock:
@smurawski sometimes the solution is just too obvious :-) Doesn't look as good, but works as expected. I like that.

PR #961, opened 2016-03-02
Adding a username method to the SSH and WinRM transport classes

adamleff:
In some cases, an instance's driver might need knowledge of the
transport username in order to perform a task, such as resetting
the password for the user on a GCE Windows instance.

This change exposes the username via a public method to avoid
such hacks as accessing the `#config` private method.

### DIFF
--- a/lib/kitchen/transport/ssh.rb
+++ b/lib/kitchen/transport/ssh.rb
@@ -75,6 +75,15 @@ module Kitchen
         self
       end
 
+      # Returns the configured username. This is occassionally needed
+      # by an instance's driver to perform other tasks, such as a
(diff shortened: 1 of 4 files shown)


mwrock:
:+1:

smurawski:
```
@instance.transport[:username]
```
Doesn't work for you?  It's a publicly exposed key hanging off a property of instance...

adamleff:
Y'know, I really hate when you're right, @smurawski. :)

I completely missed this in `Kitchen::Configurable`:

https://github.com/test-kitchen/test-kitchen/blob/master/lib/kitchen/configurable.rb#L65

This PR is not necessary.

PR #971, opened 2016-03-14
--format #{config[:format]} in chef-client command

apowers:
The standard doc format has way too much data for a poor human to interpret, especially when running against multiple operating systems in parallel. Allow poor humans to specify the minimal formatter for chef-client.

### DIFF
--- a/lib/kitchen/provisioner/chef_zero.rb
+++ b/lib/kitchen/provisioner/chef_zero.rb
@@ -90,6 +90,8 @@ module Kitchen
       # @param args [Array<String>] array of flags
       # @api private
       # rubocop:disable Metrics/CyclomaticComplexity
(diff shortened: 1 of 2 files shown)


apowers:
It looks like test-kitchen can't pass tests because test-kitchen has broken tests?
Did your AWS credentials expire?
````
>>>>>> ------Exception-------
>>>>>> Message: Failed to complete #create action: [unable to sign request without credentials set]
>>>>>> ----------------------
````

lamont-granquist:
I think it'd be preferable to have a new option which allowed passing entirely arbitrary flags to the chef-client command because otherwise PRs of this nature never end.

coderanger:
You can already set arbitrary config file options, so I bet you can do this via that.

PR #973, opened 2016-03-22
Shell verifier should ensure env vars are strings

jsok:
Encountered while using the shell verified along with `kitchen-docker` which sets `KITCHEN_PORT=22` in the verifier state.
However mixlib-shellout throws an exception:

    TypeError: no implicit conversion of Fixnum into String

as it expects the environment to be all strings: https://github.com/chef/mixlib-shellout/blob/master/lib/mixlib/shellout/unix.rb#L174

Therefore, this PR sanitises the env vars which the shell verifier passes to shellout.


mwrock:
:+1:

cheeseplus:
+1

smurawski:
:+1:

PR #976, opened 2016-03-26
Use latest mixlib-install 1.0.2

mwrock:
~~This is currently blocked on a release of current mixlib-install master.~~

Prinary motivation for this is to use released TK to run acceptance tests in chef and chef-dk to help validate that the released version of test-kitchen works.

cc @sersut

### DIFF
--- a/lib/kitchen/provisioner/chef_base.rb
+++ b/lib/kitchen/provisioner/chef_base.rb
@@ -26,6 +26,8 @@ require "kitchen/provisioner/chef/common_sandbox"
 require "kitchen/provisioner/chef/librarian"
 require "kitchen/util"
 require "mixlib/install"
+require "mixlib/install/script_generator"
+
 begin
   require "chef-config/config"
   require "chef-config/workstation_config_loader"
@@ -132,18 +134,8 @@ module Kitchen
 
       # (see Base#install_command)
       def install_command
-        return unless config[:require_chef_omnibus]
-
-        version = config[:require_chef_omnibus].to_s.downcase
-
-        # Passing "true" to mixlib-install currently breaks the windows metadata_url
-        # TODO: remove this line once https://github.com/chef/mixlib-install/pull/22
-        # is accepted and released
-        version = "" if version == "true" && powershell_shell?
-
-        installer = Mixlib::Install.new(version, powershell_shell?, install_options)
-        config[:chef_omnibus_root] = installer.root
-        prefix_command(sudo(installer.install_command))
+        return unless config[:require_chef_omnibus] || config[:product_name]
+        prefix_command(sudo(install_script_contents))
       end
 
       private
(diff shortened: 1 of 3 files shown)


sersut:
Thanks for taking a stab at this @mwrock. To share my context about this change first:

When not using mixlib-install, test-kitchen creates a script that downloads `install.sh` from omnitruck and runs it. With mixlib-install however test-kitchen now has the contents of the scripts to run. In my previous work I found that we need to first render the contents of the script on the remote machine and set the `install_command` to run the rendered script in order to support all the current parameters of test-kitchen including `sudo`. 

That was why in https://github.com/sersut/test-kitchen/pull/1 there is some dance in `lib/kitchen/provisioner/base.rb` to run the `install_command` after `init_command`. And because of that dance, there are concerns about backwards compatibility which led to that branch become a long living branch. 

So I am curious about the tests you have done with this change. I tried to run acceptance tests of mixlib-install [here](https://github.com/chef/mixlib-install/tree/master/acceptance) but ran into some issues:

```
 >>>>>> Message: Could not load or activate winrm-fs. (winrm-fs is not part of the bundle. Add it to Gemfile.)
```

Probably not related. I will continue to get those tests to run.

mwrock:
@sersut tk uses winrm-fs instead of the deprecated winrm-transport so you will just need to tweak your Gemfile accordingly.

I did notice the code you referred to in the provisioner base in your branch. I think given the changes over the last couple releases of tk, leaving that code alone should work. Test-kitchen since 1.5 grabs the script from mixlib-install and renders it to a file on the test instance.

sersut:
Thanks for the info @mwrock. Fixing that I am now seeing the error here:

https://gist.github.com/sersut/113e36b2ec14440546f4

with 

https://github.com/chef/mixlib-install/blob/master/acceptance/current/.kitchen.yml

I think that is the core reason why we needed to do the dance in sersut/test-kitchen#1. 

The main difference is before mixlib-install, the command we run in the test instance downloads the install.sh and runs it like `sudo install.sh`. After mixlib-install the command we run in the test instance **is** the contents of install.sh. So running it with prepending `sudo -E` does not work.

mwrock:
ah ok @sersut and I was testing with a windows test instance and therefore not running into this. I see it now on a linux instance. Will getr that working.

mwrock:
so this is a bit hacky @sersut but its the simplest option I can think of without potentially breaking other TK provisioners.

mwrock:
Actually I think this is better now. It writes to a file on non windows and then executes via sudo.

mwrock:
Updated PR to reflect the just released mixlib-install 1.0.0 and removed the WIP marker. @sersut does the approach here for linux look fine to you now? Where we `cat` it to a file and run that as `sudo`?

schisamo:
We should wait for the impending mixlib-install 1.0.1.

mwrock:
I'll update the pr once that is released but still want to at least queue +1s. so this is ready to merge.

mwrock:
this is updated to use at least mixlib-install 1.0.1 and I also noticed chef_base was not forwarding along `install_msi_url` making that impossible to override so I added that.

mwrock:
note the appveyor break will be resolved with https://github.com/test-kitchen/test-kitchen/pull/981

mwrock:
rebased with master to address appveyor and now aligning with mixlib-1.0.2

patrick-wright:
:shipit:

cheeseplus:
:shipit:

PR #981, opened 2016-03-30
Retrofit to use fixed winrm-fs release

mwrock:
winrm-fs 0.4.0 is released. This fixes a bug with uploading encrypted data bags on windows.

PR #984, opened 2016-03-31
fix winrm transport run_from_file_command

rdoorn:
the script file was written to the directory name, instead of the scrip file within the directory, which it executes after the uploading. this fixes this by writing to the correct target_path before executing this same path.


mwrock:
Thanks for this!! But this will be covered in https://github.com/test-kitchen/test-kitchen/pull/981

PR #989, opened 2016-04-01
release 1.7.0

mwrock:
Changed the `Improvement` label to be a "Implemented Enhancement" instead of "Fixed Bug" so thats why there is change to the changelog further down. I also removed a rake task that I added out of ignorance over a year ago and I am likely the only one who ever used it.


adamleff:
:+1:

PR #1000, opened 2016-04-07
Don't require dev dependencies to build

jkeiser:
if I run non-bundled `rake install` I’d like it to work—it makes my job easier when building (that way I don’t have to `bundle install` each dir before rake installing, which could end up with us taking extra unnecessary deps into the build)

`rake` should act as normal when running with a full bundle, but if you `—without` some stuff, the tasks will get disabled instead of the rakefile failing to compile

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,8 +3,7 @@
 .bundle
 .config
 .yardoc
-bin
-!bin/kitchen
+binstubs
 Gemfile*.lock
 InstalledFiles
 _yardoc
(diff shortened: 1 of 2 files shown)


mwrock:
Beyond seeing this turn green, my only concern is around wrapping finstyle. We'd never want a default rake to pass without running the `style` task.

jkeiser:
@mwrockx a default `rake` would fail to run if that happened, since the `style` task wouldn't be there.

jkeiser:
(And I can verify that does indeed happen :))

mwrock:
Ahh right makes sense. I was afraid it would just puts a message but looks good. I'm :+1: on green

tyler-ball:
:+1:

PR #1002, opened 2016-04-08
Relax cucumber and aruba restrictions

jkeiser:
chef-dk is now installing a more recent version of cucumber-core, and it appears that test-kitchen will work fine with that too, so we may as well upgrade.


mwrock:
:+1: on green

drrk:
:+1:

PR #1003, opened 2016-04-08
Fixes busser install for older omnibus windows installs

mwrock:
This fixes https://github.com/test-kitchen/busser/issues/25

Older (chef 11) omnibus installs on Windows do not propogate the the changes to the path so commands made in the same kitchen run will not have the chef embedded bin on the path. This is exacerbated by older versions of rubygems like the one that ships with chef 11 where the binstubs do not include an absolute path to the ruby interpreter. As a result he busser installer cannot find ruby.exe.

This was fixed for the provisioners in [this commit](https://github.com/test-kitchen/test-kitchen/commit/6443145951340887e0ccd4952654b42e46022d8e), but alas, the busser still breaks.

This pr centralizes the `reload_ps1_path` method and uses it in the busser installer.

### DIFF
--- a/lib/kitchen/configurable.rb
+++ b/lib/kitchen/configurable.rb
@@ -248,6 +248,17 @@ module Kitchen
       instance ? instance.logger : Kitchen.logger
     end
 
+    # @return [String] a powershell command to reload the `PATH` environment
+    #   variable, only to be used to support old Omnibus Chef packages that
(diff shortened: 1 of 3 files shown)


mwrock:
I have tested this with chef 11.18.6 and validated that the verifier runs without error.

cheeseplus:
+1

smurawski:
:+1:

PR #1012, opened 2016-04-16
Provide the option to run all winrm commands through a scheduled task

mwrock:
This PR runs all winrm calls to the transport's `execute` method through a scheduled task that runs under a local account context of the configured transport user. The `:elevated` setting of the winrm transport must be set to true
```
transport:
  name: winrm
  elevated: true
```
and the winrm-elevated gem must be available for this to happen.

### DIFF
--- a/.kitchen.ci.yml
+++ b/.kitchen.ci.yml
@@ -13,6 +13,9 @@ provisioner:
 platforms:
   - name: ubuntu-14.04
   - name: windows-2012R2
+    transport:
+      name: winrm
+      elevated: true
 
 verifier:
   name: inspec
--- a/lib/kitchen/transport/winrm.rb
+++ b/lib/kitchen/transport/winrm.rb
@@ -42,6 +42,7 @@ module Kitchen
 
       default_config :username, "administrator"
       default_config :password, nil
+      default_config :elevated, false
       default_config :rdp_port, 3389
       default_config :connection_retries, 5
       default_config :connection_retry_sleep, 1
@@ -185,6 +186,10 @@ module Kitchen
         # @api private
         attr_reader :winrm_transport
 
+        # @return [Boolean] whether to use winrm-elevated for running commands
+        # @api private
+        attr_reader :elevated
+
         # Writes an RDP document to the local file system.
         #
         # @param opts [Hash] file options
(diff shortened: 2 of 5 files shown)


smurawski:
@mwrock will running the commands as system be an option as well? (to mimic user context when chef is run as a service or scheduled task under the system account?)

mwrock:
Good point. @smurawski I could add optional `elevated_username` and `elevated_password` settings to override the transport creds pretty easily. I'll add that before removing the WIP.

mwrock:
ok @smurawski  this can now run with a service account. Ended up being trickier than anticipated (as per usual). Had to make more mods to winrm-elevated and since file copies are done using the "normal" user, we munge `$env:temp` of the service account to the temp directory of the login user's account.

I'm not 100% confident there are no side effects here. However given the privileges of the task, it should have full rights to that directory.

smurawski:
@mwrock I'll give it a run through this afternoon!

mwrock:
winrm-elevated 0.4.0 has been released and I have squashed and removed the WIP here. This is ready for review.

jkerry:
Just tested this on a stock windows 2k8r2 box with a powershell 4 upgrade and it completed without any issues or special installation hoops to jump through like chef-zero-scheduled-task.  Inspec requiring powershell 3 or greater makes this functionality nearly necessary. Merge it yesterday please.

mwrock:
In case others want to test this but are unsure how, here's a guide:
1.Edit your Gemfile to include these entries:
```
  gem 'test-kitchen', github: 'test-kitchen', branch: 'elevated'
  gem 'winrm-fs', '~> 0.4'
  gem 'winrm-elevated', '~> 0.4'
```

2.Create a recipe one would expect to fail via vanilla winrm:
```
node.default['windows']['feature_provider'] = 'dism'

windows_feature 'NetFx3ServerFeatures' do
  action :install
end
windows_feature 'NetFx3' do
  action :install
end
```

3.Tell your `.kitchen.yml` to use an elevated transport:
```
transport:
  name: winrm
  elevated: true
```

4.Converge the recipie on a stock 2012 R2 box

This will install the .net 3.5 feature which requires a trip to WindowsUpdate that fails under normal conditions with something like:
```
       STDOUT: Deployment Image Servicing and Management tool
       Version: 6.3.9600.17031

       Image Version: 6.3.9600.17031

       Enabling feature(s)

       Error: 0x800f0906

       The source files could not be downloaded.
       Use the "source" option to specify the location of the files that are required to restore the feature. For more information on specifying a source location, see http://go.microsoft.com/fwlink/?LinkId=243077.

       The DISM log file can be found at C:\Windows\Logs\DISM\dism.log
```

smurawski:
👍 Sorry for the long turnaround :shipit:

Annih:
Thanks a lot @mwrock  for this feature!

smudgerdan:
Just updated to 0.14.13 of Chef DK and tried this out and it works as expected, very sexy :+1:

vinyar:
Sorry for the silly question if the answer is obvious. It took me some digging in the changelog to get here. Is this awesome new feature documented anywhere?

smurawski:
@vinyar for the moment just about 3 comments above yours - https://github.com/test-kitchen/test-kitchen/pull/1012#issuecomment-216685606

mwrock:
@vinyar I just submitted https://github.com/chef/chef-web-docs/pull/84 to get this in chef docs.

PR #1014, opened 2016-04-20
Add native policyfile resolution support

danielsdeleo:
Auto detects policyfiles and uses them for resolution. It's a bit wonky in that it only works with the `chef_zero` provisioner, but `chef_solo` seems to still be the default, so you have to at least add `provisioner: chef_zero` to your kitchen YAML. I tested with a kitchen YAML like:

```yaml
provisioner: chef_zero

platforms:
  - name: ubuntu-14.04

suites:
  - name: default
    attributes:
      rando_attr: hello from rando
    named_run_list: foo 
```

I confirmed the named run list functionality works, and also attributes work as expected, so it will fix https://github.com/chef/chef-dk/issues/460

### DIFF
--- a/lib/kitchen/provisioner/chef/common_sandbox.rb
+++ b/lib/kitchen/provisioner/chef/common_sandbox.rb
@@ -302,6 +302,11 @@ module Kitchen
             warn("Ignored run_list: #{config[:run_list].inspect}")
           end
           policylock = policyfile.gsub(/\.rb\Z/, ".lock.json")
+          unless File.exist?(policylock)
+            Kitchen.mutex.synchronize do
+              Chef::Policyfile.new(policyfile, sandbox_path, logger).compile
+            end
+          end
           policy_name = JSON.parse(IO.read(policylock))["name"]
           policy_group = "local"
           config[:attributes].merge(:policy_name => policy_name, :policy_group => policy_group)
--- a/lib/kitchen/provisioner/chef/common_sandbox.rb
+++ b/lib/kitchen/provisioner/chef/common_sandbox.rb
@@ -16,6 +16,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+require "json"
+
 module Kitchen
 
   module Provisioner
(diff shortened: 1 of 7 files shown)


lib/kitchen/provisioner/chef/common_sandbox.rb
```diff
@@ -86,6 +88,14 @@ def all_files_in_cookbooks
             select { |fn| File.file?(fn) && ! %w[. ..].include?(fn) }
         end
 
+        # @return [String] an absolute path to a Policyfile, relative to the
+        #   kitchen root
+        # @api private
+        def policyfile
```
mwrock: This is also in chef_base. can we consolidate?

lib/kitchen/provisioner/chef/common_sandbox.rb
```diff
@@ -278,6 +295,27 @@ def prepare_json
...
         end
 
+        def update_dna_for_policyfile
+          if !config[:run_list].nil? && !config[:run_list].empty?
+            warn("You must set your run_list in your policyfile instead of "\
+                 "kitchen config. The run_list your config will be ignored.")
+            warn("Ignored run_list: #{config[:run_list].inspect}")
+          end
+          policylock = policyfile.gsub(/\.rb\Z/, ".lock.json")
+          policy_name = JSON.parse(IO.read(policylock))["name"]
```
mwrock: is it possible the lock file may not be generated here? Should we provide a better error if so?

mwrock:
While we are at it, why not move https://github.com/chef/chef-dk/blob/master/lib/kitchen/provisioner/policyfile_zero.rb here?

lib/kitchen/provisioner/chef/common_sandbox.rb
danielsdeleo: This logic is also duplicated for the berks and librarian resolvers. If you want I could refactor all of them, but I didn't want to have one pre-refactored and the others not-refactored.

lib/kitchen/provisioner/chef/common_sandbox.rb
danielsdeleo: Good catch. I thought that `chef export` would get called first, but actually it's not and indeed that case will fail with `ENOENT`.

danielsdeleo:
A note about the `policyfile_zero` provisioner: I'd like this to replace that and have that one start emitting deprecation warnings as soon as this patch makes its way into ChefDK. 95% of the policyfile functionality as it relates to kitchen is that of a resolver, and at the moment those are not pluggable, so it's a lot better to have it built-in (auto detection and so on). This will also be easier to use with `named_run_list` since the data munger is patched to make it work the way users are used to. I guess tl;dr, with this patch there's no reason to use the `policyfile_zero` provisioner any more and we might as well only maintain one thing that does this.

lib/kitchen/provisioner/chef/common_sandbox.rb
mwrock: oh ok. thats fine. we can refactor separately.

mwrock:
Oh got it @danielsdeleo yeah this is WAY better than the provisioner. I'm :+1: here pending we handle the `ENOENT` scenario above.

danielsdeleo:
@mwrock done

adamleff:
This is sweet. 👍

PR #1023, opened 2016-05-10
uploads: reuse connections+disable compression

lamont-granquist:
knocks 25% of the time off of uploads or so, not a magic bullet, but
probably worth it.

### DIFF
--- a/lib/kitchen/driver/ssh_base.rb
+++ b/lib/kitchen/driver/ssh_base.rb
@@ -292,12 +292,13 @@ module Kitchen
         return if locals.nil? || Array(locals).empty?
 
         info("Transferring files to #{instance.to_str}")
-        debug("TIMING: scp uploading files asynchronously (Kitchen::Driver::SSHBase)")
+        debug("TIMING: scp asynch upload (Kitchen::Driver::SSHBase)")
         elapsed = Benchmark.measure do
           waits = locals.map { |local| connection.upload_path(local, remote) }
           waits.each(&:wait)
         end
-        debug("TIMING: scp uploading files asynchronously (Kitchen::Driver::SSHBase) took #{Util.duration(elapsed.real)}")
+        delta = Util.duration(elapsed.real)
+        debug("TIMING: scp async upload (Kitchen::Driver::SSHBase) took #{delta}")
         debug("Transfer complete")
       rescue SSHFailed, Net::SSH::Exception => ex
         raise ActionFailed, ex.message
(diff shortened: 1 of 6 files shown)


lamont-granquist:
TIMING: scp uploading files synchronously (Kitchen::Transport::Ssh) took (5m44.50s) seconds

vs

TIMING: scp uploading files asynchronously (Kitchen::Transport::Ssh) took (4m35.72s) seconds

most of that is just twiddling the compression default.

lamont-granquist:
I strongly suspect the "asynchronous" net-ssh commands are not so asynchronous...

mwrock:
looks like there are a couple lines that need shortening

mwrock:
So you think the perf gain is more due to turning off compression that enabling async? Was there any significant difference without the compression change?

I imagine most use cases here deal with a moderate number of smallish files where compression will not play a significant role, but also weird that their would be a significant cpu/perf penalty incurred by compression on a modern core.

lamont-granquist:
The difference in the timing due to "async" was 0-10 seconds and within the jitter I was seeing between tests.  The 60 second difference was the only significant difference.  My intuition is that the whole process is bottlenecked on CPU because of encryption and hmac and ruby, and compression is purely additive to that CPU load so it slows everything down -- which also explains why async seemed to not do much, since we're running on MRI and the most we can ever use is a single thread.

We should consider replacing finstyle with chefstyle, but there's still too many unenabled cops in chefstyle...

lamont-granquist:
i also tested net-ssh yesterday and it certainly looks like the way net-scp uses net-ssh should be asynchronous, which adds more evidence to the idea that we're just contending for a single core here.

i'm doing timing tests of scp on travis to determine cipher performance, and will compare to net-scp to see what can be done about cipher/hmac/kex algorithm, and will fire off another PR to allow tweaking all of that, along with the best defaults i can find.

mwrock:
:+1:

kamaradclimber:
From our tests, most of the time is indeed spent in transfering many (5.5k in our case) very small files.

We've written a small transport to solve that issue (https://github.com/criteo/kitchen-transport-speedy) and results are pretty satisfying.

It might help other users (although there are other alternatives)

lamont-granquist:
tar having to be present is a huge :-1: if you want to patch it, it should use Archive::Tar.

i'd still like to know what the hell net-scp is doing that makes it so gawdawfully slow with multiple files.  this whole thing is screaming out that there is a either a huge perf bug in net-scp or a huge perf bug in how we're using net-scp.

cheeseplus:
ditto to @lamont-granquist's feelings - we don't want to depend on tar and determining where the actual perf issue lies. It's great you created your own transport and @coderanger did lots of investigation into these issues in creating https://github.com/coderanger/kitchen-sync which are related to why we've shot down past TAR based implementations as well

PR #1027, opened 2016-05-11
Empty string for the config setting for proxies did not really work

smurawski:
/cc @cheeseplus

### DIFF
--- a/lib/kitchen/configurable.rb
+++ b/lib/kitchen/configurable.rb
@@ -306,34 +306,28 @@ module Kitchen
     # @param code [String] the shell code to be wrapped
     # @return [String] wrapped shell code
     # @api private
-    # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
-    # rubocop:disable Metrics/PerceivedComplexity, Metrics/CyclomaticComplexity
     def wrap_shell_code(code)
-      env = []
-      if config[:http_proxy] && !config[:http_proxy].empty?
-        env << shell_env_var("http_proxy", config[:http_proxy])
-        env << shell_env_var("HTTP_PROXY", config[:http_proxy])
-      elsif config.key? :http_proxy
-        # Maybe a logging line here?
+      env = proxy_settings
+      if powershell_shell?
+        env.join("\n").concat("\n").concat(code)
       else
-        export_proxy(env, "http")
+        Util.wrap_command(env.join("\n").concat("\n").concat(code))
       end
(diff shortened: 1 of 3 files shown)


jjasghar:
💥 goes the dyno-mite!

smurawski:
The problem with the PR we merged a while back is

```
!config[:http_proxy].empty?
```

since 

```
"".empty? == true
```

so, it doesn't see the empty setting for the proxy and defaults to using the environmental variables.

lib/kitchen/configurable.rb
```diff
@@ -306,28 +306,28 @@ def validate_config!
...
+        env.join("\n").concat("\n").concat(code)
       else
-        export_proxy(env, "ftp")
+        Util.wrap_command(env.join("\n").concat("\n").concat(code))
       end
+    end
+
+    # rubocop:disable Metrics/AbcSize, Metrics/PerceivedComplexity, Metrics/CyclomaticComplexity
+    def proxy_settings
+      env = [:http_proxy, :https_proxy, :ftp_proxy].collect do |protocol|
```
mwrock: you could do:
```
env = [:http_proxy, :https_proxy, :ftp_proxy].each_with_object([]) do |protocol, set_env|
        if !config.key?(protocol) || config[protocol].nil?
          export_proxy(set_env, protocol)
        elsif proxy_config_setting_present?(protocol)
          set_env << shell_env_var("#{protocol.downcase}", config[protocol])
          set_env << shell_env_var("#{protocol.upcase}", config[protocol])
        end
      end
```
and that may allow you to get rid of the rubocop suppression.

lib/kitchen/configurable.rb
```diff
@@ -306,28 +306,28 @@ def validate_config!
...
-      if config[:https_proxy] && !config[:https_proxy].empty?
-        env << shell_env_var("https_proxy", config[:https_proxy])
-        env << shell_env_var("HTTPS_PROXY", config[:https_proxy])
-      else
-        export_proxy(env, "https")
-      end
-      if config[:ftp_proxy] && !config[:ftp_proxy].empty?
-        env << shell_env_var("ftp_proxy", config[:ftp_proxy])
-        env << shell_env_var("FTP_PROXY", config[:ftp_proxy])
+      env = proxy_settings
```
mwrock: try `env = (proxy_settings << code).join(\n)`

mwrock:
Im :+1: with or without my suggestions

cheeseplus:
👍

PR #1028, opened 2016-05-12
Update `chef_omnbius_url` default value

schisamo:
The correct install.sh URL for Omnitruck is now:
https://omnitruck.chef.io/install.sh

/cc @test-kitchen/core


adamleff:
👍

cheeseplus:
+1

PR #1030, opened 2016-05-16
Use tar if available to speed up the file transfert

intelradoux:
The idea is to test if "tar" is available on target, 
if so use it to tar the local file, transfer only 1 file with compressed data , then un-tar on target.

It is something like 10000x quicker than scp 1 file at a time.

### DIFF
--- a/lib/kitchen/driver/ssh_base.rb
+++ b/lib/kitchen/driver/ssh_base.rb
@@ -21,6 +21,8 @@ require "thor/util"
 require "kitchen/lazy_hash"
 require "benchmark"
 
+require 'rubygems/package'
+
 module Kitchen
 
   module Driver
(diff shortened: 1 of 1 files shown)


snahelou:
+1

smurawski:
Hey @intelradoux.  Thank you for the PR and interest in improving Test-Kitchen.

We've had a history of pull requests around this type of functionality.  We are trying to keep Test-Kitchen's transport as streamlined and simple as possible, but this idea would be a great fit for a potential plugin.

I noticed that your changes updated the SshBase driver, which is a legacy driver for backward compatability with earlier drivers.  With current Test-Kitchen releases we've restructured how the transport is handled and now the transport options are as plugable as the drivers and provisioners (and verifiers).  A great example of a project that implements a new transport is [`kitchen-sync`](https://github.com/coderanger/kitchen-sync).

Issue #330, opened 2014-01-19
No way to disable colors


coderanger:
For things like jenkins and travis, it would be nice to disable color. This could also be done as a default, to only output color if stdout (or whatever is passed in as the logdev) is a TTY.

jujugrrr:
Hi,
Now that we have #481 . Is there a way to force to enable color even without tty ?
By example Jenkins with [ANSI plugin](https://wiki.jenkins-ci.org/display/JENKINS/AnsiColor+Plugin) can display colors.

 Thank you

petere:
Yeah that would be nice.

jujugrrr:
> Is there a way to force to enable color even without tty ?

do you think we could add a cli option?

stormerider:
Yeah, having this be able to be specified explicitly would be helpful to me as well. I'm using Bamboo as my CI tool to run this and it's still sending the color codes, so it must have a TTY the way the Bamboo agent executes the job... perhaps Bamboo operates differently than Jenkins in that regard.

gh2k:
@stormerider Just found this via a google search. I resorted to running `kitchen test all | cat`, which triggers the lack-of-tty detection. It's not pretty (--no-color would be much more preferable) but it works.

jujugrrr:
> @stormerider Just found this via a google search. I resorted to running kitchen test all | cat, which triggers the lack-of-tty detection. It's not pretty (--no-color would be much more preferable) but it works.

Thanks for the workaround :+1:

PR #1032, opened 2016-05-19
Add color options

jorhett:
adds commandline flag ot enable disable color output:

`--color` - This is useful when running kitchen from environment that is not a tty.
`--no-color` - Useful when running kitchen from terminals that do not support ANSI

### Can be tested with:

should not give color output:
```
chef exec kitchen destroy  | cat
```
should not give color output:
```
chef exec kitchen destroy --no-color
```

should give color output:
```
chef exec kitchen destroy --color | cat
```

This PR fetched and rebased from @Igorshp's pr#765 as requested by @cheeseplus to solve #330 and using kitchen in Chef Delivery and Jenkins.

### DIFF
--- a/lib/kitchen/cli.rb
+++ b/lib/kitchen/cli.rb
@@ -91,6 +91,10 @@ module Kitchen
       method_option :log_overwrite,
         :desc => "Set to false to prevent log overwriting each time Test Kitchen runs",
         :type => :boolean
+      method_option :color,
+        :type => :boolean,
+        :lazy_default => $stdout.tty?,
+        :desc => "Toggle color output for STDOUT logger"
     end
 
     # Sets the test_base_path method_options
@@ -349,6 +353,7 @@ module Kitchen
       unless options[:log_overwrite].nil?
         @config.log_overwrite = options[:log_overwrite]
       end
+      @config.colorize = options[:color] unless options[:color].nil?
 
       if options[:test_base_path]
         # ensure we have an absolute path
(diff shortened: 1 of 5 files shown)


jorhett:
gah. I'll update the tests tomorrow.

Igorshp:
Wow, it's been a while.  Great to see the feature might be implemented.
Thanks for picking it up! 
👍

jorhett:
100 characters line length? That's totally crazy. And if that is the line length you want, please put it in rubocop so that it fails local testing rather than waiting until it hits your PR acceptance.

smurawski:
@jorhett I'm sorry that bit you.  Those rules are available for local testing via finstyle.  The `rake style` and `rake quality` tests cover those cases.

jorhett:
@smurawski The entire rake suite was running clean before I submitted those patches.

And 100 characters? Really? Do we improve the code with unreadable line continuations, or by replacing well-named variables with n, w, v ...?

smurawski:
@jorhett that's an entirely reasonable conversation that can be had, but that is the current state of the styling rules.

jorhett:
line length aside, any reason not to merge? Sad to have missed 1.9.0

smurawski:
👍  LGTM - Other @test-kitchen/core review?

smurawski:
@jorhett We're getting better at shipping more quickly, so once merged, it shouldn't be too long before we kick out an updated gem.

cheeseplus:
+1

PR #1033, opened 2016-05-19
Use command_prefix provided by Kitchen::Provisioner::Base in shell provisioner

pstengel:
Kitchen::Provisioner::Base provides a `:command_prefix` option for... prefixing commands. Use this in the shell provisioner so that we can do things like prefixing environment variables for use within the script.

### DIFF
--- a/lib/kitchen/provisioner/shell.rb
+++ b/lib/kitchen/provisioner/shell.rb
@@ -79,7 +79,7 @@ module Kitchen
         )
         code = powershell_shell? ? %{& "#{script}"} : sudo(script)
 
-        wrap_shell_code(prefix_command(code))
+        prefix_command(wrap_shell_code(code))
       end
 
       private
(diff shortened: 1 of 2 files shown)


mwrock:
Here `prefix_command` is provided to `wrap_shell_code` while other provisioners have the opposite order such as `chef_zero` as an example:
```
      def run_command
        cmd = modern? ? local_mode_command : shim_command

        prefix_command(
          wrap_shell_code(
            [cmd, *chef_client_args].join(" ").
            tap { |str| str.insert(0, reload_ps1_path) if windows_os? }
          )
        )
      end
```
I'd consider doing the same here for the sake of consistency unless you have a specific reason to do otherwise.

pstengel:
@mwrock 2759518

mwrock:
just need to retrofit the test now :)

pstengel:
@mwrock Whoops! 😄 6183127 should do it.

mwrock:
:+1:

smurawski:
:+1:

PR #1034, opened 2016-05-20
Buffer errors until the end of an action

smurawski:
Updates the behavior where an earlier failure would block later instances.  With this PR, all the errors for the action are buffered until the end of the action for all instances.  If there were any errors, they are reported (one in the console - the rest in the log).

### DIFF
--- a/lib/kitchen/command.rb
+++ b/lib/kitchen/command.rb
@@ -170,15 +170,26 @@ module Kitchen
         concurrency.times { queue << nil }
 
         threads = []
+        errors = []
         concurrency.times do
           threads << Thread.new do
             while instance = queue.pop
-              instance.public_send(action, *args)
-              instance.cleanup!
+              begin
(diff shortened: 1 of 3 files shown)


mwrock:
:+1: I just tested this and tweaked one of the suiles to have a syntax error. With this PR, all instances complete.

I also ran the same suites with the currently released 1.8.0 but the results were basically the same. Unfortunately the nature of the thread related bugs are tough to reproduce consistently. Based on the code changes here, I do think this will be more reliable and certainly won't make things less reliable.

smurawski:
I tested this by tweaking a provisioner to raise an unhandled error and validating that, with varying levels of concurrency specified, it would work its way through all the specified suites, rather than dying at the first unhandled error.  All the errors would be rolled up and available with stack traces in the log.

cheeseplus:
👍

PR #1038, opened 2016-05-25
Support overriding inherent cookbook resolution order

### LINKED ISSUE #1036: Test Kitchen now uses local profile and gives no option to use berks
## Description

When I run kitchen with the chef_zero provisioner it will try to read a local policyfile, even when I want to use the berksfile for kitchen. We use berksfile for kitchen to use local resources and want to test it that way.
## Kitchen Version

1.8
## ChefDK Version

0.14.25
## Platform Version

Windows 7
## Replication Case

Run test kitchen from a cookbook where there is a berksfile and policyfile. Before 1.8 it would use the berksfile. After 1.8 it uses the policyfile and doesn't allow you to switch back.
## Kitchen Output

The relevant output of the test-kitchen run or a link to a gist of the entire run, if there is one.

The debug output (kitchen -l debug) may be useful, but please link to a gist, or truncate it.

Please do not paste the full run out put here.  If the full output would be useful, link to it in a gist.

Output: https://gist.github.com/mhedgpeth/188190d5d02bd70084cb0cb643e2824a
## Kitchen Diagnose

Please include the output of `kitchen diagnose --all` or `kitchen diagnose --loader` (if the first failed).  This can be helpful in troubleshooting, so please include it in a gist.

Diagnose Gist: https://gist.github.com/mhedgpeth/7ed5a01ad744f9e4d54cf3db1d2c6f8f
## NOTE:

This repository is for core issues with the Test-Kitchen framework, so if the issue is with a plugin you may be referred to file the issue with the appropriate plugin.

itmustbejj:
Fixes #1036 .

Right now, there is a fixed order of policyfile, berksfile, cheffile, cookbooks_dir, metadata_rb that `prepare_cookbooks` will use to chose how to resolve cookbook dependencies in TK. It choses the first option that passes a `File.exists?` check. This PR provides the config[:preferred_resolver] option to only match the preferred resolver, regardless of it's order in the if/else block in `prepare_cookbooks`.

I've put a fair bit of thought into this, and I'm pretty happy with the result. I spent a good chunk of time writing a first pass at this, and then continually banged my head trying to refactor away all the complexity I introduced. I then took my dog for a walk, came back, and rewrote this completely in about 10 minutes. I think it's way more readable and non-intrusive.


mwrock:
I like it. It needs tests before merge but I think this is a good approach.

cheeseplus:
I'm putting this into accepted but still requires tests before merge.

itmustbejj:
@cheeseplus Sorry I've been pulled in a couple directions. I'll see about getting the tests committed this week.

lib/kitchen/provisioner/chef/common_sandbox.rb
```diff
@@ -92,29 +92,45 @@ def all_files_in_cookbooks
         #   kitchen root
         # @api private
         def policyfile
-          basename = config[:policyfile_path] || "Policyfile.rb"
-          File.join(config[:kitchen_root], basename)
+          if [nil, "policyfile", :policyfile].include?(config[:preferred_resolver])
```
smurawski: we have this exact same logic in chef_base.. rather than duplicate it here - why not expose a method off chef_base that common_sandbox can use?

lib/kitchen/provisioner/chef_base.rb
```diff
@@ -167,22 +173,34 @@ def install_options
       #   kitchen root
       # @api private
       def policyfile
-        policyfile_basename = config[:policyfile_path] || "Policyfile.rb"
-        File.join(config[:kitchen_root], policyfile_basename)
+        if [nil, "policyfile", :policyfile].include?(config[:preferred_resolver])
```
smurawski: Instead of checking the preferred resolver each time in `policyfile`, `berksfile`, and `cheffile`, maybe we could resolve the preferred provider earlier and leave these methods alone?

smurawski:
Rather than implementing those preferred resolver checks in several locations, I'd suggest extracting some of the loader logic added to `#load_needed_dependencies!` to a separate method, building  preferred resolver check method (that could be public and used in common_sandbox).

The challenge I hit when trying to retrofit some tests into this PR is the intense amount of mocking that needs to happen to test these changes.  If we can extract this behavior to a couple of methods, our testing surface for changes becomes much more approachable.

PR #1042, opened 2016-05-27
Fix CHANGELOG.md links

smurawski:
And get the rake task to use the right version string (`rake release` tags with "v" + version number).


mwrock:
:+1:

adamleff:
👍

Issue #1035, opened 2016-05-21
Message: SCP upload failed (open failed (1))


swimfrog:
## Description

test-kitchen (latest master branch) fails upload using chef_zero provisioner. Appears to be caused by #1023 . Works fine if I check out commit 57927ea3d34eae1a138380ff0638cc66dd397f8e, but not latest master.

## Kitchen Version

Test Kitchen version 1.8.0 (latest master, checked out using appbundle-updater)

Debug log:

```
$ kitchen converge -l debug
-----> Starting Kitchen (v1.8.0)
D      [Vagrant command] BEGIN (vagrant --version)
D      [Vagrant command] END (0m0.22s)
D      Policyfile found at /Users/user/documents/workspace/project/chef-zero/policy/PolicyFile.rb, using Policyfile to resolve dependencies
-----> Converging <node-linux-64-x64-xxx>...
       Preparing files for transfer
D      Creating local sandbox in /var/folders/g7/5xk_0ym15s1cr8pz098080ph0000gn/T/node-linux-64-x64-xxx-sandbox-20160520-86318-1fazke9
       Preparing dna.json
D      Creating dna.json from {:test=>true, :logstash=>{:local_debug=>true, :local_debug_port=>65012}, :policy_name=>"system", :policy_group=>"local"}
       Exporting cookbook dependencies from Policyfile /var/folders/g7/5xk_0ym15s1cr8pz098080ph0000gn/T/node-linux-64-x64-xxx-sandbox-20160520-86318-1fazke9...
D      [local command] BEGIN (chef export /Users/user/documents/workspace/project/chef-zero/policy/PolicyFile.rb /var/folders/g7/5xk_0ym15s1cr8pz098080ph0000gn/T/node-linux-64-x64-xxx-sandbox-20160520-86318-1fazke9 --force)
       Exported policy 'system' to /var/folders/g7/5xk_0ym15s1cr8pz098080ph0000gn/T/node-linux-64-x64-xxx-sandbox-20160520-86318-1fazke9
       
       To converge this system with the exported policy, run:
         cd /var/folders/g7/5xk_0ym15s1cr8pz098080ph0000gn/T/node-linux-64-x64-xxx-sandbox-20160520-86318-1fazke9
         chef-client -z
D      [local command] END (0m25.01s)
       Removing non-cookbook files before transfer
       Preparing data_bags
D      Using data_bags from /Users/user/documents/workspace/project/databags-vagrant/data_bags
[...]

PR #1043, opened 2016-06-01
#1035 fix scp upload failed

steenzout:
Here is a [link](https://travis-ci.org/saucelabs-ansible/packer/builds/134507319) of a failed build on travis-ci that demonstrates this problem (I managed to replicate the same problem in my laptop).

I don't know exactly why but using the code below makes it work on my laptop.


steenzout:
it seems this patch prevents async uploads therefore it may not be what you want.

cheeseplus:
Closing this in favor of #1047

PR #1049, opened 2016-06-08
Smurawski/specify cookbook resolver v2

smurawski:
Adding tests to #1038 to validate provider resolution.  cc @itmustbejj


smurawski:
There's just way to much coupling on this implementation (and between chef_base and common_sandbox).  I'm going to suggest some changes to the initial implementation.

PR #1053, opened 2016-06-09
Add fallback support for `policyfile` for compat with the older policyfile_zero

coderanger:
Because yay compat! We can revert this at some point but the code is minimal enough that it seems fine to leave for a while. Were there any other compat breaks when the provisioner code was moved?

### DIFF
--- a/lib/kitchen/provisioner/chef/common_sandbox.rb
+++ b/lib/kitchen/provisioner/chef/common_sandbox.rb
@@ -92,7 +92,7 @@ module Kitchen
         #   kitchen root
         # @api private
         def policyfile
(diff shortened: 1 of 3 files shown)


smurawski:
@coderanger I haven't heard any other compat breaks with the provisioner code move.  The biggest change overall that I've seen is #1038, where the order of precedence for Policyfiles over Berksfiles or Cheffiles so that as long as a Policyfile.rb is present there is no way to pick the other as the resolver for that kitchen run.

smurawski:
👍  on green

coderanger:
I don't think the appvayor failure is related?

mwrock:
:+1: I'll have a look at the stupid mercurial break.

PR #1055, opened 2016-06-13
Retry `Kitchen::Provisioner#run_command` after allowed exit codes

### LINKED ISSUE #1016: Base Provisioner Should Support Reboot And Continue
## Description

If a provisioner does something to a machine that requires a reboot before continuing, the provisioner should be able to notify test-kitchen that a reboot is imminent or required.  Test-Kitchen should then either be able to attempt to reconnect after a pre-determined timeout and/or initiate a reboot of the remote node and attempt to reconnect.  After reconnecting, the provisioner should optionally re-run before moving on to the verify stage.

smurawski:
- [x] Allow user to define exit codes to retry Provision#run_command
- [x] Allow user to define wait time between attempts
- [x] Allow user to define exit codes to retry 
- [x] Allow user to define maximum number of retry attempts

Resolves #1016

### DIFF
--- a/lib/kitchen/command.rb
+++ b/lib/kitchen/command.rb
@@ -170,7 +170,6 @@ module Kitchen
         concurrency.times do
           threads << Thread.new do
             while instance = queue.pop
-              puts "running #{instance.name}"
               run_action_in_thread(action, instance, *args)
             end
           end
--- a/lib/kitchen/command.rb
+++ b/lib/kitchen/command.rb
@@ -176,14 +176,19 @@ module Kitchen
           end
         end
         threads.map(&:join)
+        report_errors
+      end
+
+      # private
+
+      def report_errors
         unless @action_errors.empty?
           msg = ["#{@action_errors.length} actions failed.",
-                 @action_errors.map { |e| ">>>>>>     #{e.message}"}].join("\n")
+                 @action_errors.map { |e| ">>>>>>     #{e.message}" }].join("\n")
           raise ActionFailed.new(msg, @action_errors)
         end
       end
 
-      # private
       def concurrency_setting(instances)
         concurrency = 1
         if options[:concurrency]
--- a/lib/kitchen/command.rb
+++ b/lib/kitchen/command.rb
@@ -170,13 +170,16 @@ module Kitchen
         concurrency.times do
           threads << Thread.new do
             while instance = queue.pop
+              puts "running #{instance.name}"
               run_action_in_thread(action, instance, *args)
             end
           end
         end
         threads.map(&:join)
         unless @action_errors.empty?
-          raise ActionFailed.new("#{@action_errors.length} actions failed.", @action_errors)
+          msg = ["#{@action_errors.length} actions failed.",
+                 @action_errors.map { |e| ">>>>>>     #{e.message}"}].join("\n")
+          raise ActionFailed.new(msg, @action_errors)
         end
       end
 
@@ -192,6 +195,8 @@ module Kitchen
 
       def run_action_in_thread(action, instance, *args)
         instance.public_send(action, *args)
+      rescue Kitchen::InstanceFailure => e
+        @action_errors << e
       rescue Kitchen::ActionFailed => e
         new_error = Kitchen::ActionFailed.new("#{e.message} on #{instance.name}")
         new_error.set_backtrace(e.backtrace)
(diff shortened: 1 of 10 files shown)


smurawski:
I've tested this against Windows, Centos, and Ubuntu guests.  The Windows guests worked the most consistently.  The Centos and Ubuntu guests were more of a race to see if Chef could finish before the system shut down. (See https://github.com/chef/chef/issues/5026)

smurawski:
With this PR - all provisioners support three new configuration settings.

* `retry_on_exit_code` - which takes an array of exit codes that can indicate that kitchen should retry the converge command.  Defaults to an empty array.
* `max_retries` - number of times to retry the converge before passing along the failed status.  Defaults to 1.
* `wait_for_retry` - number of seconds to wait between converge attempts.  Defaults to 30.

smurawski:
Appveyor build failure seems specific to the build box.  More "symlink is unimplemented" errors, which seem to be common amongst appveyor builds recently.

cheeseplus:
+1

lib/kitchen/command.rb
```diff
@@ -170,17 +170,25 @@ def run_action(action, instances, *args)
         concurrency.times do
           threads << Thread.new do
             while instance = queue.pop
+              puts "running #{instance.name}"
```
adamleff: Is this debug output? If not, we should probably use the logger methods instead.

lib/kitchen/provisioner/chef_zero.rb
```diff
@@ -76,14 +76,18 @@ def run_command
 
         prefix_command(
           wrap_shell_code(
-            [cmd, *chef_client_args].join(" ").
+            [cmd, *chef_client_args, last_exit_code].join(" ").
```
adamleff: This feels a bit weird to me since we do some windows-specific stuff in this method call, but then also do some more in `#last_exit_code`.

Perhaps we should create a `#format_chef_command` method, pass `cmd` and `*chef_client_args` to it, and then it returns a string with the appropriate windows/PS-specific manipulations if needed.

lib/kitchen/transport/base.rb
```diff
@@ -97,10 +105,43 @@ def close
...
+        # Execute a command on the remote host and retry
+        #
+        # @param command [String] command string to execute
+        # @param retryable_exit_codes [Array] Array of exit codes to retry against
+        # @param max_retries [Fixnum] maximum number of retry attempts
+        # @param wait_time [Fixnum] number of seconds to wait before retrying command
+        # @raise [TransportFailed] if the command does not exit successfully,
+        #   which may vary by implementation
+        def execute_with_retry(command, retryable_exit_codes = [], max_retries, wait_time)
+          max_retries = 1 if max_retries.nil?
```
adamleff: Should we set these defaults in the `default_config` calls instead of here?

lib/kitchen/transport/base.rb
```diff
@@ -97,10 +105,43 @@ def close
...
+          wait_time = 30 if wait_time.nil?
+          tries = 0
+          begin
+            tries += 1
+            debug("Attempting to execute command - try #{tries} of #{max_retries}.")
+            execute(command)
+          rescue Kitchen::Transport::TransportFailed => e
+            if retry? tries, max_retries, retryable_exit_codes, e.exit_code
```
adamleff: Style nit: we should wrap the args in parens.

adamleff:
Some minor things, all non-blockers. 👍

lib/kitchen/command.rb
mwrock: I had the same question and then noticed `puts` getting used for these in several places in TK. a bit weird but separate from this PR

smurawski: Oops.. that should be pulled.  I was troubleshooting the difference between actionfailed and instancefailed.  Thanks.

lib/kitchen/provisioner/chef_zero.rb
smurawski: Yeah, unfortunately, to get the real exit code from PowerShell, we've got to do some gymnastics after chef-client exits.

lib/kitchen/transport/base.rb
smurawski: So, I'm open to doing that, but I also wanted to guard against unexpected nils in the .kitchen.yml heirarchy that may have been overlooked/forgotten, but probably best to surface that in the .kitchen.yml / default_config.

mwrock:
:+1:

lamont-granquist:
👍

carpnick:
Thanks for doing this @smurawski.  This one was a long time coming to TK.  Thanks for putting in the effort on chef and TK to get this done.

PR #1057, opened 2016-06-15
Fix appveyor builds

smurawski:


mwrock:
sheesh. I wonder if its `pry-stackexplorer` thats the culprit here?

mwrock:
lets just blame it all on WMF5

smurawski:
https://github.com/brigade/overcommit/issues/387  Now we are back to failing on the test-kitchen run failing to find the mercurial bits to download and install.

mwrock:
:+1: good old filezilla

smurawski:
Had to find something that was easy to install that wasn't on an appveyor box :)

PR #1058, opened 2016-06-16
Release 1.10.0

smurawski:
Prepping a release.

There are a couple of PRs outstanding #1055 and #1057 that could be merged first.  

If we merge #1055, then we'll bump the version to 1.10.0 since that's a more impactful change.


mwrock:
just adding my :+1: expecting this to be 0.10.0

PR #1067, opened 2016-06-24
bumping version for a gem repackage

mwrock:


PrajaktaPurohit:
+1

smurawski:
@mwrock 👍  Though we might want to make gitattributes for linux line endings first, so we don't get bit again.

PR #1068, opened 2016-06-27
ignore_proxy:  Option to ignore the proxy settings running chef-client

MarkGibbons:
I run chef-client in an enterprise environment where the client will not
have access to the internet.  I would like to test without proxies being
specified.  I also use omnibus to install chef on my instances.  These
two requirements lead to this work flow. No_proxy setting are just not the
same as not specifying the proxy values and in many cases no_proxy settings
 do not work.

Work flow before:
- Set the proxies in .kitchen.yml
- Kitchen converge, runs omnibus, fails running chef-client
- Remove the proxy setting.
- Kitchen converge.
- Add the proxy settings
- Kitchen verify.

Work flow with ignore_proxy set
- Kitchen test  (proxy settings are only used for omnibus chef-client install and for setup)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *.gem
 *.rbc
+*.un~
 .bundle
 .config
 .yardoc
--- a/lib/kitchen/provisioner/chef_base.rb
+++ b/lib/kitchen/provisioner/chef_base.rb
@@ -45,6 +45,7 @@ module Kitchen
       default_config :require_chef_omnibus, true
       default_config :chef_omnibus_url, "https://omnitruck.chef.io/install.sh"
       default_config :chef_omnibus_install_options, nil
+      default_config :chef_ignore_proxy, false
       default_config :run_list, []
       default_config :attributes, {}
       default_config :config_path, nil
@@ -427,7 +428,7 @@ module Kitchen
           File.dirname(__FILE__), %w{.. .. .. support }, handler_filename
         )
         FileUtils.cp(source, File.join(sandbox_path, handler_filename))
-        File.open(File.join(sandbox_path, "client_no_updated_resources.rb"), "wb") do |file|
+        File.open(File.join(sandbox_path, config_no_updates), "wb") do |file|
           file.write(format_config_file(data))
           file.write("\n\n")
           file.write("handler_file = File.join(File.dirname(__FILE__), '#{handler_filename}')\n")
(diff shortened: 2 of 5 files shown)


cheeseplus:
Have you tried specifying your own omnibus_url or providing a local package? That is typically how this issue is addressed and I'm not clear that this is entirely necessary given this facility.

MarkGibbons:
@cheeseplus ,

I've specified my own local url and done local packages.  But, I don't want to maintain my own copies of the chef-client.  "Latest" which I run pretty frequently, doesn't really work if I have to figure out what that is every day. 

I also still need to have proxies turned on for test kitchen setup, and off for converge.  I don't see that specifying the omnibus location fixes that.  Is there someway to be selective about proxy use during setup?

Thanks,
  Mark

oifland:
As I understand it, this patch allows different proxy settings for the omnibus command vs the chef-client command. I think that functionality is helpful, though we might make it a little more clear that the ignore_proxy = true only affects the chef-client run. Perhaps name the setting 'ignore_proxy_for_converge' or maybe make a converge section under provisioner with ignore_proxy in it?

MarkGibbons:
The other name I thought of for the option was chef_client_noproxy.   Running test kitchen against 7 different supported OS versions and having to update the .kitchen.yml multiple times per cycle per OS version is what I'm trying to avoid.  

Thanks,
   Mark

MarkGibbons:
Now running as a permanent local change, merging with each new release of test kitchen.

cheeseplus:
If you could rename this one to `chef_ignore_proxy` or something clearer and rebase we can get this one merged.

MarkGibbons:
Added another commit to fix an error using the enforce_idempotency .kitchen.yml option.  I found the problem testing my original change.

MarkGibbons:
Renamed the option and rebased.

ryangribble:
Is this going to get merged for the next kitchen release?  The "enforce idempotency" feature added in #875 is broken, until the commit on this PR (https://github.com/test-kitchen/test-kitchen/pull/1068/commits/75fc4f1c8ed8f55fe87358e3308a8fc5dfd0dddf) is merged.  

Alternatively perhaps this commit could be split out onto a separate PR and merged straight away?

MarkGibbons:
Not much chance. Withdrawing this.

PR #1069, opened 2016-06-29
Make RakeTask#config public.

gregsymons:
Given the comment on initialize, I think that a public config accessor was the
original intent, but when define was made private, the accessor was made private
as well. I found having a public accessor convenient for "faking" multinode
support by overriding the rake task to spin up and destroy the dependency
instances before and after the real task.


mwrock:
I'm :+1: here. If you rebase on current master, that should get the appveyor test green.

cheeseplus:
+1

PR #1071, opened 2016-07-04
Use a less volatile recipe for ci tests

mwrock:
This replaces the current package based integration test with a less volitile directory creation test. All we really need here is for a real provisioner and verifier to run and it doesn't so much matter what they do. I originally set this up with a package because that was the absolute simplest way to run a cookbook as is from the supermarket, but the reality is its only a few lines of code to add a test cookbook with a simple directory resource.

Ideally we would use the shell provisioner to keep this chef agnostic, but until we break the chef provisioners out of core, its nice to have this coverage.

### DIFF
--- a/.kitchen.ci.yml
+++ b/.kitchen.ci.yml
@@ -23,4 +23,4 @@ verifier:
 suites:
   - name: default
     run_list:
-      - recipe[filezilla]
+      - recipe[test_cookbook::default]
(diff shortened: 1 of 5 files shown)


mwrock:
chef_apply...so broken...will fix later

mwrock:
No longer WIP.

drrk:
Looks good 👍

cheeseplus:
+1

Issue #593, opened 2015-02-03
--force-formatter is passed to a version of chef-client that does not support it.


damm:
```
/usr/local/lib64/ruby/gems/1.9.1/gems/mixlib-cli-1.5.0/lib/mixlib/cli.rb:191:in `parse_options': invalid option: --force-formatter (OptionParser::InvalidOption)

       	from /usr/local/lib64/ruby/gems/1.9.1/gems/chef-10.16.4/lib/chef/application.rb:64:in `reconfigure'
       	from /usr/local/lib64/ruby/gems/1.9.1/gems/chef-10.16.4/lib/chef/application/solo.rb:155:in `reconfigure'
       	from /usr/local/lib64/ruby/gems/1.9.1/gems/chef-10.16.4/lib/chef/application.rb:70:in `run'
       from /usr/local/lib64/ruby/gems/1.9.1/gems/chef-10.16.4/bin/chef-solo:25:in `<top (required)>'
       	from /opt/chef/bin/chef-solo:23:in `load'
       	from /opt/chef/bin/chef-solo:23:in `<main>'
```  

See https://github.com/test-kitchen/test-kitchen/blob/a83398eba6d850971805447a42270925847acc08/lib/kitchen/provisioner/chef_solo.rb#L50

While I don't disagree with the intent; I am in a special edge case where I'm trying to support an older version of Chef (that should have been upgraded years ago).

BobbyRyterski:
I'd also like to see this fixed.

edmundcraske:
Indeed, should be easy enough to keep backward compatibility here.

lamont-granquist:
Chef 10 is EOL and no longer supported.

fnichol:
If someone on this thread can get the version of Chef where this flag is introduced, I think this can be a simple enough PR. Any helpers?

lamont-granquist:
Should be 11.0.0

BobbyRyterski:
Looks like `--force-formatter` was added in [this commit](https://github.com/chef/chef/commit/1c0e6c22ebf025de71c8d8fde8c273bd654f1553), first available in 11.0.0.

fnichol:
Excellent, thanks all!

Ajedi32:
So what's the workaround for this? Just downgrade to an older version of Chef on the host machine?

lamont-granquist:
upgrade Chef on the virt you're testing.  10.x is EOL and no longer supported.

[...]

PR #1072, opened 2016-07-11
Added support for Chef 10

acondrat:
I found two things required to support Chef 10:
1. do not pass --force-formatter argument
2. handle :auto log-level

### DIFF
--- a/lib/kitchen/provisioner/chef_base.rb
+++ b/lib/kitchen/provisioner/chef_base.rb
@@ -52,7 +52,11 @@ module Kitchen
       default_config :attributes, {}
       default_config :config_path, nil
       default_config :log_file, nil
-      default_config :log_level, "auto"
+      if Gem.loaded_specs['ohai'].version < Gem::Version.new("8.5.1")
+        default_config :log_level, "info"
+      else
+        default_config :log_level, "auto"
+      end
       default_config :profile_ruby, false
       # The older policyfile_zero used `policyfile` so support it for compat.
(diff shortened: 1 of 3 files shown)


cheeseplus:
I am really reluctant to merge this on the basis that Chef 10 is woefully unsupported both from a security perspective and also the ruby is so ancient it is also unsupported.

acondrat:
Some people (like me) still have legacy projects to maintain. I'd vote for migrating to chef 12 with my both hands but it's not always easy. According to https://github.com/test-kitchen/test-kitchen/issues/593 - I am not the only one.

cheeseplus:
I totally understand where you're coming from here as I work in the field with customers all the time here but at some point we need to let Chef 10 die and force people to upgrade.

That said, the code seems relatively unobtrusive and shouldn't break anything. I just don't want to spend any of our cycles on Chef 10 issues at this point in time. I'll see what other maintainers say (many of us are at ChefConf now) but thus far most have not been in favor for the reasons outlined above.

jjasghar:
Academically this makes sense, but the pragmatist in me is screaming 👎 on this PR; but that's just me.

mwrock:
I'm :+1: on this. I'm not crazy about supporting older versions, but if users like @acondrat have a need and are willing to submit a PR that is clean and low impact to address that need, then rejoice! I believe this PR fits that bill.

cheeseplus:
After discussion and review, +1 as well

PR #1076, opened 2016-07-14
Adding test_base_path CLI arg to the diagnose command

tyler-ball:
\cc @test-kitchen/maintainers 

Previously this value would get set/merged into the config on things like `kitchen verify`, but you could not use it with `kitchen diagnose`


mwrock:
:+1:

smurawski:
:+1:

Issue #458, opened 2014-06-06
Provide some way for Chef to know it's running under test


mfischer-zd:
It would be nice if test-kitchen provided a published way for Chef to know that it's running under test-kitchen control.  This would allow us to modify its behavior in testing.  (I'll concede that this is normally undesirable behavior, but there are cases where it's useful or necessary.)

An example of a case where we need it would be access to secret data (we don't use encrypted data bags; we have our own vault).  In production, we have a method `Chef::Node.secret` that fetches the secret from the vault, and its implementation would necessarily differ in a test harness because developer environments and CI bots like Travis won't have access to the vault.  So we'd want the `secret` method to behave differently when run under test-kitchen to do something else (e.g. provide fake secret data).

Obviously we can put a `testing: true` attribute into each suite and/or provisioner definition, but that's pretty clunky and would require it be put into every cookbook's `.kitchen.yml` file.

What do you think about having test-kitchen set an environment variable like `CHEF_TESTING=1` (feel free to select a different name) before executing `chef-client`?

fooshards:
Use case:  Many oracle products, and some other nicer better behaved products gloriously fail installations or at runtime if a host can't properly resolve its external hostname.  If it contains localhost, it fails via some hardcoded logic in their app.  This is obviously disastrous in test-kitchen context, and the solution is to take control of / insert a line in the hostfile.

This is not a desired behavior in production or managed environments.  But it is a NECESSITY, when dealing with disposable, local VM's.  I'd love a canonical way to add some compatibility to a recipe by saying: `only_if node['test_kitchen']` or something similarly succinct.

coderanger:
I should probably state for the record that my earlier arguments were bad and I was totally wrong.

lamont-granquist:
I would be :+1: on this.  I do think that in the wild about 80% of the time it'll be used badly when another approach would be better.  I would bet that half the people on this thread that want this feature shouldn't be using it.  But the other 20% of the use cases aren't possible without it, other than the silly implementation-dependent hacks we have now.  Another case where I firmly believe in the Right to Keep and Bear Footguns.

coderanger:
@lamont-granquist Said far better than I could.

charlesjohnson:
I feel equal parts validated and annoyed.

I continue to be :+1:

martinb3:
> Another case where I firmly believe in the Right to Keep and Bear Footguns.

:+1:

Sauraus:
[...]

PR #1081, opened 2016-07-19
Add environment variables $TEST_KITCHEN and $CI

coderanger:
This would address #458 in a way that works across all provisioners. For Chef we might maybe want to also inject this as a node attribute but I think `if ENV['TEST_KITCHEN']` in recipe code should be fine as is.

### DIFF
--- a/lib/kitchen/configurable.rb
+++ b/lib/kitchen/configurable.rb
@@ -312,7 +312,10 @@ module Kitchen
     end
 
     def env_wrapped(code)
-      (proxy_settings << code).join("\n")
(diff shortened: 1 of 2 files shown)


coderanger:
Also I'm tempted to copy the value of `$CI` from the machine where `kitchen` is run to also be visible in the provisioner and verifier in the same way. That would allow CI-only behaviors.

mwrock:
:+1:

mfischer-zd:
:+1:

cheeseplus:
+1

PR #1082, opened 2016-07-19
Shell verifier enhancements for running test suite within an instance

ytsarev:
This set of commits reuploads test suite under verifier control on each `kitchen verify` execution.
This is critical while developing new tests and extending test suite.

### DIFF
--- a/lib/kitchen/verifier/shell.rb
+++ b/lib/kitchen/verifier/shell.rb
@@ -66,6 +66,8 @@ module Kitchen
       def prepare_command
         return unless config[:remote_upload_path]
         commands = []
+        commands << [sudo('rm -rf'), config[:remote_upload_path] + ' &&']
+        commands << [sudo('mkdir -p'), config[:remote_upload_path] + ' &&']
         commands << [sudo('cp -r'),
                      File.join(config[:root_path],
                                File.basename(config[:upload_path]) + '/*'),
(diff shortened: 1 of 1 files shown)


ytsarev:
```
Offenses:
lib/kitchen/verifier/shell.rb:79:32: C: Style/AlignParameters: Align the parameters of a method call if they span more than one line.
                               File.basename(config[:upload_path]) + "/*"),
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
I strongly believe that the argument is properly aligned and a cop provides false positive here. I also can not reproduce the failure with local `rubocop` of `0.32.1`. 
@cheeseplus could you please look into it ?

smurawski:
@ytsarev As far as style checks go, this project uses `finstyle` which has rubocop pinned at 0.32.1 and specifies a certain base rule set.  

To replicate the failure locally, you can use `rake style` or `bundle exec finstyle` to check your local content.

ytsarev:
@smurawski thanks a lot for clarification! I still think that the `Style/AlignParameters` cop was buggy here (and new rubocop fixes that), but keep it in one line is also fine to me.

Additional question . The new configuration options definitely miss some documentation. Can you suggest the best place for the shell verifier documentation ? I didn't find any existing one

lamont-granquist:
looks like this is a dup with #1382, would need to be reopened after #1382 lands if that doesn't address this.

PR #1083, opened 2016-07-20
Added better routine to install Busser+Plugins

yeoldegrove:
Added better routine to install Busser+Plugins (if not already installed).

This realy optimizes the latency time for the busser run if it is already installed.


yeoldegrove:
Any chance this could be merged?

PR #1085, opened 2016-07-26
Escape paths before running policyfile commands

coderanger:
Some of my coworkers noticed that TK+Policyfiles breaks if there are spaces in some of the enclosing folder names. As the tests say, this comes up with Jenkins a lot, but good hygiene. Notably this is not the full Windows path escape function because [it is huge and annoying](https://github.com/poise/poise/blob/master/lib/poise/utils/win32.rb#L65-L101) and I think this should cover all non-what-were-you-thinking cases.

### DIFF
--- a/.cane
+++ b/.cane
@@ -5,4 +5,5 @@
 --abc-exclude Kitchen::Driver::SSHBase#verify
 --abc-exclude Kitchen::Configurable#wrap_shell_code
 --style-exclude lib/vendor/**/*.rb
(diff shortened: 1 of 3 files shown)


johnbellone:
🎸

mwrock:
:+1:

mwrock:
Needs a rebase

coderanger:
Boo rebase yay merge commits ;-)

PR #1091, opened 2016-07-28
Ability to work with Instances over SSH tunnel.

EYurchenko:
Inspired by:
https://github.com/test-kitchen/test-kitchen/pull/294
https://github.com/test-kitchen/test-kitchen/pull/933

In your .kitchen.yml file you use:
```
transport:
  name: ssh
  ssh_gateway: <gateway>
  ssh_gateway_username: <username at the gateway>
```
and all ssh operations will be tunnelled through this <gateway>.
Basically we replace Net::SSH session in transport/ssh.rb with Net::SSH::Gateway, since this moment all ssh operations go through the tunnel. Remote end 'not being ready' also reported correctly:
```
local@ey-test-kitchen:~/test-kitchen-ey$ bundle exec kitchen converge -l debug
-----> Starting Kitchen (v1.10.2)
D      Berksfile found at /home/local/test-kitchen-ey/Berksfile, loading Berkshelf
D      Berkshelf 4.3.5 library loaded
-----> Creating <default-centos-7>...
D      [SSH] opening connection to local@10.236.62.20<{:user_known_hosts_file=>"/dev/null", :paranoid=>false, :port=>22, :compression=>false, :compression_level=>0, :keepalive=>true, :keepalive_interval=>60, :timeout=>15}> via eyurchenko@10.236.11.190
D      [SSH] connection failed (#<Net::SSH::ConnectionTimeout: timeout during server version negotiating>)
       Waiting for SSH service on 10.236.62.20:22, retrying in 3 seconds
D      [SSH] opening connection to local@10.236.62.20<{:user_known_hosts_file=>"/dev/null", :paranoid=>false, :port=>22, :compression=>false, :compression_level=>0, :keepalive=>true, :keepalive_interval=>60, :timeout=>15}> via eyurchenko@10.236.11.190
D      [SSH] local@10.236.62.20<{:user_known_hosts_file=>"/dev/null", :paranoid=>false, :port=>22, :compression=>false, :compression_level=>0, :keepalive=>true, :keepalive_interval=>60, :timeout=>15}> (echo '[SSH] Established')
       [SSH] Established
D      SSH ready on <default-centos-7>
...
```

### DIFF
--- a/lib/kitchen/transport/ssh.rb
+++ b/lib/kitchen/transport/ssh.rb
@@ -256,25 +256,9 @@ module Kitchen
         # @return [Net::SSH::Connection::Session] the SSH connection session
         # @api private
         def establish_connection_via_gateway(opts)
-          logger.debug("[SSH] opening connection to #{self} via " \
-            "#{ssh_gateway_username}@#{ssh_gateway}")
-          Net::SSH::Gateway.new(ssh_gateway,
-            ssh_gateway_username, options).ssh(hostname, username, options)
-        rescue *RESCUE_EXCEPTIONS_ON_ESTABLISH => e
-          if (opts[:retries] -= 1) > 0
-            message = if opts[:message]
-              logger.debug("[SSH] connection failed (#{e.inspect})")
-              opts[:message]
-            else
-              "[SSH] connection failed, retrying in #{opts[:delay]} seconds " \
-                "(#{e.inspect})"
-            end
-            logger.info(message)
-            sleep(opts[:delay])
-            retry
-          else
-            logger.warn("[SSH] connection failed, terminating (#{e.inspect})")
-            raise SshFailed, "SSH session could not be established"
+          retry_connection(opts) do
+            Net::SSH::Gateway.new(ssh_gateway,
+              ssh_gateway_username, options).ssh(hostname, username, options)
           end
         end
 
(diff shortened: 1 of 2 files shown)


lib/kitchen/transport/ssh.rb
```diff
@@ -226,6 +235,49 @@ def wait_until_ready
...
+        # @option opts [String] :message an optional message to be logged on
+        #   debug (overriding the default) when a rescuable exception is raised
+        # @return [Net::SSH::Connection::Session] the SSH connection session
+        # @api private
+        def establish_connection_via_gateway(opts)
+          logger.debug("[SSH] opening connection to #{self} via " \
+            "#{ssh_gateway_username}@#{ssh_gateway}")
+          Net::SSH::Gateway.new(ssh_gateway,
+            ssh_gateway_username, options).ssh(hostname, username, options)
+        rescue *RESCUE_EXCEPTIONS_ON_ESTABLISH => e
```
mwrock: Would be great if we could reuse the retry logic accross the two connection strategies.

EYurchenko: I tried. Look at this commit https://github.com/test-kitchen/test-kitchen/pull/1091/commits/d171410a2dc1e7d8a2b34c0df84168069dfefb33. But rubocop was very unhappy -(

mwrock: Looks like you hat the gateway and non-gateway connection with the retry all in one method there. I think you did the right thing be making separate methods but I'd suggest extracting the retry logic which appears to be identical in both to be something like:
```
def retry_connection(opts)
  yield
# RESCUE CODE HERE
end
```
Then 

```
def establish_connection_via_gateway(opts)
  retry_connection(opts) do
    Net::SSH::Gateway.new(ssh_gateway,
        ssh_gateway_username, options).ssh(hostname, username, options)
  end
end
```

EYurchenko: Thanks for the hint! Moved the code. Although now we have comments describing parameters identical for three methods -(

mwrock:
:+1:

cheeseplus:
+1

PR #1095, opened 2016-08-09
Dont set ssh key configuration if a password is specified

mwrock:
If a user chooses not to use a ssh public key but prefers to use password authentication instead, test kitchen will not use the password even if one is specified in the config for many drivers.

This would only happen if the driver has a means of providing default ssh keys. kitchen-vagrant and kitchen-vra are two examples. Since both have a means of providing default keys that get passed to the transport, any password provided in the kitchen.yml is ignored.

This will ignore the ssh key settings if a password is explicitly set.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,13 +1,7 @@
 # -*- encoding: utf-8 -*-
 source "https://rubygems.org"
 gemspec
-
-group :guard do
-  gem "guard-minitest"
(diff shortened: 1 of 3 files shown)


cheeseplus:
+1

smford22:
👍

smurawski:
:+1: LGTM :shipit:

PR #1098, opened 2016-08-13
Check the actual value, because `password: nil` shouldn't disable sending the key.

coderanger:
This is currently breaking kitchen-docker.

### DIFF
--- a/lib/kitchen/transport/ssh.rb
+++ b/lib/kitchen/transport/ssh.rb
@@ -416,7 +416,7 @@ module Kitchen
           :ssh_gateway_username   => data[:ssh_gateway_username]
         }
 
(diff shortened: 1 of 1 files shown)


coderanger:
More specifically this breaks everything still using `SSHBase` because of `SSHBase#backcompat_merged_state` which always sets the password to nil.

/cc @mwrock This should be considered a high-prio issue with 1.11 I think.

mwrock:
oy vey :+1:

coderanger:
Going to use this an excuse to de-legacy kitchen-docker but other drivers might be sad too.

mwrock:
I'll aim to get a 1.11.1 release out this weekend and make sure next week's dk has it.

The one redeeming point of all of this is having a 1.11.1 version number.

PR #1100, opened 2016-08-13
Update Ruby versions to test on

tas50:
ChefDK is 2.1.9 now so we should use that. It's going to be 2.3 soon so
we should test on that since ruby-head is 2.4 at the moment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,9 @@
 language: ruby
 
 rvm:
-  - 2.2.4
-  - 2.1.8
+  - 2.1.9
+  - 2.2.5
+  - 2.3.1
   - ruby-head
 
 env:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -2,7 +2,8 @@ language: ruby
 
 rvm:
   - 2.2.4
-  - 2.1.8
+  - 2.1.9
+  - 2.3.0
```
mwrock: should be 2.3.1. that's what we'll be moving to

mwrock:
:+1:  after changing to 2.3.1

tas50:
I bump 2.2 to 2.2.5 as well

cheeseplus:
+1

PR #1103, opened 2016-08-19
Always run `chef install` even if the lock file exists.

coderanger:
This is required to make sure the cookbooks are cached before we run
`chef export` in CI or other non-workstation environments.

Also refactored things a tad.


mwrock:
:+1: on green

coderanger:
Not going to do it in this PR, but we should probably consider having an option to run `chef update` as part of the kitchen run automatically, would be good for development flows I think. /cc @danielsdeleo

danielsdeleo:
@coderanger I think what I'd like is an opt- in for the `chef update` where we configure it to be enabled by default in the generators. That way, most people get auto updates, but it's easier to discover why it happens and easy to turn off. 

Anyway thanks for this fix.

coderanger:
That sounds about right :) Will poke it later tonight after mayhaps.

PR #1102, opened 2016-08-15
support passing Kitchen::Config Hash keys to Kitchen::RakeTasks.new

theckman:
The current implementation of the `Kitchen::RakeTasks.new` method is that you aren't able to modify the config generated by `Kitchen::Config.new`. As such, if you want to change the `test_base_path` value, you need to duplicate most of this helper in to your `Rakefile`.

This change adds the ability to pass a Hash in to the `.new` method, and it will be merged with the config sent to `Kitchen::Config.new`. If nothing is passed in to the `.new` method, it defaults to an empty Hash (`{}`).

PR #1105, opened 2016-08-20
Fix #1104 add supplemental kitchen commands

### LINKED ISSUE #1104: Add supplemental kitchen commands to rake task
## Description

Add supplemental kitchen commands to rake task to facilitate development and troubleshooting. Specifically, add `create`, `converge`, `setup`, `verify`, and `destroy`.
## Kitchen Version

``` ruby
Test Kitchen version 1.11.1
```
## ChefDK Version

``` ruby
Chef Development Kit Version: 0.16.28
chef-client version: 12.9.41
delivery version: master (921828facad8a8bbbd767368bfc72f19bd30e7bd)
berks version: 4.3.3
kitchen version: 1.11.1
```
## Ruby Version

``` ruby
ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-darwin15]
```
## Platform Version

Mac OS X 10.11.6
## Replication Case

When using the built-in `Kitchen::RakeTasks`, there is currently no way to iteratively test minor changes to integration specs without doing a full create-converge-setup-verify-destroy cycle. Adding the aforementioned kitchen commands provides the desired development feature.
## Kitchen Output

Rake task before feature implementation:

``` bash
$ rake -T
rake kitchen:all                             # Run all test instances
...
rake kitchen:default-centos-6                # Run default-centos-6 test in...
...
```

Rake task after feature implementation:

``` bash
$ rake -T
rake kitchen:all                             # Run all test instances
...
rake kitchen:converge:all                    # Run all converge instances
...
rake kitchen:create:all                      # Run all create instances
...
rake kitchen:default-centos-6                # Run default-centos-6 test in...
...
rake kitchen:destroy:all                     # Run all destroy instances
...
rake kitchen:setup:all                       # Run all setup instances
...
rake kitchen:verify:all                      # Run all verify instances
...
```

4-20ma:
- Add supplemental kitchen commands to rake task to facilitate development
  and troubleshooting.

### DIFF
--- a/lib/kitchen/rake_tasks.rb
+++ b/lib/kitchen/rake_tasks.rb
@@ -65,7 +65,7 @@ module Kitchen
           kitchen_commands.each do |cmd|
             namespace cmd do
               # desc "Run #{instance.name} #{cmd} instance"
-              task instance.name { instance.send(cmd) }
+              task instance.name do instance.send(cmd); end
               desc "Run all #{cmd} instances"
               task "all" => config.instances.map(&:name)
             end
--- a/lib/kitchen/rake_tasks.rb
+++ b/lib/kitchen/rake_tasks.rb
@@ -55,7 +55,7 @@ module Kitchen
     # @api private
     def define
       namespace "kitchen" do
-        kitchen_commands = %w(create converge setup verify destroy)
+        kitchen_commands = %w[create converge setup verify destroy]
         config.instances.each do |instance|
           desc "Run #{instance.name} test instance"
           task instance.name do
@@ -65,10 +65,7 @@ module Kitchen
           kitchen_commands.each do |cmd|
             namespace cmd do
               # desc "Run #{instance.name} #{cmd} instance"
-              task instance.name do
-                instance.send(cmd)
-              end
-
+              task instance.name { instance.send(cmd) }
               desc "Run all #{cmd} instances"
               task "all" => config.instances.map(&:name)
             end
@@ -78,9 +75,7 @@ module Kitchen
         desc "Run all test instances"
         task "all" => config.instances.map(&:name)
 
-        kitchen_commands.each do |cmd|
-          task cmd => "#{cmd}:all"
-        end
+        kitchen_commands.each { |cmd| task cmd => "#{cmd}:all" }
       end
     end
   end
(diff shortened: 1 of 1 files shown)


coderanger:
It seems like at that point you should use the actual kitchen commands. Whats the benefit of using rake here?

4-20ma:
@coderanger good question. I use a custom `test_base_path` and `.kitchen.local.yml` which makes it cumbersome to run `kitchen` from the command line. The rake task greatly simplifies running integration tests, especially since #1102 was merged, allowing customization of `Kitchen::Config`. Additionally, I prefer a single workflow to run integration tests; in the event I need to iteratively adjust a spec or a recipe, I would not like to wait to run the entire test converge process. Being able to run the `verify` portion of `test-kitchen` iteratively allows me be more efficient.

coderanger:
It just feels wrong to basically duplicate our whole CLI tool inside Rake. This is probably a good indication that you should fix your config files, not that you should script the whole universe this way.

4-20ma:
@coderanger how would you feel about simply adding the `verify` command?

cheeseplus:
@coderanger pretty much nails my feelings here as well

coderanger:
@4-20ma The provided rake tasks are not intended for day-to-day interactive use at all. It's 99% there so you can make a `rake test` task for use on CI or similar testing platforms where a single entry point is valuable. I suggest you consider your use of rake as a frontend tool for developer workflow. If you really do want it, you can totally add these kinds of tasks in your own Rakefile but I don't think this is the use case we intend to support here so it's just going to be out of place no matter what. That said, it's not really that much code so I won't object heavily if another maintainer wants to champion it :)

cheeseplus:
To be clear, it doesn't hurt anything but also for the same reasons outlined above I'm not seeing the broad appeal - so not a -1 but also not quite a +1

4-20ma:
@coderanger @cheeseplus good points and I appreciate the consideration. I'd be honored if you accept the PR, but understand if you don't. I use rake tasks all the time to conduct odd jobs and it works for me. The proposed change makes it easier for me to perform a specific task, but I'll work around it another way if you reject this. Thanks.

cheeseplus:
This doesn't hurt anything and shouldn't break stuff.

lamont-granquist:
I'm mildly +1 on this since I don't care either way.

PR #1107, opened 2016-08-23
Add a new config option always_update_cookbooks

coderanger:
Add a new config option `always_update_cookbooks` to force both berks and policyfiles to update on each kitchen run.

The intention is to enable this by default in the skeleton generated by `chef generate`.

Open to more intuitive names for the config option. Also the removes Berks 2.x support because I think we should be done with that. I can put it back if anyone objects.


mwrock:
:+1:

danielsdeleo:
LGTM 👍

PR #1114, opened 2016-09-05
Fixing deps

cheeseplus:
We no longer need to depend on braches now that winrmv2 stuff is merged in other projects


mwrock:
We can also uncomment https://github.com/test-kitchen/test-kitchen/blob/master/features/step_definitions/gem_steps.rb with this.

cheeseplus:
Oh goody, we should do that too

mwrock:
:+1:

PR #1119, opened 2016-09-11
Add support for Windows Nano installs via chef provisioners

mwrock:
When used with the nano branch of `mixlib-install` and a provisioner config of:
```
provisioner:
  name: chef_zero
  install_msi_url: https://s3-us-west-2.amazonaws.com/nano-chef-client/chef-12.14.57.appx
```
This is successfully working to test on nano.

Still nedds much cleanup in mixlib-install and some testing to ensure we can get rid of refreshing the `$env:path` the `GetEnvironmentVariable`. I think because we use absolute paths in the chef provosioners, thats ok but needs testing.

### DIFF
--- a/lib/kitchen/configurable.rb
+++ b/lib/kitchen/configurable.rb
@@ -255,7 +255,7 @@ module Kitchen
     def reload_ps1_path
       [
         %{$env:PATH},
-        %{[System.Environment]::GetEnvironmentVariable("PATH","Machine")\n\n}
+        %(try { [System.Environment]::GetEnvironmentVariable("PATH","Machine") } catch {}\n\n)
       ].join(" = ")
     end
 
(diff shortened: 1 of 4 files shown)


mwrock:
Just made this PR much simpler. Just need to suppres errors from `[System.Environment]::GetEnvironmentVariable("PATH","Machine")` because .net core on nano does not support the second parameter. From what I can tell, most if not all cases don't really need this path refresh since we use absolute paths but its best to keep for now in case some plugin depends on it.

mwrock:
looks like i have a few tests to fix.

mwrock:
This is just waiting on a mixlib-install release and then I'll adjust the gemspec.

mwrock:
Removed the WIP flag here.

cheeseplus:
+1

smurawski:
🎉 🚀 🍰

PR #1137, opened 2016-10-16
Adding a Hostname column to kitchen list command

stuartpreston:
## Background

This PR adds the address stored in the `hostname` field within the each Test Kitchen instance statefile to the output of the `kitchen list` command.  

It does this by reading the statefile for each instance and returning the value stored against the `hostname` field.  If there is no such address yet (e.g. in the case where kitchen has not created the instance) `<None>` will be returned.

The benefit of this is mostly for demo purposes, where individual instances sometimes need to be inspected for their state (e.g. if WinRM hasn't been enabled correctly on the target instance and you need to investigate).
## Example output:

capture
## Notes for maintainer:
1. I decided to use the word **Hostname** in the table title - this matches the vocabulary used within kitchen, however it should be noted that it is down to the specific driver as to whether this is actually a Hostname or another address such as an IP address (e.g. in the AzureRM driver what is stored in this field is the value of the public IP or the internal IP address, depending on the configuration).
2. It has been suggested that the width of the table may be too large after making this change and further changes to rationalise the output may be required
3. I have not been able to successfully execute the cucumber tests locally on Windows.  Unless they are executed in the CI process I will need some help ensuring the features have been modified successfully.

Signed-off-by: Stuart Preston stuart@pendrica.com

### DIFF
--- a/features/kitchen_defaults.feature
+++ b/features/kitchen_defaults.feature
@@ -18,7 +18,7 @@ Feature: Test Kitchen defaults
       - name: default
     """
     When I successfully run `kitchen list`
-    Then the output should match /^default-win-81\s+Dummy\s+Dummy\s+Dummy\s+Winrm\s+\<Not Created\>\s+\<None\>$/
+    Then the output should match /^default-win-81\s+<None>\s+Dummy\s+Dummy\s+Dummy\s+Winrm\s+\<Not Created\>\s+\<None\>$/
 
   Scenario: Non-Windows platforms get the Ssh Transport by default
     Given a file named ".kitchen.yml" with:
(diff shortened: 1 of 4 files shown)


coderanger:
:-1: On this as a default. The list output is already getting super noisy, and this is useful for a tiny fraction of the times users want the instance list. We really need a `kitchen list -v`, in which case we could move a bunch of stuff there probably.

stuartpreston:
That's a fair comment @coderanger - I don't have any telemetry on how many times people would want to view the Hostname over (say) the verifier name, or driver name.  Still, the PR is here and thanks to the CI system passes the tests and is ready for someone to take a look at in future if they want.

cheeseplus:
@stuartpreston if you could make this optional via `kitchen list -v` or other option I'd be happy to get this in - or we can close and keep the idea +1

stuartpreston:
Let's close this one for now.

jjasghar:
I really like this idea for a `kitchen list -v` @stuartpreston want to write the feature so this is behind `-v` here?

PR #1141, opened 2016-10-25
Adding transport option "ssh_key_only".

cliles:
When bootstrapping with cloud-init, there is a small window of time where the public ssh key will not be present, this will cause a login failure and eventually fallback to a password prompt.
#704 provided a mechanism to bypass this condition by setting auth_method to publickey, but only does so when a private key is specified. When using a ssh-agent, there is no need to specify a key, but still a need to set auth_method.

This PR is useful to force net/ssh to set the auth_method to publickey but not
attempt to provide a private key. A running ssh-agent can then provide
the key.

### DIFF
--- a/lib/kitchen/transport/ssh.rb
+++ b/lib/kitchen/transport/ssh.rb
@@ -422,6 +422,10 @@ module Kitchen
           opts[:auth_methods] = ["publickey"]
         end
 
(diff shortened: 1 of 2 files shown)


cheeseplus:
+1

Issue #1146, opened 2016-11-09
Re-Enable Code Climate


smurawski:
## Description

Travis builds are broken because `codeclimate-test-reporter` was deprecated and this broke builds on master.

We need to change how we upload the codeclimate tests.

PR #1147, opened 2016-11-09
Remove codeclimate-test-reporter as that has been deprecated

smurawski:
Fixes the build for master. 

Tracking bug to re-enable codeclimate reporting - https://github.com/test-kitchen/test-kitchen/issues/1146

Signed-off-by: Steven Murawski <steven.murawski@gmail.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,8 +64,3 @@ matrix:
 
 notifications:
   irc: "chat.freenode.net#kitchenci"
-
-addons:
-  code_climate:
-    repo_token:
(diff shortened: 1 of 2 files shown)


coderanger:
Just be clear, where are you seeing that this is deprecated? It changed with the 1.0 release but I don't see anything saying it is deprecated.

smurawski:
@coderanger the `codeclimate-test-reporter` as used by our build was deprecated and was breaking the build.  there is another method to get the codeclimate results, and #1146 was filed to get that work done. https://travis-ci.org/test-kitchen/test-kitchen/jobs/174499510#L383

coderanger:
@smurawski https://github.com/codeclimate/ruby-test-reporter#upgrading-from-pre-10-versions you still need the gem, the part that needs to change is the build task and the spec helper.

smurawski:
@coderanger that's fine, but I didn't have the cycles to investigate the changes at the moment, so having build that wasn't breaking on an ancillary service is more important.

If you've got the cycles and want fix it up to re-enable, go for it.  Otherwise, I'll dig in tomorrow.

coderanger:
Sure, no argument on removing it if no one has time right now. Just wanted to make sure I didn't miss anything because I use the same gem on some other projects (but with a semver pin so I'm not immediately affected by the breaking changes).

Issue #809, opened 2015-08-05
Test Kitchen should use omnitruck's -d option by default


lamont-granquist:
The -d option to install.sh sets up a caching directory (something under /tmp/kitchen should get used) so that you can get 'latest' semantics while at the same time not re-pulling down artifacts that you already have on the image.

edehde:
Any update on this?

We are currently using `chef_omnibus_install_options: -d /tmp/vagrant-cache/chef-omnibus` with kitchen-vagrant (and vagrant cachier).

afiune:
We are starting to work on this feature. Stay tuned! 🎉

PR #1149, opened 2016-11-10
Added `cache` interface for Drivers so that provisioners can leverage 

afiune:
Introducing a new interface `cache_directory` inside the Drivers for provisioners
to leverage. When a driver is ready to implement it, the provisioner will
detect it and use the cache directory to pass the `-d` option to omnitruck 
so we can cache the packages and avoid re-pulling down artifacts that you
already have on the image.

Closes https://github.com/test-kitchen/test-kitchen/issues/809

Signed-off-by: Salim Afiune <afiune@chef.io>

### DIFF
--- a/lib/kitchen/provisioner/chef_base.rb
+++ b/lib/kitchen/provisioner/chef_base.rb
@@ -47,6 +47,13 @@ module Kitchen
 
       default_config :require_chef_omnibus, true
       default_config :chef_omnibus_url, "https://omnitruck.chef.io/install.sh"
+      # `chef_omnibus_cache` is an option that will let us be specific with the
+      # directory we want omnibus packages to be cached, it will be useful for
+      # other drivers that can mount or share a folder to avoid downloading the
+      # same software every converge
+      default_config :chef_omnibus_cache do |provisioner|
+        provisioner.windows_os? ? "$env:TEMP\\kitchen\\omnibus" : "/tmp/kitchen/omnibus"
+      end
       default_config :chef_omnibus_install_options, nil
       default_config :run_list, []
       default_config :attributes, {}
@@ -156,6 +163,15 @@ module Kitchen
       # @return [Hash] an option hash for the install commands
       # @api private
       def install_options
+        # Verify if the `-d` option has already been passed, then we
+        # don't use config[:chef_omnibus_cache]
+        cache_dir_option = ' ' << "-d #{config[:chef_omnibus_cache]}"
+        if config[:chef_omnibus_install_options].nil?
+          config[:chef_omnibus_install_options] = cache_dir_option
+        elsif config[:chef_omnibus_install_options].match(/\s*-d\s*/).nil?
+          config[:chef_omnibus_install_options] << cache_dir_option
+        end
+
         project = /\s*-P (\w+)\s*/.match(config[:chef_omnibus_install_options])
         {
           :omnibus_url => config[:chef_omnibus_url],
--- a/lib/kitchen/provisioner/chef_base.rb
+++ b/lib/kitchen/provisioner/chef_base.rb
@@ -163,15 +163,7 @@ module Kitchen
       # @return [Hash] an option hash for the install commands
       # @api private
       def install_options
-        # Verify if the `-d` option has already been passed, then we
-        # don't use config[:chef_omnibus_cache]
-        cache_dir_option = ' ' << "-d #{config[:chef_omnibus_cache]}"
-        if config[:chef_omnibus_install_options].nil?
-          config[:chef_omnibus_install_options] = cache_dir_option
-        elsif config[:chef_omnibus_install_options].match(/\s*-d\s*/).nil?
-          config[:chef_omnibus_install_options] << cache_dir_option
-        end
-
+        verify_omnibus_directory_option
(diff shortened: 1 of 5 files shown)


coderanger:
How will this cache anything currently? I feel like there is a second half that isn't being mentioned.

afiune:
You are right Noah, we are trying to do a first iteration of this feature that will not affect other drivers but they could potentially take advantage of this config. 

An example is this PR that I created in `kitchen-vagrant`: https://github.com/test-kitchen/kitchen-vagrant/pull/248

coderanger:
@afiune Are there any drivers other than Vagrant that this could be used for? It won't help for anything on a remote cloud, docker and dokken already have their own systems for this, so this seems like maybe it should be 100% on the vagrant plugin side.

tduffield:
@coderanger Any driver _could_ take advantage of this cache directory. There are scenarios where caches could be mounted via EBS or Docker volumes. And given that the directory needs to be specified inside Chef::Base, it makes more sense to put this setting here rather than having the provider monkeypatch the install command.

tduffield:
@afiune could we add some tests to `chef_base_spec.rb` to test the expected defaults?

coderanger:
@tduffield While not strictly speaking false, this still feels pretty YAGNI unless someone has an actual plan for how this would work on anything other than Vagrant. I could see maybe the libvirt or hyperv plugins, but I don't think either of those is currently set up for any kind of shared folders or similar. So yeah, calling YAGNI unless someone can cite actual stuffs that would use this :)

tduffield:
@coderanger okay, do you have a suggestion on what an alternative would be?

coderanger:
@tduffield Make this a driver option and implement it in the driver? Another option would be to include a provisioner along with the vagrant driver, like how dokken works.

mwrock:
@coderanger tapping into the omnibus cache needs to happen in the core provisioner. True that it accomplishes nothing unless a driver takes advantage of it (and vagrant is one of the few drivers that can) but doing it this way provides much simpler implementation than if you wanted to do it all in the driver.

coderanger:
If nothing else can we add some comments for future-us so that we remember why the code is there despite probably only being used in one plugin?

afiune:
@coderanger will do!

spec/kitchen/provisioner/chef_base_spec.rb
```diff
@@ -284,6 +293,25 @@ def calculate_path(path, _opts = {})
         cmd
       end
 
+      it "will use chef_omnibus_cache to provide a cache directory" do
```
tduffield: Any chance we could also include a test case where the user provides install_options other than -d? Maybe -v?

afiune: sure thing

afiune: just did

tduffield: So you confirmed that it won't clobber the -d. Can you add a test case where we append -d to install_options that already include other values?

spec/kitchen/provisioner/chef_base_spec.rb
```diff
@@ -438,6 +479,21 @@ def calculate_path(path, _opts = {})
...
+          with(default_version, true, install_opts_clone).returns(installer)
+        cmd
+      end
+
+      it "will have the same behavior on windows" do
+        config[:chef_omnibus_cache] = "$env:TEMP\\dummy\\place"
+        config[:chef_omnibus_install_options] = "-P cool -v 123"
+        install_opts_clone = install_opts.clone
+        install_opts_clone[:sudo_command] = ""
+        install_opts_clone[:install_flags] = "-P cool"
```
tduffield: This should be `-project` instead of `-P`

spec/kitchen/provisioner/chef_base_spec.rb
```diff
@@ -438,6 +479,21 @@ def calculate_path(path, _opts = {})
...
+        install_opts_clone[:install_flags] << " -v 123"
```
tduffield: This should be `-version` instead of `-v`

spec/kitchen/provisioner/chef_base_spec.rb
afiune: fixed

lib/kitchen/provisioner/chef_base.rb
```diff
@@ -47,6 +47,13 @@ class ChefBase < Base
 
       default_config :require_chef_omnibus, true
       default_config :chef_omnibus_url, "https://omnitruck.chef.io/install.sh"
+      # `chef_omnibus_cache` is an option that will let us be specific with the
+      # directory we want omnibus packages to be cached, it will be useful for
+      # other drivers that can mount or share a folder to avoid downloading the
+      # same software every converge
+      default_config :chef_omnibus_cache do |provisioner|
```
tduffield: provisioners outside of Chef _could_ take advantage of this paradigm.  Should we rename it (and the internal path) to something more genertic? Maybe `:provisioner_cache` and `/tmp/provisioner/cache`? Then, the Chef provisioner itself could leverage the provisioner cache in its own way.

PR #1155, opened 2016-11-27
Fix to work with Thor 0.19.2

coderanger:
Thor released a new version today (November 26th) that requires the default value for an option match that option's type. As such, Test Kitchen is currently non-functional via gem install unless you back-rev the `thor` gem. This fix seems to be all that we need, currently scouring for other mismatches but it complained about the `init` command even though it wasn't the command being run so I think this should cover it.

Given the "gem/bundle installs are 100% busted" we should probably fast track this to a patch release.

### DIFF
--- a/lib/kitchen/generator/init.rb
+++ b/lib/kitchen/generator/init.rb
@@ -34,7 +34,7 @@ module Kitchen
       class_option :driver,
         :type => :array,
         :aliases => "-D",
-        :default => "kitchen-vagrant",
+        :default => %w{kitchen-vagrant},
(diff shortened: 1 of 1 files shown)


coderanger:
I can merge this in the morning and do release prep but I don't think I have rubygems push perms.

smurawski:
@coderanger I'll get you set up with push rights in the morning.

coderanger:
I mentioned on Slack but just for the record here: Thor released 0.19.3 late last night to bump this error down to a warning. Still good to fix but no need for an ASAP release now.

smurawski:
Thanks @coderanger!

PR #1160, opened 2016-12-02
I don't really understand how no tests caught this.

thommay:
( @afiune caught this, happily )
Signed-off-by: Thom May <thom@chef.io>


mwrock:
oh man i hit this a week or so ago and didn't get why i never hit it again but was focused on another bug and never followed up.

PR #1173, opened 2016-12-24
Encrypt data bags automatically on provisioning sandbox

lawrencejones:
Hey all, this is a feature that I've been really missing when writing tests for cookbooks recently. The tl;dr is:

> You can now use unencrypted data bag fixtures in `/test/fixtures/data_bags/*` etc that Kitchen will encrypt when it pushes them into the testing sandbox. This allows users to edit these fixtures with their favourite editor, and see the contents of the fixtures when looking at a git diff.

I'm open to feedback if people feel like we should use a different config key etc, so I haven't finalised this with a changelog and version bump just yet. I'll do that if people think this looks like the right approach.

---

Users can now supply an `encrypt_data_bags` boolean key to the
provisioner config inside .kitchen.yml that tells Kitchen to encrypt the
data bags it pushes into the test machine when converging.

This allows users to maintain test data bag fixtures that are in
plaintext while accessing them as encrypted bags inside their recipes,
facilitating easier development and removing the need to edit JSON
fixtures using Chef solo/zero.

### DIFF
--- a/test-kitchen.gemspec
+++ b/test-kitchen.gemspec
@@ -30,6 +30,7 @@ Gem::Specification.new do |gem|
   gem.add_dependency "safe_yaml",       "~> 1.0"
   gem.add_dependency "thor",            "~> 0.18"
   gem.add_dependency "mixlib-install",  ">= 1.2", "< 3.0"
+  gem.add_dependency "chef",            "~> 12.17"
 
   gem.add_development_dependency "pry"
   gem.add_development_dependency "pry-byebug"
--- a/lib/kitchen/provisioner/chef/common_sandbox.rb
+++ b/lib/kitchen/provisioner/chef/common_sandbox.rb
@@ -17,6 +17,7 @@
 # limitations under the License.
 
 require "json"
+require "chef/encrypted_data_bag_item"
 
 module Kitchen
   module Provisioner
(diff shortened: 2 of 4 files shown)


test-kitchen.gemspec
```diff
@@ -30,6 +30,7 @@ Gem::Specification.new do |gem|
   gem.add_dependency "safe_yaml",       "~> 1.0"
   gem.add_dependency "thor",            "~> 0.18"
   gem.add_dependency "mixlib-install",  ">= 1.2", "< 3.0"
+  gem.add_dependency "chef",            "~> 12.17"
```
mwrock: We take pains to avoid a direct dependency on chef keeping test kitchen config manager agnostic. If you absolutely need this, I would recommend making this a separate provisIoner plugin.

lawrencejones: Sorry for the delay, been a long Christmas break!

I thought this might be an issue :/ unfortunately I can't see any robust way of encrypting the data bags without using the Chef codebase to do so.

Have you got any examples of plugins that augment the default sandbox provisioning process? I think this is a huge usability boost when writing kitchen tests, and wouldn't mind publishing it as a separate plugin if need be.

mwrock:
https://github.com/mwrock/kitchen-nodes is an example provisIoner plugin that augments the node files in the sandbox.

lawrencejones:
@mwrock that's exactly what I was looking for. I'll get this refactored into a separate gem and we can avoid adding chef as a hard dependency.

Thanks for all your help! Going to close this PR in favour of your suggested approach.

lawrencejones:
For anyone who finds this PR and is looking for a solution, https://github.com/lawrencejones/kitchen-encrypt-databags should solve this for you.

PR #1174, opened 2017-01-07
Remove ruby 2.1.9 from test matrix

mwrock:
The nio4r gem bumb to 2.0.0 is not compatible with ruby 2.1 anbd test-kitchen builds are broken. Since ruby 2.1 is retired, this removes it from our travis and appveyor matrix.

Signed-off-by: Matt Wrock <matt@mattwrock.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: ruby
 
 rvm:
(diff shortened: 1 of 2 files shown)


cheeseplus:
:shipit:

PR #1178, opened 2017-01-17
Export no_proxy from kitchen config

itmustbejj:
Currently the only way no_proxy is set is if you are using proxy settings from ENV vars. This addresses the use case of when you define your proxy settings in a kitchen config.

### DIFF
--- a/lib/kitchen/configurable.rb
+++ b/lib/kitchen/configurable.rb
@@ -317,7 +317,7 @@ module Kitchen
     end
 
     def proxy_setting_keys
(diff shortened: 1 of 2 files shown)


coderanger:
Seems reasonable once tests pass.

itmustbejj:
With my change to `proxy_settings_keys` the extra logic in `proxy_settings` becomes redundant, and at that point it's just wrapping `resolve_proxy_settings_from_config`. I opted to just call `resolve_proxy_settings_from_config` directly.

itmustbejj:
There are also no other references to `proxy_settings` in the code.

PR #1188, opened 2017-01-26
Transfer tests when using shell verifier

xmik:
When using shell verifier with option: `remote_exec: true`, I find out that it did not transfer any test files into test-kitchen instance. Thus, I added boolean option: `transfer_files` to shell verifier. See the 1st commit message. I think this is an essential option for this verifier.

In the 2nd commit I wanted to avoid repetition, so I moved some methods into base verifier and removed  them from busser and shell verifiers. I don't know if you would like that commit.

### DIFF
--- a/lib/kitchen/verifier/base.rb
+++ b/lib/kitchen/verifier/base.rb
@@ -227,6 +227,49 @@ module Kitchen
       def prefix_command(script)
         config[:command_prefix] ? "#{config[:command_prefix]} #{script}" : script
       end
+
+      # Returns an Array of common helper filenames currently residing on the
+      # local workstation.
+      #
+      # @return [Array<String>] array of helper files
+      # @api private
+      def helper_files
+        glob = File.join(config[:test_base_path], "helpers", "*/**/*")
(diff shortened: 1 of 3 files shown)


xmik:
This PR may satisfy the intention of the PR: https://github.com/test-kitchen/test-kitchen/pull/1082

thewyzard44:
I got started on #1382 before I noticed your pull request, here.  I like what you're up to, but there are many other things wrong with the shell verifier that need addressed at the same time.  Let's start with keeping the changes contained to the shell verifier and then we can potentially refactor into the base class at a later date.

Give me a couple more days to mature 1382, but I'd appreciate your input to make sure that your use cases are covered.  My initial question for you is what are your thoughts about simply conditioning the sandbox upload if remote_exec is set to true?

xmik:
Hi @thewyzard44, I didn't have time to answer earlier, sorry. 

I think it is sensible to avoid additional option `transfer_files` and just transfer the files if `remote_exec` is set to true. It makes sense.

lamont-granquist:
closing in favor of #1382, please open a new issue after #1382 lands if there's still problems.

PR #1189, opened 2017-01-30
Pinning the same as berks

cheeseplus:
Prevents fun errors like the below

```
Could not load or activate Berkshelf (Unable to activate berkshelf-5.5.0, because thor-0.19.4 conflicts with thor (< 0.19.2, ~> 0.19))
```

A more comprehensive fix of not depending on the berkshelf gem and simply shelling out should be forthcoming.


tas50:
We should merge this and ship it. I'm constantly removing gems

PR #1198, opened 2017-02-26
Require Ruby 2.2.2+ and bump test deps

tas50:
We're constraining deps as the Ruby world moves on. It's time to require
Ruby 2.2.2 as all other tooling in the Chef ecosystem does.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ branches:
 
 matrix:
   include:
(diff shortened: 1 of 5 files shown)


coderanger:
Huge -1 on this, I run 2.1 on my CI tests for versions of Chef that shipped with 2.1 and many of those are still being supported for community cookbooks.

PR #1201, opened 2017-03-03
Update to latest train (and net-ssh 4)

tduffield:
Net::SSH 4 fixes a lot of the monkey patching we had to do in our tests.

Note: The failing cucumber test is due to readline 5.4.0 (see https://github.com/ConnorAtherton/rb-readline/pull/135). 

Signed-off-by: Tom Duffield <tom@chef.io>


tduffield:
Please note: this is required to unblock the ChefDK pipeline

PR #1205, opened 2017-03-10
Add support for auth proxy tests

tduffield:
Signed-off-by: Tom Duffield <tom@chef.io>


tduffield:
This will be necessary for the proxy tests to pass with the proxy_tests master.

PR #1214, opened 2017-04-02
**Experimental** Example of config deprecations

wrightp:
This PR is *experimental* and shows an example of deprecating only a few provisioner options: `require_chef_omnibus`, `chef_omnibus_install_options` and `chef_omnibus_url`

I want to get this out early so we can beat it up and get it right.

The overall goal for these changes is to unify the two mixlib-install code paths. One using the latest API interface and the other using the ScriptGenerator. By using the latest interface we gain the most up the date benefits and features of mixlib-install. This also includes the user-agent header feature.

Phase 1 is to release a patch or minor version that provides deprecation warnings AND the new settings to fix the warnings. In this PR we show that we still use the original settings in the code base. The new settings translate into the old settings. No existing feature changes, and all unit tests should pass as is.

require_chef_omnibus is two features in one - versioning and skipping the bootstrap install. When set to true it installs latest, when set to a version it will try to install the specified version, and when set to false it will not install anything.

The new settings: product_name, product_version, and channel are to be used in conjunction with one another. For versioning, require_chef_omnibus will be set based on the product_version. For skipping the bootstrap a new setting was added: skip_bootstrap. See the code for how it sets the old setting.

When this is ready for deprecation we will change the #add_config_deprecation! level_log to :error and remove the translation logic. Then we can update the setting throughout the code base. This will be considered Phase 2 although it can be done iteratively since we set different log levels on the deprecations.

chef_omnibus_url was renamed to install_script_url. The new mixlib-install API automatically manages the install script URLs based on platform. A new option will need to be added to allow the URLs paths to be configurable. Once this is done, install_script_url will default to nil when chef_omnibus_url is removed.

chef_omnibus_install_options informs the user of the preferred way to bootstrap chef. Since this setting is used for a different install path there is no need to translate the values.

The deprecation logic was added to Configurable to allow all config options (not just the provisioner settings) to be deprecated in the same way.

Known warts: 
- the logic for executing the warn method in Configurable. I'll get to that. I just wanted to see it work.
- I may move the deprecation checking logic in ChefBase#initialize to another method or class. Not sure yet.
- No tests for the deprecation logic in Configurable. Getting the deprecations to set the existing settings correctly work was my priority this round. I'll continue to add tests.

@cheeseplus please add mentions
@schisamo

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,3 +13,5 @@ end
 group :test do
   gem "codeclimate-test-reporter", "~> 1.0", ">= 1.0.3", require: nil
 end
+
+gem "mixlib-install", github: "chef/mixlib-install"
--- a/lib/kitchen/configurable.rb
+++ b/lib/kitchen/configurable.rb
@@ -405,14 +405,19 @@ module Kitchen
       end
 
       unless warnings.empty?
-        caller[0].instance_eval do
-          warn warnings.map { |w| w[:message] }.join("\n")
-        end
+        output = concat_messages(warnings)
+        caller[0].instance_eval { warn(output) }
       end
 
-      unless errors.empty?
-        raise DeprecationError, errors.map { |e| e[:message] }.join("\n")
-      end
+      raise DeprecationError, concat_messages(errors) unless errors.empty?
+    end
+
+    # Join config deprecation messages into a single message.
+    #
+    # @param [Array<Hash>] of config deprecations
+    # @return [String] of concatenated deprecation messages
+    def concat_messages(deprecations)
+      deprecations.map { |d| d[:message] }.join("\n")
     end
 
     # Class methods which will be mixed in on inclusion of Configurable module.
--- a/lib/kitchen/configurable.rb
+++ b/lib/kitchen/configurable.rb
@@ -176,17 +176,18 @@ module Kitchen
     end
 
     # Add a new config deprecation message to the config_deprecations collection.
+    # Setting a log level of :pending will take no action
     # Setting a log level of :warn will log the warning and provide details on how to
     # proactively fix the deprecation.
-    # Setting a log level of :error will raise a DeprecationError with the details for each
-    # provisioner setting.
+    # Setting a log level of :error will raise a DeprecationError with the
+    # details for each setting.
     #
     # @param log_level [Symbol] deprecation log level
-    # @param setting_name [String] name of the deprecated provisioner setting
+    # @param setting_name [String] name of the deprecated setting
     # @param message [String] message providing details to fix the issue
     # @raise [ArgumentError] if an invalid log level is set
     def add_config_deprecation!(log_level, setting_name, message)
-      log_levels = [:warn, :error]
+      log_levels = [:pending, :warn, :error]
       unless log_levels.include?(log_level)
         raise ArgumentError, "Config deprecation log level must be one of: #{log_levels.join(",")}"
       end
@@ -397,10 +398,13 @@ module Kitchen
       errors = []
 
       config_deprecations.each do |dep|
-        if dep[:log_level] == :error
+        case dep[:log_level]
+        when :error
           errors << dep
-        else
+        when :warn
           warnings << dep
+        else
+          # into the nether
         end
       end
 
--- a/lib/kitchen/configurable.rb
+++ b/lib/kitchen/configurable.rb
@@ -192,7 +192,7 @@ module Kitchen
       end
 
       tense = log_level == :error ? "has been" : "will be"
-      message.prepend("#{setting_name} setting #{tense} deprecated\n")
+      message.prepend("** #{setting_name} setting #{tense} deprecated\n")
       config_deprecations << { log_level: log_level, message: message }
     end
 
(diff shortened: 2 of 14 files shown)


lib/kitchen/provisioner/chef_base.rb
```diff
@@ -115,6 +131,28 @@ class ChefBase < Base
...
+          config[:require_chef_omnibus] = true
+        end
+
+        if config.key?(:chef_omnibus_url)
+          add_config_deprecation! :warn, "chef_omnibus_url", <<-EOF
+  To override the install script url use 'install_script_url'.
+  Note that Windows will now automatically use the appropriate install.ps1 script.
+EOF
+        else
+          config[:chef_omnibus_url] = config[:install_script_url]
```
wrightp: Looks like this broke script_generator.
https://github.com/chef/mixlib-install/blob/master/lib/mixlib/install/script_generator.rb#L213-L218
I'll look into this, but we may have to set the url to install.sh like before as default.

wrightp: @cheeseplus mentioned that the override for this setting should allow users to pull their own install scripts or the install scripts from an a local server if test instance don't have internet access. mixlib-install would have to be updated to make this configurable for the core API. Also note this ability is broken for Windows instances due to how the script_generator works. Are we sure we should even support this as a feature? If not, we can remove this setting.

wrightp: What is a current legitimate use case for changing chef_omnibus_url? The only one I can think of is possibly hitting https://omnitruck-acceptance.chef.io, but I don't see how that would be useful for a TK test suite.

I think we should deprecate this setting. Allow mixlib-install to manage the install script urls internally within TK since changing it, even to .../install.ps1, actually breaks things anyway.

wrightp: You can ignore this. I think I got this all sorted out.

lib/kitchen/configurable.rb
```diff
@@ -174,6 +175,27 @@ def windows_os?
...
+    # @param message [String] message providing details to fix the issue
+    # @raise [ArgumentError] if an invalid log level is set
+    def add_config_deprecation!(log_level, setting_name, message)
+      log_levels = [:warn, :error]
+      unless log_levels.include?(log_level)
+        raise ArgumentError, "Config deprecation log level must be one of: #{log_levels.join(",")}"
+      end
+
+      tense = log_level == :error ? "has been" : "will be"
+      message.prepend("** #{setting_name} setting #{tense} deprecated\n")
```
coderanger: Mutating strings in logging APIs is real fast road to bad times.

wrightp: agreed. I will get to this the entire logging fiasco.

lib/kitchen/provisioner/chef_zero.rb
```diff
@@ -223,6 +228,19 @@ def shim_command
...
+      # Return ruby bin directory. This method will be removed once
+      # chef_omnibus_root and ruby_bindir are deprecated.
+      #
+      # @return [String] ruby bin directory
+      # @api private
+      def ruby_bindir
+        if config.key?(:ruby_bindir)
+          config[:ruby_bindir]
+        else
+          remote_path_join(%W{#{config[:chef_omnibus_root]} embedded bin})
```
wrightp: fix windows path

lib/kitchen/verifier/busser.rb
```diff
@@ -198,9 +196,14 @@ def helper_files
       end
 
       def install_command_vars
-        ruby = remote_path_join(config[:ruby_bindir], "ruby")
+        ruby_bindir = if config.key?(:ruby_bindir)
+                        config[:ruby_bindir]
+                      else
+                        remote_path_join(%W{#{config[:chef_omnibus_root]} embedded bin})
```
wrightp: fix windows path (test)

lib/kitchen/provisioner/chef_base.rb
```diff
@@ -115,6 +125,55 @@ class ChefBase < Base
...
+          config[:chef_omnibus_install_options] = nil
+        end
+
+        if config.key?(:chef_metadata_url)
+          add_config_deprecation! :pending, "Provisioner: chef_metadata_url", <<-EOF.gsub(/^\s*/, "")
+            'chef_metadata_url' is no longer used and can be safely removed from the config.
+          EOF
+        end
+
+        if config.key?(:install_msi_url)
```
wrightp: this does work by several layers of redirection. need to look into this

Gemfile
```diff
@@ -13,3 +13,5 @@ end
 group :test do
   gem "codeclimate-test-reporter", "~> 1.0", ">= 1.0.3", require: nil
 end
+
+gem "mixlib-install", github: "chef/mixlib-install"
```
schisamo: I assume this will be pulled in via `test-kitchen.gemspec` before we merge this?

wrightp:
This was an experimental branch created to begin a discussion. It has served it's purpose and an RFC draft is in progress.

PR #1220, opened 2017-04-22
WIP: Move proxy driver to use Provisioner::Base and not the legacy SSHBase…

smurawski:
🚧 
….  All the defaults from SSH base should be handled elsewhere (like in transport).

This change allows the proxy driver to work with the reboot and continue behavior introduced in 1.10.0.

Signed-off-by: Steven Murawski <steven.murawski@gmail.com>

### DIFF
--- a/lib/kitchen/driver/proxy.rb
+++ b/lib/kitchen/driver/proxy.rb
@@ -30,7 +30,7 @@ module Kitchen
     # the driver was created.
     #
     # @author Seth Chisamore <schisamo@opscode.com>
-    class Proxy < Kitchen::Driver::SSHBase
+    class Proxy < Kitchen::Driver::Base
       plugin_version Kitchen::VERSION
 
       required_config :host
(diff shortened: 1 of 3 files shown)


coderanger:
@smurawski Do you remember off-hand what was unfinished here?

robbkidd:
Rebased to resolve merge conflicts and re-run tests.

lib/kitchen/driver/proxy.rb
```diff
@@ -42,29 +42,36 @@ class Proxy < Kitchen::Driver::SSHBase
       def create(state)
         # TODO: Once this isn't using SSHBase, it should call `super` to support pre_create_command.
```
robbkidd: Any thoughts on this comment? This PR moves off of SSHBase, so should a `super` be added?

robbkidd:
```
Running kitchen create -c 5 ...
-----> Starting Kitchen (v1.20.0)
-----> Creating <default-ubuntu-1604>...
root@localhost password:
```

... replaced by ...

```
-----> Installing Chef Omnibus (install only if missing)
       Downloading https://omnitruck.chef.io/install.sh to file /tmp/install.sh
       Trying wget...
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```

... so ... progress?

I think the `proxy` driver punting to a transport fails when in a proxied environment because there appears to be no code in the Transport plugin umbrella to handle `http_proxy` (and its cousins) coming from the kitchen config or from the environment.

`configurable` things have [`resolve_proxy_settings_from_config`](https://github.com/test-kitchen/test-kitchen/blob/7bc2c9a25a028175a44ed05a8999ac0997debd07/lib/kitchen/configurable.rb#L359) available via a `wrap_shell_code` method, but nothing in the core transports use this method to inject environment variables into the command to run on the "remote" host. I suspect `proxy` has been working heretofore because, descending from `SSHBase`—it had [`env_cmd`](https://github.com/test-kitchen/test-kitchen/blob/7bc2c9a25a028175a44ed05a8999ac0997debd07/lib/kitchen/driver/ssh_base.rb#L266) under the hood shimming in all those proxy settings to any command run on the host-under-test.

robbkidd:
I reckon I'm going to pause on getting this fixed up.

I see that the SSH transport does not make use of any of `http_proxy`&friends from `Configurable` and it probably should. I assume that it has not used them previously because transports were more about moving files to the host-under-test than they were about executing commands upon them. But it a world where the `proxy` driver is going to delegate command execution to a transport, `Transport::Base` or `Transport::SSH` and `Transport::Winrm` will need to take advantage of the `wrap_shell_code` (or its underlying `resolve_proxy_settings_from_config`) available from `Configurable`.

PR #1233, opened 2017-05-20
remove badges

lamont-granquist:
i really want to get rid of the dev dependency on codeclimate.  gem deps are actively harmful if not necessary or maintained.

i'm skeptical that anyone uses any of these though and that they're included more out of religious guilt because we think we 'should' be using them.

- the simple code coverage metric across the project is pretty useless (https://labs.ig.com/code-coverage-100-percent-tragedy)
- nobody is checking the dependency status or we would have caught the fact that we were still using berkshelf 4.x in tests
- code climate is still available, and i tend to see no effort to clean up overly complicated classes
- inline docs is meh

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 .yardoc
 binstubs
 Gemfile*.lock
+Gemfile.local
 InstalledFiles
 _yardoc
 coverage
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,8 +64,3 @@ matrix:
 
 notifications:
   irc: "chat.freenode.net#kitchenci"
-
-addons:
-  code_climate:
-    repo_token:
-      secure: "lcqi3hbalLTinxwsl+um1aN1dgijBrODSMseGEJMJGkofz3VZ+Ofuuwp9x9VjvewuiUFHsiPD5SQ6hx8N+L3wXB6qA+HTmIrXecL7VjehbImEyOuu4/++vcTdpTINAd2Qt/KuJ1eY0okSwVmvtX1VD8gYD8yrlMKdj6uexf8Bgs="
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,6 @@ bundler_args: --without guard integration
 
 script:
   - bundle exec rake
-  - bundle exec codeclimate-test-reporter
   - bundle install --with integration
   - export KITCHEN_YAML=.kitchen.ci.yml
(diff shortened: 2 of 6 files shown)


coderanger:
Why remove coverage checks?

lamont-granquist:
clearly nobody is getting the gemnasium e-mails and acting on them.  what we have to do to reconcile gem versions when we build chef-dk is a more effective check (or will be a more effective check after it gets fixed to actually work -- i'm still digging into how we pinned to berks 4.x here)

are people actually using code coverage checks in this project?  if its not being actively used it should get removed.  i'm way too busy with other things to care and simplecov is just another dep to conflict with some other gem and block shipping product to me...

coderanger:
Why are dev gems even being installed or considered as part of ChefDK?

lamont-granquist:
we ship `chef verify` which needs dev gems to test against, in order for that to be replicatable by end-users and behind firewalls and stuff we need to ship dev gems.  we don't try to lock them all in one mega-shitpile lockfile, but we do generate individual Gemfile.locks per gem that gets verified which is then pinned by the chef-dk runtime lockfile constraints.

even if we didn't ship them installed and users installed them (which would require some additional tomfoolery to generate and ship the Gemfile.lock and then not ship the gems) that still doesn't reduce the possibility of conflicts when we're trying to run `chef verify` in order to test and ship.

lamont-granquist:
lets see if i can come up with a compromise...

lamont-granquist:
okay this removes simplecov as a dep but lets users add it optionally via GEMFILE_MOD or Gemfile.local

lamont-granquist:
The existence of `chef verify` as a user API makes that somewhat difficult to do, and I don't see how to walk that one back...

coderanger:
Either remove the command or make it clear that it downloads gems on the fly as needed. I don't really see why "check that the packages work" is a thing we ship to users, that's our job as packagers.

lamont-granquist:
I can get behind that, but we still need some machinery to construct the Gemfile.lock without shipping the gems.

lamont-granquist:
`bundle install --deployment` and nuking the `vendor/bundle` might work.

we could even ship the `vendor/bundle` i suppose as well...

pretend its a rails app...

coderanger:
All just academic for now, getting 2.0 out the door takes priority :)

lamont-granquist:
Well I'm neck deep in literally all that shit now, so its like a line or two of code away from possibly being how we ship 2.0...

PR #1240, opened 2017-06-05
Adds ability to set auth methods for ssh transport

logicminds:
* Previously the user was not able to set the auth_method to be
    used by ssh.  So if a third party gem implemented a
    new net::ssh auth_method there was no way to use the gem
    because the auth_method could not be changed.  Additionally,
    NET::SSH is not able to read local ssh config files either.

  * This code fixes that by allowing the user to set an auth_method
    of their choosing.  Example: 'gssapi-with-mic' method
    http://www.rubydoc.info/gems/net-ssh-krb/0.4.0

  * As a side affect with using gssapi-with-mic anybody wanting
    kerberos ticket authentication will now have the ability to do
    so when specifying 'gssapi-with-mic' as the auth method.  Although
    you must also install the net-ssh-krb gem.

### DIFF
--- a/lib/kitchen/transport/ssh.rb
+++ b/lib/kitchen/transport/ssh.rb
@@ -408,6 +408,8 @@ module Kitchen
           max_wait_until_ready: data[:max_wait_until_ready],
           ssh_gateway: data[:ssh_gateway],
           ssh_gateway_username: data[:ssh_gateway_username],
+          auth_methods: data[:auth_methods] ||
+              %w(publickey hostbased password keyboard-interactive),
         }
 
         if data[:ssh_key] && !data[:password]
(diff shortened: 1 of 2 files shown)


logicminds:
I am putting this up for review.  I have not smoke tested this yet and was hoping someone could take a quick look to see if there is where I need to change the code.  I need this in order to perform kerberos ticket authentication.

gordonbondon:
For anyone still waiting for this functionality, looks like @logicminds has already implemented this in a form of kitchen plugin [here](https://github.com/nwops/kitchen-kerberos). Works like a charm. Thanks!

logicminds:
I totally forgot about this PR.  I have been using my plugin for months now.  Should we just close this PR?

PR #1243, opened 2017-06-29
Adds pre_create_command for running arbitrary commands

sean797:
This allows .kitchen.yml to contains the following:

```yaml
driver:
  pre_create_command: echo 'this runs before an instance is created'
```

Adding this to core since the discussion at https://github.com/test-kitchen/kitchen-docker/pull/249 kitchen-vagrant already has this functionally, but this adds it to all drivers.

### DIFF
--- a/spec/kitchen/driver/base_spec.rb
+++ b/spec/kitchen/driver/base_spec.rb
@@ -88,6 +88,24 @@ describe Kitchen::Driver::Base do
     end
   end
 
+  describe ".pre_create_command" do
+    it "run command" do
+      Kitchen::Driver::Base.any_instance.stubs(:run_command).returns(true)
+
+      config[:pre_create_command] = "echo works 2&>1 > /dev/null"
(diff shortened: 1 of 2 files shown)


sean797:
Thanks for taking a look @coderanger. Updated, rebased and add some tests since tests are now passing 👍

spec/kitchen/driver/base_spec.rb
```diff
@@ -88,6 +88,18 @@ class Slow < Base
     end
   end
 
+  describe ".pre_create_command" do
+    it "run command" do
+      config[:pre_create_command] = "echo works 2&>1 > /dev/null"
```
coderanger: This will fail on Windows. You can use `driver.expects(:run_command).with("echo works 2&>1 > /dev/null")` to avoid actually running anything. Similarly with the fail test, that one is more technically okay but just for unit test speed it would be nice to use Mocha to stub it raising the exception rather than actually running a failed command :)

sean797:
Ugh Windows... Anyway updated and test are 🍏

PR #1249, opened 2017-07-21
Add missing parameter "data" to method def and call

AntonOfTheWoods:
The variable data is referred to but out of scope, adding as a
parameter.

### DIFF
--- a/lib/kitchen/provisioner/chef_base.rb
+++ b/lib/kitchen/provisioner/chef_base.rb
@@ -415,13 +415,13 @@ module Kitchen
           file.write(format_config_file(data))
         end
 
-        prepare_config_idempotency_check if config[:enforce_idempotency]
+        prepare_config_idempotency_check(data) if config[:enforce_idempotency]
       end
 
       # Writes a configuration file to the sandbox directory
       # to check for idempotency of the run.
       # @api private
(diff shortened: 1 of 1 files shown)


DPD-30:
I just submitted pr 1252 with the exact same change.  You beat me to it Nice work.  Now to get one of them merged.

AntonOfTheWoods:
<rant> Yep, the problem is that the wonderfully complete and super tests don't work. I must confess the whole Chef (ruby?) ecosystem seems to have this really pleasing tendency of having tests work for about 3 days and then stop working, so when you want to fix really basic bugs you are subjected to 4 hours of fixing rubbish that has nothing to do with your fix...</rant> I am super busy until middle of next week but will try and put aside some time I would have spent relaxing to fix this.

coderanger:
@AntonOfTheWoods Test Kitchen is maintained entirely through volunteer effort and most of that has been going to docs and triage lately. We welcome any help in improving maintenance.

gordonbondon:
Looks like checks here were failing not because of changes made in this PR. Tests were fixed since then. 
Can build be restarted? It would be really nice to have this fix merged and released.

MarkGibbons:
This was fixed by #1282

PR #1254, opened 2017-07-31
use a more efficient sort for gem specs

robbkidd:
Rubocop barked about an inefficient sort. Use sort_by instead. OK.


thommay:
you almost certainly don't want to run chefstyle in appveyor too, the line endings stuff is endless sadness

robbkidd:
@thommay I've told chefrubostylecop that line-endings should be `linefeed` regardless of platform. Let's try again!

PR #1255, opened 2017-07-31
reduce warnings during test runs

robbkidd:
The first in a series of efforts to clean up the tests. I recommend walking the commits to review each in isolation.

### DIFF
--- a/lib/kitchen/configurable.rb
+++ b/lib/kitchen/configurable.rb
@@ -390,8 +390,8 @@ module Kitchen
       def diagnose
         {
           class: name,
-          version: @plugin_version,
-          api_version: @api_version,
+          version: @plugin_version ||= nil,
+          api_version: @api_version ||= nil,
         }
       end
 
--- a/lib/kitchen/config.rb
+++ b/lib/kitchen/config.rb
@@ -56,11 +56,6 @@ module Kitchen
     # @api private
     attr_reader :log_root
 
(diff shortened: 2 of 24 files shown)


lib/kitchen/configurable.rb
```diff
@@ -390,8 +390,8 @@ def plugin_version(version) # rubocop:disable Style/TrivialAccessors
       def diagnose
         {
           class: name,
-          version: @plugin_version,
-          api_version: @api_version,
+          version: @plugin_version ||= nil,
```
thommay: do these get used in other places? Or can this just be
```
version: @plugin_version || nil,
```

robbkidd: They get used other places. They aren't initialized in a traditional manner (i.e. in an `initialize()` method) and usage of them in some places makes a modern Ruby interpreter bark about uninitialized variables. I could not determine a good place to initialize them because they're supposed to be set specific to a kitchen plugin via the DSL.

robbkidd: Addendums:

+ `@plugin_version || nil` won't prevent the uninitialized variable warning. `||=` does.
+ In all circumstances, the uninitialized instance variables were being used as though it was expected to be initialized and be falsey to trigger alternate behavior.

thommay: thanks for the clarification. Approved!

Issue #1238, opened 2017-05-24
kitchen converge Failed to complete #converge action: [no implicit conversion of nil into String]


dvenrao:
## Description

When I run kitchen create no issues. after this when i run kitchen converge it always giving error no implicit conversion of nil into String. Using test-kitchen via ChefDK

## Kitchen Version
```ruby
Test Kitchen version 1.16.0
```

## ChefDK Version
```ruby
Chef Development Kit Version: 1.2.22
chef-client version: 12.18.31
delivery version: master (0b746cafed65a9ea1a79de3cc546e7922de9187c)
berks version: 2017-05-24T09:45:44.801656 20928] 2017-05-24T09:45:44.802156 20928] 2017-05-24T09:45:44.802656 20928] 2017-05-24T09:45:44.802656 20928] 2017-05-24T09:45:44.918679 20928] 2017-05-24T09:45:44.918679 20928] 5.6.0
kitchen version: 1.16.0
```

## Platform Version
Windows 7

## Kitchen Output

```ruby
> kitchen converge default-centos65 -l debug
-----> Starting Kitchen (v1.16.0)
D      [Vagrant command] BEGIN (vagrant --version)
D      [Vagrant command] END (0m0.00s)
D      Berksfile found at C:/Users/vdonthin/code/chef-repo/cookbooks/enterprisechef/Berksfile, loading Berkshelf
D      Berkshelf 6.0.1 library loaded
-----> Converging <default-centos65>...
       Preparing files for transfer
D      Creating local sandbox in C:/Users/vdonthin/Local Settings/Temp/default-centos65-sandbox-20170524-19136-18jx331
       Preparing dna.json
D      Creating dna.json from {:run_list=>["recipe[apache::default]"]}
       Resolving cookbook dependencies with Berkshelf 6.0.1...
D      Using Berksfile from C:/Users/vdonthin/code/chef-repo/cookbooks/enterprisechef/Berksfile
D      Cleaning up local sandbox in C:/Users/vdonthin/Local Settings/Temp/default-centos65-sandbox-20170524-19136-18jx331
>>>>>> ------Exception-------
>>>>>> Class: Kitchen::ActionFailed
>>>>>> Message: 1 actions failed.
>>>>>>     Failed to complete #converge action: [no implicit conversion of nil into String] on default-centos65
>>>>>> ----------------------
>>>>>> Please see .kitchen/logs/kitchen.log for more details
[...]

PR #1258, opened 2017-08-09
Fix Dir.glob usage

jaym:
Dir.glob is broken on Windows in a variety of ways. Changing to the
directory you want to glob before doing any sort of globbing is the safest
way to use it from past experience.

Related to https://github.com/test-kitchen/test-kitchen/issues/1238 in
that it is the very next error one would run into with berkshelf being
fixed.

### DIFF
--- a/lib/kitchen/util.rb
+++ b/lib/kitchen/util.rb
@@ -143,7 +143,7 @@ module Kitchen
     end
 
     # Lists the contents of the given directory. path will be prepended
-    # to the list returned
+    # to the list returned. '.' and '..' are never returned.
     #
     # @param path [String] the directory to list
     # @param include_dot [Boolean] if true, dot files will be included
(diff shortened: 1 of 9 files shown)


robbkidd:
Any ideas around how to add tests for these new `Util` methods?

jaym:
yep, adding

lib/kitchen/util.rb
```diff
@@ -141,5 +141,59 @@ def self.shell_helpers
...
+                         "**/*"
+                       else
+                         "*"
+                       end
+        flags = if include_dot
+                  [File::FNM_DOTMATCH]
+                else
+                  []
+                end
+        Dir.glob(glob_pattern, *flags)
```
robbkidd: Would it make sense to use `safe_glob` here instead of re-doing `Dir.chdir { Dir.glob }`? Or would that make it harder to filter out `.` and `..`?

robbkidd: Yes. After some tinkering, I see that it does. Nevermind.

jaym:
@robbkidd any other thoughts?

robbkidd:
> any other thoughts?

Nope. Brain is empty.

PR #1262, opened 2017-08-22
new install_strategy option used in conjunction with product_name

wrightp:
This PR is part of a series of new features being added to Kitchen based on [RFC-091](https://github.com/chef/chef-rfc/blob/master/rfc091-deprecate-kitchen-settings.md#new-settings)

### New provisioner option `install_strategy`
Only enabled when used with `product_name` setting.

| Values | Description |
| ------- | ----------- |
| once (default) | when using product_name Kitchen will no longer reinstall chef/chefdk if it's already installed just like the default require_chef_omnibus behavior |
| always | always install the specified product_version |
| skip | do not install chef/chefdk|

### DIFF
--- a/lib/kitchen/provisioner/chef_base.rb
+++ b/lib/kitchen/provisioner/chef_base.rb
@@ -106,6 +106,27 @@ module Kitchen
       end
       expand_path_for :encrypted_data_bag_secret_key_path
 
+      #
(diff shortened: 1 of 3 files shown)


wrightp:
@schisamo @cheeseplus @robbkidd @coderanger 
cc/ @rhass @scotthain @tduffield

PR #1267, opened 2017-08-31
add download_url and checksum provisioner options

wrightp:
This PR is part of a series of new features being added to Kitchen based on [RFC-091](https://github.com/chef/chef-rfc/blob/master/rfc091-deprecate-kitchen-settings.md#new-settings)

### New provisioner options `download_url`, `checksum`
Currently only enabled when used with `product_name` setting.

`download_url` is meant to replace `install_msi_url`.

bourne and powershell environments can now download a package from a URL for install.
`checksum` is optional when used with `download_url`.

For those curious how this works checkout https://github.com/chef/mixlib-install/pull/237

### DIFF
--- a/lib/kitchen/provisioner/chef_base.rb
+++ b/lib/kitchen/provisioner/chef_base.rb
@@ -127,6 +127,10 @@ module Kitchen
 
       default_config :architecture
 
+      default_config :download_url
+
+      default_config :checksum
+
       # Reads the local Chef::Config object (if present).  We do this because
       # we want to start bring Chef config and ChefDK tool config closer
       # together.  For example, we want to configure proxy settings in 1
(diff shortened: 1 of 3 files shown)


lib/kitchen/provisioner/chef_base.rb
```diff
@@ -127,6 +127,10 @@ class ChefBase < Base
 
       default_config :architecture
 
+      default_config :download_url
+
+      default_config :checksum
```
rhass: Is any checksum type supported?

wrightp: sha256 only since that's what the install scripts validate. This will be documented.

wrightp:
cc/ @cheeseplus @robbkidd @thommay @schisamo

PR #1281, opened 2017-09-22
Support renamed net-ssh option `verify_host_key`

### LINKED ISSUE #524: a way to override ~/.kitchen/config.yml
I could use a way to override `~/.kitchen/config.yml` inside the cookbook.  Currently, the order is

> `~/.kitchen/config.yml` overrides `.kitchen.local.yml` overrides `.kitchen.yml`

In a green field, I'd make that

> `.kitchen.local.yml` overrides `~/.kitchen/config.yml` overrides `.kitchen.yml`

but maybe another level needs to be added to avoid breaking backward compatibility.

The use case is to set the driver configuration in `~/.kitchen/config.yml`, but allow overriding it per-cookbook, because some cookbooks don't work with some drivers.

(similar to #343, but different)

cheeseplus:
Signed-off-by: Seth Thomas <sthomas@chef.io>

### DIFF
--- a/lib/kitchen/driver/ssh_base.rb
+++ b/lib/kitchen/driver/ssh_base.rb
@@ -239,7 +239,7 @@ module Kitchen
 
         opts = {}
         opts[:user_known_hosts_file] = "/dev/null"
-        opts[:paranoid] = false
+        opts[:verify_host_key] = false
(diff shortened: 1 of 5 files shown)


thommay:
as tim pointed out on chef-client you'll also want to dep on net-ssh >= 4.2

PR #1280, opened 2017-09-21
Release 1.18

tas50:
Signed-off-by: Tim Smith <tsmith@chef.io>

PR #1252, opened 2017-07-27
enforce_idempotency missing "data" error

DPD-30:
When the kitchen.yml include
provisioner:
name: chef_zero
enforce_idempotency: true
multiple_converge: 2

Error such as the following is received.

Failed to complete #converge action: [undefined local variable or method `data' for #.........
Pull request 1068 has this fix lumped in with other proxy items. test-kitchen#1068
I am breaking this out into separate PR with just the fix for enforce_idempotency.

Other references to this error can be found in bottom of pull 875 after it was merged.
test-kitchen#875

PR #1282, opened 2017-09-22
idempotent_check: Allow specificaton of enforce_idempotency

MarkGibbons:
Currently specifying enforce_idmempotency as an option will cause an error, data is undefined.
Fix the code and add tests to detect the condition and that a client rb file with the handler
that checks for idempotency is generated.

A rebase of #1252 was requested.  This does the rebase and includes tests.  This PR should replace #1280 and part of #1068.

Thanks,
  Mark

### DIFF
--- a/lib/kitchen/provisioner/chef_base.rb
+++ b/lib/kitchen/provisioner/chef_base.rb
@@ -460,13 +460,13 @@ module Kitchen
           file.write(format_config_file(data))
         end
 
-        prepare_config_idempotency_check if config[:enforce_idempotency]
+        prepare_config_idempotency_check(data) if config[:enforce_idempotency]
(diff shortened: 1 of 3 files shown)


cheeseplus:
The build break is entirely unrelated but I want to get it green before we merge mostly for peace of mind. Once that is sorted, I'll be merging this one.

MarkGibbons:
Anything you want me to do?

cheeseplus:
The build break is unrelated to any of your changes @MarkGibbons so nothing to do right now - once I resolve that then we'll see if anything further needs doing.

cheeseplus:
Rebase this one against master and we should be good to go @MarkGibbons

PR #1283, opened 2017-09-22
Pin to net-ssh 4.1.0 for now

cheeseplus:
Due to an upstream language change in https://github.com/net-ssh/net-ssh/pull/524 we are pinning to `<= 4.1.0` like https://github.com/chef/chef-dk/pull/1406 until we can update related projects like `train`. This will fix the DK while we update the world and if we can contribute an upstream fix as the code should only warn, not break.

Signed-off-by: Seth Thomas <sthomas@chef.io>

### DIFF
--- a/.kitchen.ci.yml
+++ b/.kitchen.ci.yml
@@ -11,7 +11,6 @@ provisioner:
   name: chef_zero
 
 platforms:
-  - name: ubuntu-14.04
   - name: windows-2012R2
     transport:
       name: winrm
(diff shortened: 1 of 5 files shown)


cheeseplus:
Given that we're now getting rather odd errors with travis's ENV and Proxy, it's a good time to update the tests to resemble something both modern and actually used. Related, we need to finish the work started here https://github.com/test-kitchen/test-kitchen/pull/1220 because proxy is still using legacy code.

- proxy -> dokken
- simplify and update travis settings
-

PR #1286, opened 2017-09-28
reset_command doesn't actually need to be required

coderanger:
Not sure why we made it required, but might as well fix that.


cheeseplus:
seems legit

PR #1289, opened 2017-10-01
Update CentOS 7 / Ubuntu to the latest versions

tas50:
12.04/14.04 -> 16.04
7.2 -> 7 (bento slug)


tas50:
Thanks for fixing this @cheeseplus

PR #1290, opened 2017-10-03
Remove the `driver create` and `driver discover` commands

coderanger:
Also `driver help` along with it. No one should be using these and they aren't on any core workflow so I'm arbitrarily declaring this not a compat break.


coderanger:
Travis failure is pending https://github.com/chef/proxy_tests/pull/18

PR #1291, opened 2017-10-03
improve case handling for ssh auth

p4ndepravity:
also provide custom auth method capabilities on yml

### DIFF
--- a/lib/kitchen/transport/ssh.rb
+++ b/lib/kitchen/transport/ssh.rb
@@ -415,23 +415,22 @@ module Kitchen
           ssh_gateway_username: data[:ssh_gateway_username],
         }
 
-        opts[:auth_methods] = Array(data[:auth_methods])  if data.key?(:auth_methods)
-        opts[:auth_methods] ||= ["publickey"]             if data[:ssh_key]
-        opts[:keys] = Array(data[:ssh_key])               if data[:ssh_key]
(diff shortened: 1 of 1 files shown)


lib/kitchen/transport/ssh.rb
```diff
@@ -415,23 +415,28 @@ def connection_options(data)
...
         end
 
-        opts[:password] = data[:password]           if data.key?(:password)
         opts[:forward_agent] = data[:forward_agent] if data.key?(:forward_agent)
         opts[:verbose] = data[:verbose].to_sym      if data.key?(:verbose)
 
         # disable host key verification. The hash key to use
         # depends on the version of net-ssh in use.
-        opts[verify_host_key_option] = false
+        opts[:verify_host_key_option] = false
```
coderanger: That was a method, not a symbol.

coderanger:
Can you be more specific about your goal here?

p4ndepravity:
I have been unable to successfully authenticate using ssl for the ssh connection so I needed to provide a means for using password authentication. also there was a situation where an ssh key was provided (by kitchen by default) as well as a password and the if data[:ssh_key] && !data[:password] block wasn't being executed.

cheeseplus:
As it stands this is broken and I'm still unclear what problem this is solving for so closing this one but happy to review another PR with a clearer description of the problem statement and how this solves it.

PR #1293, opened 2017-10-05
Upgrades to the shell provisioner

coderanger:
Remove the concept of a default script from the shell provisioner, add support for array arguments, and add overrides for both commands if you want to manage things directly in the kitchen config.

### DIFF
--- a/lib/kitchen/provisioner/shell.rb
+++ b/lib/kitchen/provisioner/shell.rb
@@ -37,12 +37,9 @@ module Kitchen
       end
       expand_path_for :script
 
-      # Override for the initialization command.
-      default_config :init_command, nil
-      # Override for the converge command.
-      default_config :run_command, nil
-      # Special override to set run_command and disable init_command.
+      # Run a single command instead of managing and running a script.
       default_config :command, nil
+
       # Add extra arguments to the converge script.
       default_config :arguments, []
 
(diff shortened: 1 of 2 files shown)


lib/kitchen/provisioner/shell.rb
```diff
@@ -35,11 +37,38 @@ class Shell < Base
       end
       expand_path_for :script
 
+      # Override for the initialization command.
+      default_config :init_command, nil
+      # Override for the converge command.
+      default_config :run_command, nil
+      # Special override to set run_command and disable init_command.
```
thommay: What's the value of adding `run_command` here? Given `init_command` defaults to nil, why not just support `command` and `init_command`?

coderanger: By default both `run` and `init` have their magic-y auto-defined values to facilitate the `bootstrap.sh` mode. If you set either, it just overrides that phase. If you set `command` it is effectively disabling the `bootstrap.sh` mode and saying "just run this" which I expect is what most people want, but having the selective overrides seemed like a nice thing for the future.

thommay: my YAGNI alert is going off here, honestly.

coderanger: I kind of agree, but the mode I really care about is the non-bootstrap.sh stuffs and I don't want to damage backwards compat so this gives more flexibility to anyone using the existing mode that needs an escape hatch.

coderanger: Basically I think we have two options. Either leave all three command options, or shrink to just `command` and leave the `bootstrap.sh` mode as its own thing. You can already customize the script arguments so maybe that's enough for that mode as is. Opinions @cheeseplus?

cheeseplus: I think most people don't like bootstrap.sh and tolerate it cause it was all that was there for so long so probably moving to `command` and leaving `bootstrap.sh' as it's own thing seems best right now.

coderanger:
Okay, trimmed down the options. Setting `command` will run just that command.

PR #1295, opened 2017-10-05
Exec driver

coderanger:
As discussed at the summit, a driver (and transport) for running everything locally.

Also we should probably use this pattern over in dokken too, for making the driver force the use of a specific transport (or other plugin). It's weird but it seems to work.

### DIFF
--- a/lib/kitchen/driver/exec.rb
+++ b/lib/kitchen/driver/exec.rb
@@ -13,21 +13,27 @@
 # limitations under the License.
 #
 
-require "kitchen"
-require "kitchen/version"
+require "kitchen/driver/base"
+require "kitchen/shell_out"
 require "kitchen/transport/exec"
+require "kitchen/version"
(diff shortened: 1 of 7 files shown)


lib/kitchen/driver/exec.rb
```diff
@@ -0,0 +1,68 @@
...
+    class Exec < Kitchen::Driver::Base
+      include ShellOut
+
+      plugin_version Kitchen::VERSION
+
+      default_config :reset_command, nil
+
+      no_parallel_for :create, :converge, :destroy
+
+      # Hack to inject to Exec transport in too.
```
thommay: i'm not sure what these words mean in this order :)

coderanger: "Also inject the transport when using the driver" will change since it took me a minute to parse/remember :)

PR #1296, opened 2017-10-10
Add a --debug command line option

coderanger:
This option works with the `converge`, `verify`, and `test` command and enables plugin-specific debugging behavior. I've added support in the Chef provisioners to change the log level of Chef to debug.

The goal is here to make it Very Easy™ to get debugging data when iterating with Test Kitchen. Like `kitchen converge -D`, rather than having to edit the `.kitchen.yml` each time or use fancy Erb tricks.

### DIFF
--- a/lib/kitchen/cli.rb
+++ b/lib/kitchen/cli.rb
@@ -180,6 +180,13 @@ module Kitchen
           [Future DEPRECATION, use --concurrency]
           Run a #{action} against all matching instances concurrently.
         DESC
+      if action == :converge || action == :verify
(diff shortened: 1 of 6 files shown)


drrk:
👍 Thanks for this Noah, when I made the change to split the kitchen debugging from the provisioner debugging I couldn't get my head round how to add the command line options for it.

coderanger:
@drrk Want to make that an official code review? :)

drrk:
No, because I still don't understand how it works :(

PR #1298, opened 2017-10-10
install after init and transfer

gtmanfred:
This allows for adding an artifact to the sandbox, and then having it
uploaded when the sandbox is transferred.  That artifact can then be
used with the installation method.

Used here https://github.com/saltstack/kitchen-salt/pull/164/commits/d9048deba6822d64687a06ac55b53e7aa9e6c151

Thanks,
Daniel


gtmanfred:
Nevermind, looks like install_command should actually be run command in kitchen-salt

PR #1303, opened 2017-10-20
Configuration Deprecation warnings

wrightp:
This PR is intended to lead a discussion for how to warn users of the upcoming deprecated options (part of [RFC-091](https://github.com/chef/chef-rfc/blob/master/rfc091-deprecate-kitchen-settings.md)). 

This approach is piggy-backing off of the new `kitchen doctor` command.  The change only includes the initial warning for detected deprecation attributes.

### DIFF
--- a/lib/kitchen/provisioner/chef_base.rb
+++ b/lib/kitchen/provisioner/chef_base.rb
@@ -133,6 +133,27 @@ module Kitchen
 
       default_config :checksum
 
+      # Deprecate config attributes
+      deprecate_config_for :require_chef_omnibus, Util.outdent!(<<-MSG)
+        Message here!
+      MSG
+
+      deprecate_config_for :chef_omnibus_url, Util.outdent!(<<-MSG)
+        Message here!
+      MSG
+
+      deprecate_config_for :chef_omnibus_install_options, Util.outdent!(<<-MSG)
+        Message here!
+      MSG
+
+      deprecate_config_for :install_msi_url, Util.outdent!(<<-MSG)
+        Message here!
+      MSG
+
+      deprecate_config_for :chef_metadata_url, Util.outdent!(<<-MSG)
+        Message here!
+      MSG
+
       # Reads the local Chef::Config object (if present).  We do this because
       # we want to start bring Chef config and ChefDK tool config closer
       # together.  For example, we want to configure proxy settings in 1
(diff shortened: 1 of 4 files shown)


wrightp:
I don't have privs to add reviewers...
 @coderanger  @cheeseplus  @thommay

coderanger:
So we probably want some kind of block to check the current value, as the plan is/was to have a migration period where we start with only warning on certain types of `require_chef_omnibus` values. I would probably make deprecate_config_for more like default_config where it takes either a value or a block.

wrightp:
@coderanger gotcha. My first attempt was more like you just described so I'll show how that looks today. Something like:
```ruby
# I'm guessing on the some of the details here...
deprecate_config_for :require_chef_omnibus do |provisioner|
  if provisioner[:require_chef_omnibus] != true
    warn("message") # or set some message collection
  end
end
```

coderanger:
If you want to stick to the block returning a string rather than handling output itself that makes things a bit more controllable, but also less flexible. I would err on the side of structured data for this though, since `doctor` already covers the unstructured case :)

lib/kitchen/provisioner/chef_base.rb
```diff
@@ -133,6 +133,47 @@ class ChefBase < Base
 
       default_config :checksum
 
+      # Deprecate config attributes
+      deprecate_config_for :require_chef_omnibus do |provisioner|
+        if provisioner[:require_chef_omnibus] != defaults[:require_chef_omnibus]
```
wrightp: Note that this condition is just an example for how the blocks will return messages. Value of `false` will describe how to use the `install_strategy` attr and a version value will describe how to apply that value to the `product_version` attr.

wrightp:
@coderanger Do you think this is getting colder or warmer? Not sure if you have anything specific in mind for overriding the doctor method. I'm still in the experimental phase.

coderanger:
Seems to be in the right direction. You can collapse the simple class of "if this key is present in the config, show the message" so you don't need to use a block, since that's the most common case. Otherwise looking good.

wrightp:
@coderanger @cheeseplus Finally getting back to this. Any additional input before we feel this can be merged?

cheeseplus:
I don't have anything more, unless @coderanger does I say merge.

coderanger:
The `once` strategy is already released upstream, right? I think that happened like a year ago but I lost track :)

wrightp:
`once` strategy is good to go.

PR #1306, opened 2017-10-24
adding download support to the base transport and provisioner

atheiman:
I'd like test-kitchen to be able to download files from instances post-converge by specifying them in the kitchen yaml provisioner section:

```yaml
# .kitchen.yml

provisioner:
  downloads:
    /tmp/kitchen/nodes: test/fixtures
    /tmp/kitchen/data_bags: test/fixtures
```

This config would download all of chef-zero's saved nodes and data bags back to `test/fixtures/nodes/node-name.json` and `test/fixtures/data_bags/bag-name/item-name.json`. These changes support:

- downloading a file from a remote path to a local path
- downloading a directory recursively from a remote path to a local path
- specifying the name of the downloaded file / directory or keeping the original name by only specifying a parent directory to download to
- backwards compatible - I run the download against `config[:provisioner][:downloads]`, which defaults to an empty hash resulting in no downloading
- specifying multiple files / directories to be downloaded to the same path with a shorthand:
  ```yaml
  provisioner:
    downloads:
      [/tmp/kitchen/nodes, /tmp/kitchen/data_bags]: test/fixtures
  ```

----

My goal here stems from https://github.com/mwrock/kitchen-nodes/issues/36

Im using the [kitchen-nodes provisioner](https://github.com/mwrock/kitchen-nodes) to allow test-kitchen ec2 instances to know of each other, and load their node data into chef zero at the start of each chef-client. I want to extend that to be able to save updated nodes back to the nodes directory on my workstation after a converge on an instance. These updated nodes will then be copied up to the next instance that is converged.

I'd also like to be able to download data bags for the use case of a node registers itself in a data bag and another node does something to connect to that node using data in the data bag.

### DIFF
--- a/lib/kitchen/transport/winrm.rb
+++ b/lib/kitchen/transport/winrm.rb
@@ -139,11 +139,18 @@ module Kitchen
 
         # (see Base::Connection#download)
         def download(remotes, local)
-          # This looks possible with WinRM::FS::FileManager
-          raise(
-            NotImplementedError,
-            "Downloading files from instance not yet implemented for Windows remotes"
-          )
+          # ensure the parent dir of the local target exists
+          FileUtils.mkdir_p(File.dirname(local))
+
+          Array(remotes).each do |remote|
+            file_manager.download(remote, local)
+          end
+        end
+
+        # @return [Winrm::FileManager] a file transporter
+        # @api private
+        def file_manager
+          @file_manager ||= WinRM::FS::FileManager.new(session)
         end
 
         # (see Base::Connection#wait_until_ready)
--- a/lib/kitchen/transport/winrm.rb
+++ b/lib/kitchen/transport/winrm.rb
@@ -137,6 +137,15 @@ module Kitchen
           file_transporter.upload(locals, remote)
         end
 
+        # (see Base::Connection#download)
+        def download(remotes, local)
+          # This looks possible with WinRM::FS::FileManager
+          raise(
+            NotImplementedError,
+            'Downloading files from instance not yet implemented for Windows remotes'
+          )
+        end
+
         # (see Base::Connection#wait_until_ready)
         def wait_until_ready
           delay = 3
--- a/lib/kitchen/transport/winrm.rb
+++ b/lib/kitchen/transport/winrm.rb
@@ -142,7 +142,7 @@ module Kitchen
           # This looks possible with WinRM::FS::FileManager
           raise(
             NotImplementedError,
-            'Downloading files from instance not yet implemented for Windows remotes'
+            "Downloading files from instance not yet implemented for Windows remotes"
           )
         end
 
--- a/lib/kitchen/driver/ssh_base.rb
+++ b/lib/kitchen/driver/ssh_base.rb
@@ -80,6 +80,12 @@ module Kitchen
           debug("Transfer complete")
           conn.execute(env_cmd(provisioner.prepare_command))
           conn.execute(env_cmd(provisioner.run_command))
+          info("Downloading files from #{instance.to_str}")
+          provisioner[:downloads].to_h.each do |remotes, local|
+            debug("Downloading #{Array(remotes).join(', ')} to #{local}")
+            conn.download(remotes, local)
+          end
+          debug("Download complete")
         end
       rescue Kitchen::Transport::TransportFailed => ex
         raise ActionFailed, ex.message
--- a/lib/kitchen/provisioner/base.rb
+++ b/lib/kitchen/provisioner/base.rb
@@ -47,6 +47,8 @@ module Kitchen
 
       default_config :command_prefix, nil
 
+      default_config :downloads, {}
+
       expand_path_for :test_base_path
 
       # Constructs a new provisioner by providing a configuration hash.
--- a/lib/kitchen/provisioner/base.rb
+++ b/lib/kitchen/provisioner/base.rb
@@ -78,6 +78,12 @@ module Kitchen
             config[:max_retries],
             config[:wait_for_retry]
           )
+          info("Downloading files from #{instance.to_str}")
+          config[:downloads].to_h.each do |remotes, local|
+            debug("Downloading #{Array(remotes).join(', ')} to #{local}")
+            conn.download(remotes, local)
+          end
+          debug("Download complete")
         end
       rescue Kitchen::Transport::TransportFailed => ex
         raise ActionFailed, ex.message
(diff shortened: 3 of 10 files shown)


coderanger:
We've been basically putting this off as part of the multi-node madness. The issue with this approach in particular is it would have to wait for TK 2.0 as it would break compat with all existing transport plugins :-/ So it's a nice idea but I don't think we can easily move forward at this time without much more complex fallback code to deal with transports that don't support it (and honestly we would probably do that for 2.0 too because of how slowly things move and wanting good UX).

coderanger:
So overall right idea but will need to be super duper bullet proof for backwards compat :(

atheiman:
:(

atheiman:
:( alright ill wait for this to come in 2.0

coderanger:
So as I'm reading this, I think I over-blew your changes. As long as we default to not having anything in the downloads array we can leave the subsystem dormant so that makes it easier to add in a compat-y way. Wouldn't be as automagic as the final version might be but we can definitely put in the plumbing. It's just turning on "fetch down the node JSON" by default that iwll have to wait.

atheiman:
yea i was planning to make this opt-in, something like this:

```yaml
# .kitchen.yml

provisioner:
  name: nodes
  downloads:
    /tmp/kitchen/nodes: test/fixtures
    /tmp/kitchen/data_bags: test/fixtures
```

would that be acceptable?

atheiman:
ive updated the description to try to clarify what Im doing here better

@coderanger @tas50 @thommay @cheeseplus can you kind folks review this?

this is my first time contributing to this repo, and my first time using minitest, so I'm sure there are some things you'll want changed.

atheiman:
Trying this on real kitchen nodes for the first time. It looks like this works as expected, Im able to download files from the vagrant instances post converge.

some notes:
- `/tmp/kitchen/data_bags/` is not executable (even tho `/tmp/kitchen/nodes/` is?) so it cannot be downloaded (the scp download runs as the vagrant user):
  ```
  [vagrant@default-centos-72 ~]$ ls -alh /tmp/kitchen | grep -e nodes -e data_bags
  drwx------.  3 root    root      22 Oct 27 03:59 data_bags
  drwxr-xr-x.  2 vagrant vagrant   66 Oct 27 03:59 nodes
  ```
  This can be worked around with a `directory` resource in a fixture recipe:
  ```ruby
  directory '/tmp/kitchen/data_bags' do
    recursive true
    mode '0755'
    action :create
  end
  ```
  I dont know of any way to force the download to be done by the root user. But really this kinda seems like a bug with test-kitchen or chef-zero for the nodes path to have a different mode than the data_bags path?
- the node json files that are being downloaded are not updated with the full node attributes. I think this must be due to downloading _before_ chef-zero shuts down and saves off the nodes?

atheiman:
can i get this reviewed / merged? I want to add this to the kitchen-sync sftp transport next.

coderanger:
You were working on it for a while after making the PR so you need to tell us when you’re actually ready for a review again :)

atheiman:
yea understandable, and i know y'all are busy
ready for review now tho

lib/kitchen/version.rb
```diff
@@ -17,5 +17,5 @@
 # limitations under the License.
 
 module Kitchen
-  VERSION = "1.18.0".freeze
+  VERSION = "1.19.0".freeze
```
thommay: you'll want to remove this.

lib/kitchen/transport/winrm.rb
```diff
@@ -137,6 +137,15 @@ def upload(locals, remote)
           file_transporter.upload(locals, remote)
         end
 
+        # (see Base::Connection#download)
+        def download(remotes, local)
+          # This looks possible with WinRM::FS::FileManager
```
thommay: you could probably do this reasonably trivially - https://github.com/chef/train/commit/8fb6a6bc71f26c8cbaacf977a55ff827fca25240#diff-f2edeefbb087259b847aa6f5ded7d0f4 is the corresponding work in train.

lib/kitchen/version.rb
atheiman: ok, rebased on master and now its showing no changes. is there a bot that increments it or something?

lib/kitchen/transport/winrm.rb
atheiman: I added the download and file_manager methods from there.

atheiman:
appveyor is stuck waiting for input

```
gem install bundler --quiet --no-ri --no-rdoc || gem install bundler --quiet --no-ri --no-rdoc || gem install bundler --quiet --no-ri --no-rdoc
bundler's executable "bundle" conflicts with C:/Ruby23/bin/bundle
Overwrite the executable? [yN]  
```

https://github.com/test-kitchen/test-kitchen/blob/master/appveyor.yml#L33

I dont easily see a way to pass yes to this, any ideas?

atheiman:
maybe someone can kick off another round of tests for appveyor because it looks like this has not been a problem in other prs or in my earlier commits on this pr

atheiman:
@coderanger i dont know what to do about the appveyor builds suddenly starting to fail :|

any ideas on fixing it? I dont have powershell available to me to really play with this. should I roll back to leave Windows as not implemented? I dont know how that would affect it

cheeseplus:
The AppVeyor failures are changes they've made on their end which we are hitting on several repos, I'll be trying to get this sorted on kitchen repos as well.

atheiman:
@cheeseplus ok ill delete commits i had trying to get around it

atheiman:
any updates on when this can be merged?

atheiman:
@thommay do you know when this will be released?

gtmanfred:
@atheiman i am seeing a problem with winrm's download.

I am getting 

```
>>>>>> ------Exception-------
>>>>>> Message: 1 actions failed.
>>>>>>     Failed to complete #verify action: [undefined local variable or method `session' for #<Kitchen::Transport::Winrm::Connection:0x007ffae5aaa658>] on py2-windows-2016
>>>>>> ----------------------
```

Which corresponds to here https://github.com/gtmanfred/kitchen-salt/blob/master/lib/kitchen/verifier/runtests.rb#L57

https://gist.github.com/gtmanfred/f88a1eb209b3070fd3cbbf4d3e88afdc

You can replicate it by running `py2-windows-2016` from the `kitchen` branch of https://github.com/gtmanfred/salt

This is all working fine for ssh, through the rsync transport in kitchen-sync.

I would open an issue, but github is throwing an error on the issue for this project for me right now :/

Thanks!
Daniel

gtmanfred:
looks like it should have been connection

PR #1310, opened 2017-11-03
Turn auto-retries on by default for Chef provisioners

coderanger:
I think that's the right codes from the RFC, but easy to change.


coderanger:
Roger, I'll leave 213 in there for some future when Chef can self-upgrade even though testing that should probably not happen in TK.

coderanger:
Ignoring appveryor because it looks jammed on something.

PR #1312, opened 2017-11-09
[MSYS-703] Fix code to validate retry_on_exit_code

NAshwini:
**Issue reported by the user**:
In kitchen.yml customer was sending the following attributes :
```
 retry_on_exit_code: 
   - [35, 1, 20]
  max_retries: 3
  wait_for_retry: 120 
```
By executing command `kitchen diagnose --all`, found that attribute was sent in 2D Array format as:
 ```    
retry_on_exit_code:
        -  -35
        - 1
        - 20
        max_retries: 3
        wait_for_retry: 120 
```
command ` retryable_exit_codes.include?(exit_code)` was not sufficient to check the 2D Array format (i.e [[35, 1, 20]] ) .
Hence was not able to retry kitchen execution after getting exit code `35` & customer was not able to understand why it is not retrying.

This PR validates & handles such type of issues.

### DIFF
--- a/lib/kitchen/transport/base.rb
+++ b/lib/kitchen/transport/base.rb
@@ -138,20 +138,9 @@ module Kitchen
         end
 
         def retry?(current_try, max_retries, retryable_exit_codes, exit_code)
-          retryable_exit_codes_exists = false
-          if !retryable_exit_codes.nil?
-            if retryable_exit_codes.class == String
-              retryable_exit_codes_exists = retryable_exit_codes.include?(exit_code.to_s)
-            elsif retryable_exit_codes[0].class == Array && retryable_exit_codes.class == Array
-              retryable_exit_codes_exists = !retryable_exit_codes.select{|x| x.include?(exit_code)}.empty?
-            elsif retryable_exit_codes[0].class == Integer && retryable_exit_codes.class == Array
-              retryable_exit_codes_exists = retryable_exit_codes.include?(exit_code)
-            else
-              raise TypeError, "Invalid type of retry_on_exit_code"
-            end
-          end
-
-         current_try <= max_retries && retryable_exit_codes_exists
+          current_try <= max_retries &&
+            !retryable_exit_codes.nil? &&
+            retryable_exit_codes.flatten.include?(exit_code)
         end
 
         # Builds a LoginCommand which can be used to open an interactive
--- a/lib/kitchen/transport/base.rb
+++ b/lib/kitchen/transport/base.rb
@@ -138,9 +138,20 @@ module Kitchen
         end
 
         def retry?(current_try, max_retries, retryable_exit_codes, exit_code)
-          current_try <= max_retries &&
-            !retryable_exit_codes.nil? &&
-            retryable_exit_codes.include?(exit_code)
+          retryable_exit_codes_exists = false
+          if !retryable_exit_codes.nil?
+            if retryable_exit_codes.class == String
+              retryable_exit_codes_exists = retryable_exit_codes.include?(exit_code.to_s)
+            elsif retryable_exit_codes[0].class == Array && retryable_exit_codes.class == Array
+              retryable_exit_codes_exists = !retryable_exit_codes.select{|x| x.include?(exit_code)}.empty?
+            elsif retryable_exit_codes[0].class == Integer && retryable_exit_codes.class == Array
+              retryable_exit_codes_exists = retryable_exit_codes.include?(exit_code)
+            else
+              raise TypeError, "Invalid type of retry_on_exit_code"
+            end
+          end
+
+         current_try <= max_retries && retryable_exit_codes_exists
         end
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by coderanger** [comment]
Please provide more information about this change.

File: lib/kitchen/transport/base.rb [hunk 1]
```diff
@@ -138,9 +138,20 @@ def execute_with_retry(command, retryable_exit_codes = [], max_retries = 1, wait
         end
 
         def retry?(current_try, max_retries, retryable_exit_codes, exit_code)
-          current_try <= max_retries &&
-            !retryable_exit_codes.nil? &&
-            retryable_exit_codes.include?(exit_code)
+          retryable_exit_codes_exists = false
```
coderanger: What you are probably looking for is `Array(retryable_exit_codes).flatten.map(&:to_i).include?(exit_code)`

NAshwini:
Appveyor is getting failed due to Build execution time has reached the maximum allowed time.

coderanger:
@NAshwini That is expected, we're working on it, sorry for the trouble :) Nothing in this is Windows-specific so we can just ignore it.

lib/kitchen/transport/base.rb
```diff
@@ -138,9 +138,12 @@ def execute_with_retry(command, retryable_exit_codes = [], max_retries = 1, wait
         end
 
         def retry?(current_try, max_retries, retryable_exit_codes, exit_code)
-          current_try <= max_retries &&
-            !retryable_exit_codes.nil? &&
-            retryable_exit_codes.include?(exit_code)
+          if !retryable_exit_codes.nil? && retryable_exit_codes.class == Array || retryable_exit_codes.class == Integer
```
coderanger: Explicitly checking classes like this leads to fragile code. To `to_i` will already convert things like `"1"` so if the concern is string like `"1,2"` we should either check for that explicitly or just support it.

```ruby
Array(retryable_exit_codes).flatten.flat_map{|v| v.is_a?(String) ? v.split(/,/) : v}.map(&:to_i).include?(exit_code)
```

coderanger:
Also just to be clear, why are we adding this? With the recent #1310 I don't expect many people to ever touch this setting so this can be conderied an advanced option which doesn't need as much UX smoothing.

lib/kitchen/transport/base.rb
NAshwini: Hi @coderanger, If we send values in kitchen.yml file without space as 
```
retry_on_exit_code: 
   -35
   -1 
```
then we get value in string format with negative integer as `"-35 -1"`.

And as per YML documentation `sequence use a dash followed by a space` : https://symfony.com/doc/current/components/yaml/yaml_format.html#collections.

So if we make string validation with ` Array(retryable_exit_codes).flatten.flat_map{|v| v.is_a?(String) ? v.split(/,/) : v}.map(&:to_i).include?(exit_code) ` then we are allowing the user to pass `invalid yml format` .

Please let me know if we should allow the invalid format or raise error in that case?

coderanger:
We need to rewind on the PR and clarify what the goal here is. We do not generally work around broken YAML in Test Kitchen and this would be the wrong place to start that. Please remember we cannot see the original MSYS ticket so my only context is what's given here.

NimishaS:
This PR is corresponding to https://chefio.atlassian.net/browse/MSYS-703. @btm , can you please comment if these changes are required?

NAshwini:
@coderanger, I have updated the description, Please let me know in case of any doubt

coderanger:
So it sounds like maybe throw in a `.flatten` for UX-y goodness, the rest might be overkill. Especially since we now set that value by default and users should almost never have to touch it.

spec/kitchen/transport/base_spec.rb
```diff
@@ -120,4 +120,16 @@
...
     end
   end
+
+  describe "#retry?" do
+    it "raises Invalid Type Error for retry_on_exit_code" do
+      proc { connection.retry?(2, 2, "35 1", 35) }
+        .must_raise("undefined method `flatten' for \"35 1\":String")
+    end
+
+    it "when value is send in Array in Array" do
```
btm: This could be more clear written like, "returns true when the retryable exit codes are formatted in a nested array"

spec/kitchen/transport/base_spec.rb
```diff
@@ -120,4 +120,16 @@
       connection.execute_with_retry("Hi", [123], 3, 1).must_equal "Hello"
     end
   end
+
+  describe "#retry?" do
+    it "raises Invalid Type Error for retry_on_exit_code" do
```
btm: is 'Invalid Type Error' left over from a previous iteration?

This could be more more clear as something like `it "raises an exception when multiple retryable exit codes are passed as a string"`.

NAshwini:
Thanks you @btm , I have fixed the review comments.

btm:
appveyor failure is the new 'bundle is part of ruby' failure and unrelated.

PR #1329, opened 2017-12-08
[MSYS-721] Added KITCHEN_SSH_PROXY feature to connect via http proxy

NAshwini:
The card [MSYS-721] & PR is related to use an http proxy as documented in https://net-ssh.github.io/ssh/v1/chapter-7.html . The design is that the URL for the https proxy is stored in `KITCHEN_SSH_PROXY` with port in `HTTP_PROXY_PORT`, and the user and password for authentication to the proxy (if required) comes from `HTTP_PROXY_USER` and `HTTP_PROXY_PASSWORD`.

### DIFF
--- a/lib/kitchen/transport/ssh.rb
+++ b/lib/kitchen/transport/ssh.rb
@@ -480,7 +480,7 @@ module Kitchen
           options_http_proxy = {}
           options_http_proxy[:user] = data[:http_proxy_user]
           options_http_proxy[:password] = data[:http_proxy_password]
-          opts[:proxy] = Net::SSH::Proxy::HTTP.new("#{data[:kitchen_ssh_proxy]}", data[:http_proxy_port], options_http_proxy)
+          opts[:proxy] = Net::SSH::Proxy::HTTP.new(data[:kitchen_ssh_proxy], data[:http_proxy_port], options_http_proxy)
         end
 
         if data[:ssh_key_only]
--- a/lib/kitchen/transport/ssh.rb
+++ b/lib/kitchen/transport/ssh.rb
@@ -21,6 +21,7 @@ require "kitchen"
 require "fileutils"
 require "net/ssh"
 require "net/ssh/gateway"
+require "net/ssh/proxy/http"
 require "net/scp"
 require "timeout"
 require "benchmark"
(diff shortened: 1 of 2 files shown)


lib/kitchen/transport/ssh.rb
```diff
@@ -446,6 +476,13 @@ def connection_options(data)
           opts[:auth_methods] = ["publickey"]
         end
 
+        if data[:kitchen_ssh_proxy]
+          options_http_proxy = {}
+          options_http_proxy[:user] = data[:http_proxy_user]
+          options_http_proxy[:password] = data[:http_proxy_password]
+          opts[:proxy] = Net::SSH::Proxy::HTTP.new("#{data[:kitchen_ssh_proxy]}", data[:http_proxy_port], options_http_proxy)
```
thommay: `"#{data[:kitchen_ssh_proxy]}"` should not be coerced into a string.

cheeseplus:
My only concern with this one is the naming of the additional options:

```
http_proxy_user
http_proxy_password
http_proxy_port
```

As those will strictly only apply to `kitchen_ssh_proxy` and users might be easily confused expecting these to work as generic toggles. Not a massive issue but one that might be worth clarifying a bit.

NAshwini:
@cheeseplus, @thommay, @btm should I replace name of additional options with the original one with below mapping as:
`http_proxy_user` as => `kitchen_http_proxy_user`
`http_proxy_password` as => `kitchen_http_proxy_password`
`http_proxy_port` as => `kitchen_http_proxy_port` and
`kitchen_ssh_proxy` as => `kitchen_http_proxy`

cheeseplus:
My thoughts would be slightly different, I want to make it clear that these settings _only_ affect SSH proxies so something like:

```
ssh_http_proxy_user
ssh_http_proxy_password
ssh_http_proxy_port
ssh_http_proxy
```
OR

```
ssh_proxy_user
ssh_proxy_password
ssh_proxy_port
ssh_proxy
```

@thommay @btm ? I just want to remain as consistent as possible so we don't have naming all over the place which has been a problem in the past.

thommay:
`ssh_http_proxy_user` and friends LGTM

NAshwini:
Thanks @cheeseplus , @thommay, @btm I have made changes as per review comments, But travis is getting failed with below error which does'nt seems to be related to the updation of code ` `let': let 'klass' cannot override a method in Minitest::Spec. Please use another name. (ArgumentError)
	from C:/projects/test_kitchen/spec/kitchen/platform_spec.rb:26:in `block in <top (required)>'
	from C:/Ruby23/lib/ruby/gems/2.3.0/gems/minitest-5.11.1/lib/minitest/spec.rb:83:in `class_eval'
	from C:/Ruby23/lib/ruby/gems/2.3.0/gems/minitest-5.11.1/lib/minitest/spec.rb:83:in `describe'`

PR #1340, opened 2018-01-16
Only allow one deprecation message to print per config

wrightp:
This is definitely not one of my proudest PRs. Figured I'd at least throw myself into the fire on this one.

If anything it unblocks the UX issue and we can get TK released.

Confession: This is the first time I've ever introduced a global var in Ruby...


wrightp:
@coderanger @cheeseplus

PR #1363, opened 2018-02-24
Prefer kitchen.yml to .kitchen.yml

thommay:


burtlo:
It will no longer be a hidden file! The file with all that configuration and power. I like that.

burtlo:
I definitely agree that it might be good to describe which one is being used when there two files are present.

jjasghar:
I am all about this. It’ll help when you have multiple kitchen yamls also.

robbkidd:
Will this also prefer a `kitchen.local.yml` over a `.kitchen.local.yml`?

And mix a `kitchen.yml` with a `.kitchen.local.yml` if there is no `kitchen.local.yml`?

cheeseplus:
As @robbkidd points out, we should really think about all the config files involved

- kitchen.yml
- kitchen.local.yml
- ~/.kitchen.yml
- ~/.kitchen/config.yml

Aside from the last one there, we probably want to think about those too.

thommay:
> Will this also prefer a kitchen.local.yml over a .kitchen.local.yml?

It does not currently. It could?

thommay:
> As @robbkidd points out, we should really think about all the config files involved

the files direct in your home directory we should not change, the others we can.

thommay:
> Will this also prefer a kitchen.local.yml over a .kitchen.local.yml

As it turns out, yes. But the decision is binding - if you have a `kitchen.yml`, you need a `kitchen.local.yml` (and if you tell kitchen to use `asparagus.yml`, it'll pick up `asparagus.local.yml` too).

robbkidd:
> As it turns out, yes. But the decision is binding - if you have a `kitchen.yml`, you need a `kitchen.local.yml` (and if you tell kitchen to use `asparagus.yml`, it'll pick up `asparagus.local.yml` too).

As a previous user, I would probably find this a surprising change. I'd expect if I override with `KITCHEN_YAML=asparagus.yml` that kitchen would still bring in my existing `.kitchen.local.yml`. But if we document the new behavior—and maybe include message output?—that should be alright.

```
INFO: found and using asparagus.yml as config (did not find asparagus.local.yml).
WARN: Consuming asparagus will make your pee smell funny.
```

coderanger:
@robbkidd That part isn't a change, it's how things work already https://github.com/test-kitchen/test-kitchen/blob/master/lib/kitchen/loader/yaml.rb#L203

robbkidd:
> That part isn't a change, it's how things work already.

... I revise my earlier statement: as a previous user, I've apparently been ignorant of this behavior the whole time.

PR #1368, opened 2018-02-28
include list of known plugins when one cannot be loaded

### LINKED ISSUE #879: Improve provisioner not found errors
Kitchen isn't very helpful if you specify a provisioner using hyphens instead of underscores, e.g.,

``` yaml
provisioner:
  name: chef-solo
```

Gives you just:

```
>>>>>> ------Exception-------
>>>>>> Class: Kitchen::ClientError
>>>>>> Message: Could not load the 'chef-solo' provisioner from the load path. Please ensure that your provisioner is installed as a gem or included in your Gemfile if using Bundler.
>>>>>> ----------------------
```

It'd be nice if kitchen could tell you which provisioners are available that it can find, or auto-corrected hyphens (assuming you want to enforce the convention of using underscores), or had some other mechanism to help get users back on track quickly when they make this mistake.

robbkidd:
In an effort to be friendlier, let's include a list of plugins kitchen can find when it cannot load the one named in a config.

Because the require assumes the plugin will be in the load path at `kitchen/{plugin-type}`, that is what is used to walk the load path looking for any Ruby file present at that location that defines a class.

Closes #879

### Driver

Could not load the 'nope' driver from the load path. Did you mean: dokken, dummy, exec, proxy, vagrant ? Please ensure that your driver is installed as a gem or included in your Gemfile if using Bundler.

### Provisioner

Could not load the 'nope' provisioner from the load path. Did you mean: chef_apply, chef_solo, chef_zero, dokken, dummy, shell ? Please ensure that your provisioner is spelled correctly, installed as a gem, or is included in your Gemfile if using Bundler.

### Transport

Could not load the 'nope' transport from the load path. Did you mean: dokken, dummy, exec, ssh, winrm ? Please ensure that your transport is installed as a gem or included in your Gemfile if using Bundler.

### Verifier

Could not load the 'nope' verifier from the load path. Did you mean: busser, dummy, inspec, shell ? Please ensure that your transport is installed as a gem or included in your Gemfile if using Bundler.

## TODO

- [x] filter out known base classes that don't actually implement a full plugin

### DIFF
--- a/lib/kitchen/driver.rb
+++ b/lib/kitchen/driver.rb
@@ -16,9 +16,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-require "kitchen/errors"
-require "kitchen/util"
-require "thor/util"
+require "kitchen/plugin"
 
 module Kitchen
   # A driver is responsible for carrying out the lifecycle activities of an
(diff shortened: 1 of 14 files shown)


robbkidd:
@cheeseplus `base`s removed from output. :eyes:❓

coderanger:
So this might result in some very frustrating user output if they get the error and then see the complaining driver in the list of "available" drivers :-/ That can happen if the driver is installed but not loadable due to dependency conflicts. Unfortunately there is also not a good way to actually check that :(

robbkidd:
@coderanger That's a good point. We could expand the error message to include the `LoadError` exception. Something along the lines of:

> Couldn't load X. Did you mean: X, Y, Z? If the plugin you are using appears in this list of those available, see the error information below for why the plugin was not able to load.

Totally open to better wordsmithing of that.

robbkidd:
Updated. Got carried away with a new `Plugin` module so there's not so much repetition with this gnarly load logic.

PR #1369, opened 2018-02-28
honor root_path for location of chef installer script

### LINKED ISSUE #1364: Error executing /tmp/chef-installer.sh
## Description
Due to security standards for the OS images we use with Test Kitchen (and everywhere else) the `noexec` option is set for `tmpfs` in `/etc/fstab`.  This results in this error with Test Kitchen:
```
       Preparing files for transfer
       Preparing dna.json
       Resolving cookbook dependencies with Berkshelf 6.3.1...
       Preparing validation.pem
       Preparing client.rb
       sudo: unable to execute /tmp/chef-installer.sh: Permission denied
```
The `root_path` provisioner option does not appear to change the location of `chef-installer.sh`.  It looks hard-coded here:
https://github.com/test-kitchen/test-kitchen/blob/master/lib/kitchen/provisioner/chef_base.rb#L506

## Kitchen Version
```ruby
Test Kitchen version 1.20.0
```

## ChefDK Version
```ruby
Chef Development Kit Version: 2.3.4
chef-client version: 13.6.4
delivery version: master (73ebb72a6c42b3d2ff5370c476be800fee7e5427)
berks version: 6.3.1
kitchen version: 1.20.0
inspec version: 1.50.1
```
## Platform Version
Fedora 25

## Replication Case
Setup a server with the `noexec` option set on `tmpfs` and run `test kitchen <suite>`.  
Relevant line from our `/etc/fstab` file:
```
tmpfs /dev/shm tmpfs  defaults,nodev,nosuid,noexec,size=1g  0 0
```

## Kitchen Output
```ruby
Please see above
```

robbkidd:
If `/tmp` is mounted with noexec, then the `chmod` and `sudo the_script` that are attempted later fail. Instead of hard-coding the location for the script, write and run it from the provisioner's configured
root_path.

This fixes #1364 for the code path when `product_name` is present in the Chef provisioner config.

### DIFF
--- a/lib/kitchen/provisioner/chef_base.rb
+++ b/lib/kitchen/provisioner/chef_base.rb
@@ -503,8 +503,9 @@ module Kitchen
       end
 
       def install_from_file(command)
-        install_file = "/tmp/chef-installer.sh"
-        script = ["cat > #{install_file} <<\"EOL\""]
+        install_file = "#{config[:root_path]}/chef-installer.sh"
+        script = ["mkdir -p #{config[:root_path]}"]
+        script << "cat > #{install_file} <<\"EOL\""
         script << command
         script << "EOL"
         script << "chmod +x #{install_file}"


cheeseplus:
I think so and ideally that codepath goes away/is deprecated.

Issue #1370, opened 2018-03-01
Support IPv6 addresses for WinRM


jzinn:
## Description

A "bad URI" exception occurs when running `kitchen converge` for a Windows machine when accessed via WinRM over IPv6.

The problem is that brackets are missing.  This
```
http://caec:cec6:c4ef:bb7b:1a78:d055:216d:3a78:5985/wsman
```
should be
```
http://[caec:cec6:c4ef:bb7b:1a78:d055:216d:3a78]:5985/wsman
```

In addition to the pull request I am submitting, I had to fix an IPv6 issue in the httpclient gem that test-kitchen depends on: nahi/httpclient#387.

## Kitchen Version

```ruby
Test Kitchen version 1.20.0
```

## Ruby Version

If you are not using test-kitchen via ChefDK, please provide the output of ruby --version.

```ruby
ruby 2.4.3p205 (2017-12-14 revision 61247) [x86_64-linux]
```

## Platform Version

Fedora 27

## Replication Case

1. Run `kitchen create`.

    (See https://github.com/chef-partners/chef-provisioning-vsphere/pull/81 for IPv6 issue there.)

1. Ensure `.kitchen/the-vm.yml` contains an IPv6 address for `hostname`:

    ```
    ---
    vsphere_name: the-vm-9d15976e
    username: vagrant
    password: vagrant
    server_id: 50047130-9675-6f49-1bde-b54cf7cb6901
    hostname: caec:cec6:c4ef:bb7b:1a78:d055:216d:3a78
    ````

1. Run `kitchen converge`

## Kitchen Output

```
    Finished creating <the-vm> (2m7.13s).
-----> Converging <the-vm>...
    Preparing files for transfer
    Preparing dna.json
    Resolving cookbook dependencies with Berkshelf 6.3.1...
    Removing non-cookbook files before transfer
    Preparing data
    Preparing nodes
    Preparing validation.pem
    Preparing client.rb
>>>>>> ------Exception-------
>>>>>> Class: Kitchen::ActionFailed
>>>>>> Message: 1 actions failed.
>>>>>>     Failed to complete #converge action: [bad URI(is not URI?): http://caec:cec6:c4ef:bb7b:1a78:d055:216d:3a78:5985/wsman] on the-vm
>>>>>> ----------------------
>>>>>> Please see .kitchen/logs/kitchen.log for more details
>>>>>> Also try running `kitchen diagnose --all` for configuration
```

PR #1371, opened 2018-03-01
Support IPv6 addresses for WinRM

jzinn:
Fixes test-kitchen/test-kitchen#1370

### DIFF
--- a/lib/kitchen/transport/winrm.rb
+++ b/lib/kitchen/transport/winrm.rb
@@ -48,12 +48,11 @@ module Kitchen
       default_config :connection_retry_sleep, 1
       default_config :max_wait_until_ready, 600
       default_config :winrm_transport, :negotiate
+      default_config :scheme do |transport|
+        transport[:winrm_transport] == :ssl ? "https" : "http"
+      end
       default_config :port do |transport|
         transport[:winrm_transport] == :ssl ? 5986 : 5985
-      end
-      default_config :endpoint_template do |transport|
-        scheme = transport[:winrm_transport] == :ssl ? "https" : "http"
-        "#{scheme}://%{hostname}:%{port}/wsman"
       end
 
       def finalize_config!(instance)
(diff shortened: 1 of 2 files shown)


cheeseplus:
This needs two minor style fixes to get the build green

```
Inspecting 133 files
.....................................................C...................................................................C...........
Offenses:
spec/kitchen/transport/winrm_spec.rb:72:37: C: Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.
      transport[:scheme].must_equal 'http'
                                    ^^^^^^
lib/kitchen/transport/winrm.rb:394:17: C: Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.
          path: '/wsman'
                ^^^^^^^^
```

jzinn:
Updated

cheeseplus:
My only concern here is that we're taking away an option and replacing it with a new one. I'm not sure how many people even knew about `endpoint_template` but I'd guess it's so few that we can probably get away with changing this w/o deprecation.

jzinn:
I share the concern, and gave it some thought before submitting my PR.  I decided to submit a minimal change for clarity.  Adding deprecation code might not be too much work for someone more familiar with the code base and its deprecation process.

#### Why have `endpoint_template` in the first place?

I looked for motivation for making the endpoint more configurable.  I could not find any uses of `endpoint_template`.  The git history for `endpoint_template` seems to be lost.  A quick search of some other parts of Chef didn't turn up any similar URL template features.  A small sample of other Ruby WinRM code I looked at recently doesn't seem to offer more URL configuration.

I agree that it's probably a little-used feature.

#### If an alternative to `endpoint_template` is needed

If the endpoint does needs to be more configurable, then I would suggest adding more options for the parts of the URL that need it.  For instance, in addition to `scheme`, `host`, and `port`, we could add `path`, by adding a `default_config`:

```ruby
default_config :path, "/wsman"
```
and using it:
```ruby
endpoint = URI::Generic.build(
  scheme: data.fetch(:scheme),
  host: data.fetch(:hostname),
  port: data.fetch(:port),
  path: data.fetch(:path)
).to_s
```

cheeseplus:
A thorough and complete answer 👍 

I think we're safe to change this for the better and enhance/fix later as needed.

PR #1377, opened 2018-03-01
halt if visible & hidden default configs are both present

### LINKED ISSUE #1372: ERROR and exit with message when both kitchen.yml and .kitchen.yml are present.
#1363 got merged without us addressing the very sensible suggestion from @coderanger that we should warn users about this preference when both types of kitchen configs are present.

robbkidd:
Now that preferring `kitchen.yml` over `.kitchen.yml` has been added, we should error, halt, and warn users when both are present. This ought to make troubleshooting clearer about from where configuration is actually being sourced.

The same rule applies for the default local config which is computed based on the project config.

Closes #1372

### TODO:

- [x] update init to generate not-dotted `kitchen.yml`

### DIFF
--- a/lib/kitchen/loader/yaml.rb
+++ b/lib/kitchen/loader/yaml.rb
@@ -34,9 +34,9 @@ module Kitchen
       #
       # @param options [Hash] configuration for a new loader
       # @option options [String] :project_config path to the Kitchen
-      #   config YAML file (default: `./.kitchen.yml`)
+      #   config YAML file (default: `./kitchen.yml`)
       # @option options [String] :local_config path to the Kitchen local
-      #   config YAML file (default: `./.kitchen.local.yml`)
+      #   config YAML file (default: `./kitchen.local.yml`)
       # @option options [String] :global_config path to the Kitchen global
       #   config YAML file (default: `$HOME/.kitchen/config.yml`)
       # @option options [String] :process_erb whether or not to process YAML
@@ -192,6 +192,10 @@ module Kitchen
       # @return [String] an absolute path to a Kitchen config YAML file
       # @api private
       def default_config_file
+        if File.exist?(kitchen_yml) && File.exist?(dot_kitchen_yml)
+          raise UserError, "Both #{kitchen_yml} and #{dot_kitchen_yml} found. Please use the un-dotted variant: #{kitchen_yml}."
+        end
+
         File.exist?(kitchen_yml) ? kitchen_yml : dot_kitchen_yml
       end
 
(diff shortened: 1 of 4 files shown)


lib/kitchen/loader/yaml.rb
```diff
@@ -207,11 +211,19 @@ def dot_kitchen_yml
...
-      #
+
       # @return [String] an absolute path to a Kitchen local YAML file
       # @api private
       def default_local_config_file
-        config_file.sub(/(#{File.extname(config_file)})$/, '.local\1')
+        default_local_config = config_file.sub(/(#{File.extname(config_file)})$/, '.local\1')
+
+        unhidden_config = default_local_config.sub(/([\/])\.([^\/]*)$/, '\1\2')
+        hidden_config = default_local_config.sub(/([\/])([^\.][^\/]*)$/, '\1.\2')
```
robbkidd: I'm not proud of this.

robbkidd:
The actual error messages could probably use some word-smithing.

lib/kitchen/loader/yaml.rb
```diff
@@ -192,6 +192,10 @@ def read_file(file)
       # @return [String] an absolute path to a Kitchen config YAML file
       # @api private
       def default_config_file
+        if File.exist?(kitchen_yml) && File.exist?(dot_kitchen_yml)
+          raise UserError, "Both #{kitchen_yml} and #{dot_kitchen_yml} found. Use one or the other. Test-Kitchen prefers the not-hidden filename."
+        end
```
cheeseplus: "Please use one or the other. The hidden file will be deprecated in the future."

robbkidd:
Updated. Should we also update `init` to generate a not-dotted `kitchen.yml`?

cheeseplus:
Definitely

lib/kitchen/loader/yaml.rb
coderanger: How about this?

```ruby
localize = lambda path: path.sub(/(#{File.extname(path)})$/, '.local\1')
if config_file == default_config_file && File.exist?(localize.call(kitchen_yml) && File.exist?(localize.call(dot_kitchen_yml)
```

or something like that?

robbkidd: I like it. I had a phase of this change that had a whole method to do the `.local`izing, but an anonymous lambda in here is more better.

robbkidd: Ah. Issue: by calling `kitchen_yml` and `dot_kitchen_yml` here, it only tests the existence of `kitchen.local.yml` and `.kitchen.local.yml`, but does not catch the case when `KITCHEN_YAML` is set to something and the `.local` version is computed based on that.

For example:

```
> export KITCHEN_YAML=asparagus.yml
> ls *.yml
asparagus.yml
asparagus.local.yml
.asparagus.local.yml
```

Ought to error just as when `kitchen.local.yml` and `.kitchen.local.yml` both being present.

coderanger: Hmm, I mean if it doesn't catch those I'm much less concerned, but let's see:

```ruby
localize = lambda path: path.sub(/(#{File.extname(path)})$/, '.local\1')
config_dir, config_base = File.split(config_file)
undot_config_base = config_base.gsub(/^\./, '')
dot_config_base = ".#{undot_config_base}"
if File.exist?(localize.call(File.join(config_dir, undot_config_base))) && File.exist?(localize.call(File.join(config_dir, dot_config_base)))
```

A little more verbose but maybe less magical with the regexes :)

PR #1380, opened 2018-03-07
Adds kitchen-dokken template override when driver specified

burtlo:
To provide a better experience for the kitchen dokken users it would be
helpful if there was a way to generate a kitchen configuration with more
of the fields required to make it work. As this plugin is actually several
that act in concert this may be the right approach.

It also may not be the right approach. In the situation that someone wants
to specify multiple drivers. There were also some thoughts about the `init`
command going away completely.

Perhaps this template should instead be the the default of the Chef DK.

Let's discuss.

Signed-off-by: Franklin Webber <franklin@chef.io>

### DIFF
--- a/lib/kitchen/generator/init.rb
+++ b/lib/kitchen/generator/init.rb
@@ -57,7 +57,12 @@ module Kitchen
       def init
         self.class.source_root(Kitchen.source_root.join("templates", "init"))
 
-        create_kitchen_yaml
+        if options[:driver].include?('kitchen-dokken')
+          create_kitchen_dokken_yaml
+        else
+          create_kitchen_yaml
+        end
+
         create_chefignore
         prepare_rakefile
         prepare_thorfile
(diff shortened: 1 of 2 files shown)


lib/kitchen/generator/init.rb
```diff
@@ -57,7 +57,12 @@ class Init < Thor::Group
       def init
         self.class.source_root(Kitchen.source_root.join("templates", "init"))
 
-        create_kitchen_yaml
+        if options[:driver].include?('kitchen-dokken')
```
burtlo: I don't particularly like this part. I imagine that if anyone is actually using this driver flag and wants more drivers defined then we are ignoring their requests.

burtlo:
This may not be the PR we want @robbkidd. But it could be the PR we deserve.

coderanger:
:-1: from me, `kitchen init` is not something Chef users should ever see. They should be using `chef generate`. The only reason we've kept `kitchen init` is to help non-Chef provisioner users, however as dokken only supports Chef, there is no overlap there.

burtlo:
I love "bad" ideas. They get the attention they deserve.

cheeseplus:
Closing this in favor of https://github.com/chef/chef-dk/pull/1517

Issue #1209, opened 2017-03-19
Shell verifier clobbers environment variables


smith:
## Description

When using the shell verifier and setting

```yaml
verifier:
  name: shell
  shellout_opts:
    environment:
      FOO: bar
```

the command will not be run with `FOO=bar`, because [`merge_state_to_env`](https://github.com/test-kitchen/test-kitchen/blob/3768d09fd1f0383fce57a261267bbfcf9a4f2a10/lib/kitchen/verifier/shell.rb#L87-L96) does a shallow merge, wiping out the `:environment` key.

devkid:
Can confirm this issue. Any plans on fixing this?

cheeseplus:
Unfortunately the shell verifier doesn't get a ton of love so we typically rely on folks who use it regularly to PR issues like this one. I'm confirming it as a bug but it'd be great if folks who use the provisioner day-to-day could PR a fix for this one to expedite the fix.

PR #1381, opened 2018-03-09
Shell Provisioner: make script executable

thewyzard44:
Resolves #931 and runs `chmod +x ...` in the guest, prior to executing the script

Based off of #442 but includes a guest OS check and only runs the command on *nix

PR #1382, opened 2018-03-09
Shell Verifier rewrite

thewyzard44:
Signed-off-by: Sean Zachariasen <thewyzard@hotmail.com>

Addressing #1209

This one needs some love.  This initial commit makes the options merge correctly, but the rest of the verifier needs touched up in some surrounding areas.  There's incongruity between 2 different methods of local execution and no effort has been applied by the OA to make sure the environment gets applied when running remotely

### DIFF
--- a/lib/kitchen/verifier/shell.rb
+++ b/lib/kitchen/verifier/shell.rb
@@ -41,11 +41,15 @@ module Kitchen
       def call(state)
         info("[#{name}] Verify on instance #{instance.name} with state=#{state}")
         sleep_if_set
-        @merged_environment = config[:environment].merge(state_to_env(state))
+        @merged_environment = state_to_env(state).merge(config[:environment] || {})
 
         if config[:remote_exec]
-          instance.transport.connection(state) do |conn|
-            conn.execute(remote_command)
+          begin
+            instance.transport.connection(state) do |conn|
+              conn.execute(remote_command)
+            end
+          rescue Kitchen::Transport::TransportFailed => ex
+            raise ActionFailed, ex.message
           end
         else
           shellout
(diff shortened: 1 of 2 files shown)


coderanger:
What is the end goal here?

thewyzard44:
I want to remove this `run_command` method https://github.com/test-kitchen/test-kitchen/blob/master/lib/kitchen/verifier/shell.rb#L55-L62.  It is not called in any normal command flow.

If anyone is calling that manually (and I'd be surprised if you are because it is broken) let me know.

I will try to deprecate it instead, if the new deprecation framework will allow this.

lib/kitchen/verifier/shell.rb
```diff
@@ -73,8 +102,29 @@ def sleep_if_set
         end
       end
 
+      # Merges environment variables with :shellout_opts as specified in the config
+      #
+      # @return [Hash] options to be passed to shellout
+      # @api private
+      def shellout_opts
+        config[:shellout_opts].dup.tap do |options|
```
thewyzard44: cover 'if config[:shellout_opts].nil?`

thewyzard44:
apologies - I answered coderanger OOB...  but in short, I'd like this verifier to work locally, or remotely, and to pass the expected environment when and wherever it runs.

It may be prudent to use a `script:` value in parity with the shell provisioner, if you just want to run a shell script.  But RFC, it'll have issues similar to #1381.

And so should this rewrite include a `script:` option?  Or should it be left as-is and expect the end user to spectify `sh -c myscript` as an option to `command:`

This is a forward moving question for the kitchen team in that they didn't even really like `chmod +x myscript` but took it due to lack of options....   to quote coderanger "sigh windows"

And also a community question for those vested, or interested, in a good shell verifier

cheeseplus:
The community we have is the community that shows up with opinions and PRs so I think so long as we're moving it forward in a "less bad" direction then we are good.

damacus:
Hey @thewyzard44 do we still want this PR and how can we help you get this finished?

PR #1385, opened 2018-03-16
post_create_command for executing arbitrary commands after create action

blbradley:
Hello!

I thought Test Kitchen might need this.

I haven't wrote any Ruby in a few years. So, forgive my lack of incite. This is basically copy/paste from the `pre_create_command` hook.

I have used this successfully through the `kitchen-google` driver. See [blbradley/kitchen-google@hook-post_create_command](https://github.com/blbradley/kitchen-google/tree/hook-post_create_command).

### DIFF
--- a/lib/kitchen/driver/base.rb
+++ b/lib/kitchen/driver/base.rb
@@ -150,6 +150,18 @@ module Kitchen
         end
       end
 
+      # Run command if config[:post_create_command] is set
+      def post_create_command
+        if config[:post_create_command]
+          begin
+            run_command(config[:post_create_command])
+          rescue ShellCommandFailed => error
+            raise ActionFailed,
+              "post_create_command '#{config[:post_create_command]}' failed to execute #{error}"
+          end
+        end
+      end
+
       # Intercepts any bare #puts calls in subclasses and issues an INFO log
       # event instead.
       #
--- a/lib/kitchen/driver/base.rb
+++ b/lib/kitchen/driver/base.rb
@@ -30,6 +30,7 @@ module Kitchen
       include ShellOut
 
       default_config :pre_create_command, nil
+      default_config :post_create_command, nil
 
       # Creates a new Driver object using the provided configuration data
       # which will be merged with any default configuration.
--- a/spec/kitchen/driver/base_spec.rb
+++ b/spec/kitchen/driver/base_spec.rb
@@ -106,6 +106,24 @@ describe Kitchen::Driver::Base do
     end
   end
 
+  describe ".post_create_command" do
+    it "run command" do
+      Kitchen::Driver::Base.any_instance.stubs(:run_command).returns(true)
+
+      config[:post_create_command] = "echo works 2&>1 > /dev/null"
+      driver.expects(:run_command).with("echo works 2&>1 > /dev/null")
+      driver.send(:post_create_command)
+    end
+
+    it "error if cannot run" do
+      class ShellCommandFailed < Kitchen::ShellOut::ShellCommandFailed; end
+      Kitchen::Driver::Base.any_instance.stubs(:run_command).raises(ShellCommandFailed, "Expected process to exit with [0], but received '1'")
+
+      config[:post_create_command] = "touch /this/dir/does/not/exist 2&>1 > /dev/null"
+      proc { driver.send(:post_create_command) }.must_raise Kitchen::ActionFailed
+    end
+  end
+
   describe ".no_parallel_for" do
     it "registers no serial actions when none are declared" do
       Kitchen::Driver::Speedy.serial_actions.must_be_nil


coderanger:
What kind of use case do you think this would be good for? We're trying to be careful about not turning TK into a generic orchestration framework so we like to have use cases in mind.

blbradley:
I'd like to do a DNS record update via a script after instance creation but before installation of the provisioner.

coderanger:
That sounds like something to handle inside the provisioner in most cases.

blbradley:
Yes! I agree with you and have implemented my use case privately with `kitchen-salt` and the new `prepare_salt_environment` parameter. However, I could do this in a `post_create_command` with more contextual information about the instance(like a KITCHEN_HOSTNAME or KITCHEN_SUITE environment variable). This would ensure my DNS record is live in _plenty_ of time.

I invite @chetanddesai  and @hectcastro to come forth and talk about their use cases.

cheeseplus:
I'm not as opposed as I once was since we've allowed `pre_create_command` and this would be very much under "use at your own risk". There are a handful of times this can be super useful and until we restructure things in 2.0 this will cover those cases.

brodygov:
👍 Would like to see this merged.

Primary use case: situations where the provisioner itself breaks or fails because it isn't able to correctly coordinate with other bootstrapping steps that happen on the instance independently. For example, we currently have major problems where the kitchen provisioner fails 80% of the time because the periodic apt upgrade service is in the middle of running `apt upgrade` and is holding the dpkg lock, so when the kitchen provisioner attempts to install chef it fails.

This could be addressed by a simple `sleep 30`, or better still by polling for a sentinel file dropped by systemd or cloud-init once system boot has entirely finished. But without a `post_create_command` there isn't any good hook for us to coordinate with actions that might be happening on the instance at startup or as part of user-data.

brodygov:
P.S. For the above mentioned use cases, because there are many more drivers than provisioners, it might be simpler to add a hook into the provisioner at the beginning of converge (just before installing chef) instead of into the driver at the end of create.

cheeseplus:
The user_data coordination problem is semi-known and I have some ideas for that - however the problem that you're experiencing there specifically can be mitigated by putting that logic in the recipe to look for an artifact/file that is dropped by user_data to signify this is done.

The `post_create_command` itself is smelly because it's solving a problem that isn't being solved well due to the intersection of underlying systems. The better and more holistic fix is to push this logic into kitchen-ec2 (or similar cloud driver) specifically OR have something in chef-client that performs some arbitrary check - the latter bit is hard cause there isn't anything universal hence why we recommend doing this in the provisioner in conjunction with user_data.

We still might merge this but this is a massive footgun that is compensating for something that I believe can be solved better at other layers.

brodygov:
No it can't be solved by putting anything in a recipe, because _chef installation_ fails.

cheeseplus:
If it's the installation of chef itself failing because apt-locks then the only option (otherwise) is building that into chef-client itself (unlikely) or shoving something into kitchen-ec2 to catch that user_data is done running (doable).

I still would prefer this be solved at a lower layer but short of baking images there isn't much you can do in that situation at present so I understand why this feature would be attractive. We're not sitting here trying to be unhelpful, this is mostly trying to prevent what @coderanger was mentioning earlier:

> We're trying to be careful about not turning TK into a generic orchestration framework

I'll think on this one tonight but unless I find something intractable I think we'll allow this in lieu of anything better.

coderanger:
This patch doesn't actually implement this feature. So regardless of discussions about if we should merge this feature, this PR is not actually mergable right now.

blbradley:
@coderanger What's the problem with the patch?

coderanger:
Okay, just double checked this but yeah, this PR doesn't actually implement anything. `pre_create_command` works because that is called from the base implementation of the `create` method, which we expect to be called by all plugins via `super`. For this to work it would need new cabling in the Instance class to actually call `post_create_command` at some point.

brodygov:
What would you think about implementing this in the provisioner instead as a `pre_install_command`? That way there are no changes needed to any of the drivers.

https://github.com/test-kitchen/test-kitchen/pull/1411

coderanger:
That would be unrelated. Your new PR is running the commands on the remote side, this one (and `pre_create_command`) run them locally.

brodygov:
Ah I misunderstood. All the use cases I have are for commands run on the remote side.

coderanger:
Heh, that is one of many reasons we have been hesitant to add features like this. They have a tendency to multiply when you aren't watching. The correct solution for remote-side stuff is multi-plugin support combined with the existing shell provisioner (which is why I did a major revamp on it a few months ago).

brodygov:
That sounds like a sure fire sign that the current architecture isn't flexible enough to solve user problems.

Is multi-plugin support underway? It's pretty frustrating to see 4 years of "we aren't going to merge simple fixes because we might someday implement a better architecture."

cheeseplus:
This is an open source project with incredibly limited bandwidth of  barely two non-fulltime maintainers. I'm sorry our free software frustrates you but I've had folks yelling at me for 2 years and this certainly doesn't help.

If you'd like to help maintain the project, there is plenty of work to be done but that's another conversation.

cheeseplus:
Closing this one as the PR won't function as indicated.

blbradley:
@coderanger 

> For this to work it would need new cabling in the Instance class to actually call post_create_command at some point.

This was likely the info I needed to make this patch work without expecting driver devs to call `post_create_command` themselves. I couldn't figure that out on my own, I believe.

@cheeseplus If the the info in the above block quote was implemented, would this patch still be considered?

FWIW, test-kitchen is a great tool. I've used it to implement test suites in the SaltStack plugin ecosystem with no issues.

blbradley:
P.S. After _two years_ of using it, this is the only time I've needed to file an issue or patch. NOT BAD, EH?

PR #1397, opened 2018-04-10
🚧 WIP: adds sending telemetry on TK actions 🚧

robbkidd:
🚧 Work In Progress 🚧

Adds the transmission of telemetry to a designated host based on [Chef RFC 95](https://github.com/chef/chef-rfc/blob/master/rfc095-test-kitchen-telemetry.md).

Current implementation wraps the high-level CLI command and each action that command triggers in a timer to send data to a data collector.

### TODO

- [ ] interaction with the user to select opt-in/opt-out
- [ ] notification to the user on opt-in/out status during a run
- [ ] identify a way to log telemetry messages sent so that the user has an opportunity to review

### DIFF
--- a/lib/kitchen/configurable.rb
+++ b/lib/kitchen/configurable.rb
@@ -133,6 +133,16 @@ module Kitchen
       result
     end
 
+    # Returns a Hash of useful information for telemetry
+    #
+    # @return [Hash] a hash of interesting stuff
+    def telemetry
+      {
+        driver_name: name,
+        driver_version: self.class.version,
+      }
+    end
+
     # Returns the name of this plugin, suitable for display in a CLI.
     #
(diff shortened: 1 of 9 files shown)


lib/kitchen/configurable.rb
```diff
@@ -133,6 +133,16 @@ def diagnose_plugin
       result
     end
 
+    # Returns a Hash of useful information for telemetry
+    #
+    # @return [Hash] a hash of interesting stuff
+    def telemetry
+      {
+        plugin_name: name,
+        plugin_version: self.class.version,
```
robbkidd: Might need a different version lookup here. I notice during a dokken run that this comes up nil/null.

lamont-granquist:
this is rather stalled, gonna close for now.

PR #1401, opened 2018-04-17
Minor technical cleanup and unify behavior for files and directories.

coderanger:
Bugged me so I fixed it :)


cheeseplus:
This still good to go? @coderanger

coderanger:
I think so, it's a pretty isolated change. Just slightly cleaner.

Issue #329, opened 2014-01-17
Add lifecycle hooks to various phases (including provisioners)


sethvargo:
Example #257 

@fnichol is going to tack on some more information/whiteboard picture :smile:.

CpuID:
*bump* any further updates to this one?

lamont-granquist:
@fnichol @tyler-ball buuuuuump

PurrBiscuit:
+1 for multi-provisioning

lxhunter:
bump

sochoa:
bump

sawanoboly:
+1

francoiskha:
+1

habelson:
+1

z0mi3ie:
+1 this would be very useful

willejs:
+1000

Chili-Man:
any updates on this ?

cheeseplus:
Related https://github.com/test-kitchen/test-kitchen/pull/564

pburkholder:
+1 on behalf of a commercial user I work with. (Thanks, @cheeseplus, for tagging for future release)

chetanddesai:
+1 for `post_create_command` - anyone know of a work around to hack this until the `post_create_command` is available?  Trying to execute a command after vagrant is up, but before Chef starts.

mwrock:
A hack would be to call `kitchen exec <command>` after `kitchen create` and before `kitchen converge`

pietervogelaar:
+1 I would really like this!

swimfrog:
+1

darwin67:
+1

netflash:
+1

RamblingCookieMonster:
+1, for an implementation that would allow a run-multiple-provisioners-in-sequence scenario.

Cheers!

anvarich:
+1

gurumaia:
+1

burythehammer:
+1, would love to run both a Terraform and Ansible provisioner in sequence. Found a workaround (trigger ansible within terraform), but it doesn't feel like the right way to do it at all.

azazi-sa:
+1

sputmayer:
+1

rylarson:
+1

jestallin:
+1

DI-DaveGoodine:
+1

seanguo:
+1

thommay:
[...]

PR #1411, opened 2018-05-03
Add a pre_install_command to the base provisioner.

brodygov:
This is useful for addressing any kind of bootstrapping problem that
might occur prior to provisioner installation requiring custom code.

For example, it is often useful to coordinate the provisioner
installation with actions that start at boot on an instance via some
kind of init system (e.g. systemd, cloud-init).

Users can by default add a "pre_install_command" script to the kitchen.yml with any commands necessary to run prior to installing the provisioner.

This example pauses the provisioner until systemd has finished booting the instance:

```yaml
provisioner:
  name: chef_zero
  pre_install_command: "while ! systemctl is-system-running; do echo Waiting for system boot to finish; sleep 15; done"
```

If this functionality isn't appropriate for provisioner subclasses, they can override `pre_install_command()` to change the behavior.

Somewhat related to #1385

### DIFF
--- a/lib/kitchen/provisioner/base.rb
+++ b/lib/kitchen/provisioner/base.rb
@@ -45,6 +45,8 @@ module Kitchen
         provisioner.windows_os? ? nil : "sudo -E"
       end
 
+      default_config :pre_install_command, nil
+
       default_config :command_prefix, nil
 
       default_config :downloads, {}
@@ -68,6 +70,8 @@ module Kitchen
         sandbox_dirs = Util.list_directory(sandbox_path)
 
         instance.transport.connection(state) do |conn|
+          info("Initializing provisioner on #{instance.to_str}")
+          conn.execute(pre_install_command)
           conn.execute(install_command)
           conn.execute(init_command)
           info("Transferring files to #{instance.to_str}")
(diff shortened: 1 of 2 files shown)


coderanger:
I think this is pushing it on the level of fiddly we want to support, probably over the line. Eventually you just have to make a plugin subclass.

brodygov:
I'm not really understanding the resistance here. It's very common to need to have more custom bootstrapping behavior in order to get chef installed.

https://github.com/test-kitchen/test-kitchen/pull/311
https://github.com/test-kitchen/test-kitchen/issues/329

In our case, it's currently not possible to mirror the actual production bootstrapping process in the test infrastructure because the test-kitchen chef provisioners don't allow us to wait or perform custom initial steps. We could use the shell provisioner, but since we do actually want to run chef there are a lot of features that the chef provisioners bring.

cheeseplus:
We have to own the maintenance of whatever is merged and we've had plenty of bad experiences with "fiddly" features that solve a niche but not general use case causing us undue amounts of grief.

cheeseplus:
Likely superseded by https://github.com/test-kitchen/test-kitchen/pull/1428

coderanger:
Yeah, this would now be a `pre_converge` lifecycle hook I think.

cheeseplus:
Closing this one as it should be covered and if not then we'll want the implementation/fix done in the context of the new lifecycle hooks.

PR #1419, opened 2018-06-11
handle net-ssh >= 5.0 (WIP)

Val:


test-kitchen.gemspec
```diff
@@ -25,8 +25,8 @@ Gem::Specification.new do |gem|
 
   gem.add_dependency "mixlib-shellout", ">= 1.2", "< 3.0"
   gem.add_dependency "net-scp",         "~> 1.1"
-  gem.add_dependency "net-ssh", ">= 2.9", "< 5.0"
-  gem.add_dependency "net-ssh-gateway", "~> 1.2"
+  gem.add_dependency "net-ssh", ">= 5.0.1"
```
coderanger: This needs to be fixed to stay at least marginally compatible with what we had before lest we break a lot of things.

PR #1427, opened 2018-06-28
Add post_create_command config option for drivers

tarcinil:
Adds post_create_command config and method into Driver::Base for drivers to implement logic.


cheeseplus:
Superseded by https://github.com/test-kitchen/test-kitchen/pull/1428

PR #1428, opened 2018-06-28
Lifecycle hooks

coderanger:
ｈｕｌｋ　ｓｍａｓｈ. To copy from the release notes because it's midnight and I'm tired:

The life cycle hooks system allows running commands before or after any phase
of Test Kitchen (`create`, `converge`, `verify`, or `destroy`). Commands can be
run either locally on your workstation (the default) or remotely on the test instance.

These hooks are configured under a new `lifecycle:` section in `kitchen.yml`:

```yaml
lifecycle:
  pre_create: echo before
  post_create:
  - echo after
  - local: echo also after
  - remote: echo after but in the instance
```

You can also configure hooks on a single platform or suite:

```yaml
platforms:
- name: ubuntu-18.04
  lifecycle:
    pre_converge:
    - remote: apt update

suites:
- name: default
  lifecycle:
    post_verify:
    - my_coverage_formatter
```


coderanger:
@tarcinil :)

coderanger:
Someone should probably test this with Windows just for the sake of testing, I didn't have a win box handy.

tarcinil:
@coderanger Great Work.  Thank you for the effort.  

I have an idea for an UX enhancement.  I will put it to words and review fully after work today.

lib/kitchen/lifecycle_hooks.rb
```diff
@@ -0,0 +1,84 @@
...
+      state = nil
+      hook_data.each do |hook|
+        # Coerce the common case of a bare string to be a local command. This
+        # is to match the behavior of the old `pre_create_command` semi-hook.
+        hook = { local: hook } if hook.is_a?(String)
+        if hook.include?(:local)
+          # Local command execution on the workstation.
+          cmd = hook.delete(:local)
+          run_command(cmd, hook)
+        elsif hook.include?(:remote)
```
tarcinil: Would it be possible to `raise UserError, "Unable to use a remote in  #{hook.phase}, as no instance is created to execute against."` if a remote is used in to pre_create or post_destroy phase?

This would ensure that remote is only used when `instance` is available and prevents possible bug reports when possibly attempted.

thewyzard44:
I'll poke this from windows in the next day or so
+1!!

coderanger:
I should add that I have no reason to suspect this will fail on Windows. It uses the standard APIs, I just haven't actually tried it :)

tarcinil:
This fixed #329.

tarcinil:
When is this getting cut into a gem release?  Are you thinking that #1432 would be included?

cheeseplus:
I've been out for vacation but this will get a release today or tomorrow assuming we merge/deny #1432

tarcinil:
@cheeseplus Thanks for letting me know.  This is something that we've been looking forward to.

aaronlippold:
Very very cool

darkn3rd:
This is so awesome.

PR #1430, opened 2018-07-02
Synchronize calls to chdir to be thread safe

### LINKED ISSUE #1416: Race condition: conflicting chdir during another chdir block
## Description

Sometime, when doing kitchen actions like converge or verify in concurrent mode (ex: `kitchen verify -c`), I may have some:
```
~/.gem/ruby/2.5.0/gems/test-kitchen-1.21.1/lib/kitchen/util.rb:194: warning: conflicting chdir during another chdir block
~/.gem/ruby/2.5.0/gems/test-kitchen-1.21.1/lib/kitchen/util.rb:194: warning: conflicting chdir during another chdir block
~/.gem/ruby/2.5.0/gems/test-kitchen-1.21.1/lib/kitchen/util.rb:163: warning: conflicting chdir during another chdir block
~/.gem/ruby/2.5.0/gems/test-kitchen-1.21.1/lib/kitchen/util.rb:163: warning: conflicting chdir during another chdir block
```

and then, later the run failed with some file not found errors like:
```
LoadError:
  cannot load such file -- common_config
```

Re-running it may work or fail on another file, running it without `-c` works without problem.

## Kitchen Version

Tell us which version of test-kitchen you are using (`kitchen --version`).  

```ruby
Test Kitchen version 1.21.1
```

## Ruby Version

If you are not using test-kitchen via ChefDK, please provide the output of ruby --version.

```ruby
ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-linux]
```

## Replication Case

Multiple suites run in concurrency,
```
driver name: docker_cli
transport name: docker_cli
provisioner name: chef_zero
serverspec for the tests
```

s-bernard:
Fixes #1416

### DIFF
--- a/lib/kitchen.rb
+++ b/lib/kitchen.rb
@@ -61,6 +61,9 @@ module Kitchen
     # @return [Mutex] a common mutex for global coordination
     attr_accessor :mutex
 
+    # @return [Mutex] a mutex used for Dir.chdir coordination
+    attr_accessor :mutex_chdir
+
     # Returns the root path of the Kitchen gem source code.
     #
     # @return [Pathname] root path of gem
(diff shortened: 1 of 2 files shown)


s-bernard:
I have made the requested changes. Tell me if you need something else.

PR #1431, opened 2018-07-03
using preferred config name for list output

### LINKED ISSUE #1423: Missing config errors point to the old preferred config name
If you try to run `kitchen list` in a directory without a kitchen.yml file you get an error message that mentions it can't find the .kitchen.yml file not the kitchen.yml file. We should point users to the new preferred location.

```
>>>>>> ------Exception-------
>>>>>> Class: Kitchen::UserError
>>>>>> Message: Kitchen YAML file /Users/tsmith/dev/work/chef/.kitchen.yml does not exist.
>>>>>> ----------------------
```

tarcinil:
when config_file doesn't exist, say that kitchen_yml doesn't exist, adding test to check for raised exception when read is used without a config file existing

fixes #1423

### DIFF
--- a/lib/kitchen/loader/yaml.rb
+++ b/lib/kitchen/loader/yaml.rb
@@ -62,7 +62,7 @@ module Kitchen
       # @return [Hash] merged configuration data
       def read
         unless File.exist?(config_file)
-          raise UserError, "Kitchen YAML file #{kitchen_yml} does not exist."
+          raise UserError, "Kitchen YAML file #{config_file} does not exist."
         end
 
         Util.symbolized_hash(combined_hash)
--- a/lib/kitchen/loader/yaml.rb
+++ b/lib/kitchen/loader/yaml.rb
@@ -196,7 +196,7 @@ module Kitchen
           raise UserError, "Both #{kitchen_yml} and #{dot_kitchen_yml} found. Please use the un-dotted variant: #{kitchen_yml}."
         end
 
-        unless File.exist?(kitchen_yml) && File.exist?(dot_kitchen_yml)
+        if !File.exist?(kitchen_yml) && !File.exist?(dot_kitchen_yml)
           return kitchen_yml
         end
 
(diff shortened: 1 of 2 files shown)


coderanger:
So the problem here is that this will give incorrect information in cases where the default config location has been changed. What we should probably do is change `default_config_file` to return `kitchen_yml` if neither are present.

tarcinil:
@coderanger 

I have changed the method for `#default_config_file` so that it is returning `#kitchen_yml` when neither file exists.

The change seems to anger the existing tests. 

```
  1) Error:
Kitchen::Loader::YAML::.initialize#test_0002_when kitchen.yml not present, falls back to .kitchen.yml:
NoMethodError: undefined method `[]' for nil:NilClass
    /Users/vburton/Code/test-kitchen/spec/kitchen/loader/yaml_spec.rb:57:in `block (3 levels) in <top (required)>'

  2) Error:
Kitchen::Loader::YAML::.initialize#test_0007_sets local_config based on Dir.pwd by default:
NoMethodError: undefined method `[]' for nil:NilClass
    /Users/vburton/Code/test-kitchen/spec/kitchen/loader/yaml_spec.rb:96:in `block (3 levels) in <top (required)>'
```

I am going to review the test errors tomorrow and added WIP to the title.

coderanger:
Yeah, I would totally believe the tests are a bit overly brittle right now :)

tarcinil:
Seems interesting.  When using #kitchen_yml as the return, #diagnose return nil and causes those tests to fail.  

I have some ideas, but I’ve been busy.

tarcinil:
Closed and re-opened PR to re-trigger build due to Travis CI issues.

tarcinil:
@coderanger 
This is ready for review.

tas50:
I just recreated a pretty similar fix locally. This one is already open though. Thanks for taking this on. It will avoid some user confusion in the future.

PR #1432, opened 2018-07-05
Improve environment variables in lifecycle hooks

jsirex:
Let populate whole state via KITCHEN_INSTACE_variable. For example,
with vagrant driver you will get the following variables:

* KITCHEN_INSTANCE_HOSTNAME
* KITCHEN_INSTANCE_PORT
* KITCHEN_INSTANCE_USERNAME
* KITCHEN_INSTANCE_SSH_KEY
* KITCHEN_INSTANCE_LAST_ACTION
* KITCHEN_INSTANCE_LAST_ERROR

### DIFF
--- a/lib/kitchen/lifecycle_hooks.rb
+++ b/lib/kitchen/lifecycle_hooks.rb
@@ -91,8 +91,11 @@ module Kitchen
         "KITCHEN_INSTANCE_NAME" => instance.name,
         "KITCHEN_SUITE_NAME" => instance.suite.name,
         "KITCHEN_PLATFORM_NAME" => instance.platform.name,
-        "KITCHEN_INSTANCE_HOSTNAME" => state[:hostname].to_s,
       }
+      # Make state also available through the environment
+      state.each_pair do |k, v|
+        environment["KITCHEN_INSTANCE_#{k.to_s.upcase}"] = v.to_s
+      end
       # If the user specified env vars too, fix them up because symbol keys
       # make mixlib-shellout sad.
       if hook[:environment]


coderanger:
What specific use case are you looking for with this? I would rather be cautious about exposing data since we can never take things back.

jsirex:
I don't see reason you expose KITCHEN_INSTANCE_HOSTNAME, but don't expose KITCHEN_INSTANCE_USERNAME/KITCHEN_INSTANCE_SSH_KEY

There is case when you want to run script locally on host machine but pass to the script connection information about kitchen instance. Somebody may want to upload/download some data during phase

In my particular case I want to integrate terraform and local provisioning with null_resources. So I'm using test-kitchen to spin up services and run terraform to provision. Current kitchen-terraform plugin is not suitable in my case.

But with hooks it can be done very nice:
```yaml
...
lifecycle:
  pre_create:
    - local: ./converge.sh plan
  pre_converge:
    # Disable Internet for testing in air-gapped env, but allow local-net
    # TODO: can we get better solution?
    - remote: sudo ip route del default || /bin/true
    - remote: sudo ip route add 10.0.0.0/8 via 10.0.2.2 || /bin/true
  post_converge:
    - local: ./converge.sh
```

converge.sh:

```bash
#!/bin/bash

WDIR=$(cd $(dirname $0); pwd)
TERRAFORM_DIR=$WDIR/terraform
STATE_FILE=$WDIR/.kitchen/$KITCHEN_INSTANCE_NAME.tfstate
ACTION=${1:-apply}

tf_init() {
    terraform init $TERRAFORM_DIR
}

tf_default_args(){
    echo  -input=false \
          -state $STATE_FILE \
          -var "ssh_user=$KITCHEN_INSTANCE_USERNAME" \
          -var "ssh_private_key=$KITCHEN_INSTANCE_SSH_KEY" \
          -var "ssh_host=$KITCHEN_INSTANCE_HOSTNAME" \
          -var "ssh_port=$KITCHEN_INSTANCE_PORT" \
          -target "module.$KITCHEN_SUITE_NAME"
}

tf_plan() {
    terraform plan $(tf_default_args) $TERRAFORM_DIR
}

tf_apply() {
    terraform apply -auto-approve $(tf_default_args) $TERRAFORM_DIR
}

echo "Dropping previous state file to force reconverge"
rm -f $STATE_FILE

# MAIN
tf_init
tf_$ACTION
```

coderanger:
That sound waaaaay outside of the scope of Test Kitchen. This is not a generic machine provisioning harness.

jsirex:
I just showed one particular use-case. It fits test-kitchen requirements well: I can specify platforms where to test terraform modules, suites - what to test, verifier aka inspec - validate.

Let forget for a while about my use-case. You already exposed part of state by https://github.com/test-kitchen/test-kitchen/blob/master/lib/kitchen/lifecycle_hooks.rb#L94
So why don't to expose all state hash?

Without this patch you know suite name, hostname to connect to (usually localhost). But there is no way to get username, port, and private key to use to be able to connect. Let say I want to upload some files from my local machine at `pre_converge` phase. This patch will help me. Or I want at the end of verify phase download reports and store them somewhere locally. Hooks are great here too.

coderanger:
Things like username and port don't generally change from one suite to another, or at least are hardwired based on external information (like which OS image is being used). The hostname, however, is fully dynamic in almost all drivers. Uploading files to the target is already supported by putting them under the `data/` folder, and downloading things at the end of the run is supported by the `downloads` configuration key. The reason to not dump the whole state hash into the env is mostly that the state system was never meant to be user-facing and many of the keys in there are effectively private variables. Exposing them means that someone is going to use them, which means more unhappy users when they inevitably change.

coderanger:
(specifically adding the port might be reasonable since I think there's at least one driver where that is dynamic, but even that is pushing it :)

jsirex:
All points sound reasonable. So you can close this PR or modify it in any way.

PS. I can't easily find any documentation about `data/` or `downloads` configuration key.

cheeseplus:
We definitely need to fix some docs issues for downloads/data but I'll close this PR out as @coderanger notes and have opened https://github.com/test-kitchen/kitchen-docs/issues/85

PR #1435, opened 2018-07-11
Catch 'Operation already in progress' as seen on Ubuntu on WSL

bdwyertech:
Issue:
```
Waiting for SSH service on 10.0.0.165:22, retrying in 3 seconds
Waiting for SSH service on 10.0.0.165:22, retrying in 3 seconds
Waiting for SSH service on 10.0.0.165:22, retrying in 3 seconds
Errno::EALREADY: Operation already in progress - connect(2) for 10.0.0.165:22
```

Adding `Errno::EALREADY` to the rescue'ables for retry seems to make it go away.

Probably some type of race condition -- [This issue seems very similar.](https://github.com/celluloid/celluloid/issues/436)

Signed-off-by: Brian Dwyer <brian.dwyer@broadridge.com>

### DIFF
--- a/lib/kitchen/transport/ssh.rb
+++ b/lib/kitchen/transport/ssh.rb
@@ -229,7 +229,7 @@ module Kitchen
         PING_COMMAND = "echo '[SSH] Established'".freeze
 
         RESCUE_EXCEPTIONS_ON_ESTABLISH = [
-          Errno::EACCES, Errno::EADDRINUSE, Errno::ECONNREFUSED, Errno::ETIMEDOUT,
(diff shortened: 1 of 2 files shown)


cheeseplus:
Could we get the tests updated as well https://github.com/test-kitchen/test-kitchen/blob/69f549281ec6487119764346d94b142607ee29be/spec/kitchen/transport/ssh_spec.rb?

bdwyertech:
Sure, I'll take a look later or Monday @cheeseplus

bdwyertech:
Tests cases updated @cheeseplus -- Also found another error class missing so added that one too.

PR #1459, opened 2018-09-07
Middleman -> Hugo conversion

cheeseplus:
This transitions the site to Hugo from Middleman which lines up with other Chef software projects.

The front page remains identical to what the Middleman site offered at the cost of being raw HTML instead of an HTML slim template.
screenshot 2018-09-19 14 46 19

The docs themselves look have been re-organized with better and more appropriate navigation menu entries
screenshot 2018-09-19 14 49 11

The Getting Started section has been ported relatively unmodified but future plans are to radically simplify the section and move relevant content under Reference.


cheeseplus:
Now with a custom theme and not breaking the build. Should be nearly ready to go once infra updates go through.

PR #1475, opened 2018-10-07
handle net-ssh >= 5.0

### LINKED ISSUE #1474: test-kitchen does not support net-ssh > 5.0
## Description

`test-kitchen` cannot be included in a project using `net-ssh` with version > 5.0

## Kitchen Version

```Shell
$ bundle exec kitchen --version
Test Kitchen version 1.23.2
```

_I did not use **ChefDK**_.

## Ruby Version

```Shell
$ chruby
* ruby-2.5.1
$ ruby --version
ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-linux]
```

## Platform Version

I'm running Debian stretch:

``Shell
$ cat /etc/debian_version 
9.5
``

## Replication Case

Try a `bundle` command with this `Gemfile`
```Ruby
gem 'test-kitchen', require: false
gem 'net-ssh', '>= 5.0', require: false
```

Result:
```Shell
Resolving dependencies...
Your Gemfile has no gem server sources. If you need gems that are not already on your machine, add a
line like this to your Gemfile:
source 'https://rubygems.org'
Bundler could not find compatible versions for gem "net-ssh":
  In Gemfile:
    net-ssh (>= 5.0)

    test-kitchen was resolved to 1.23.2, which depends on
      net-ssh (< 5.0, >= 2.9)
```

Val:
Fix #1474

Signed-off-by: Laurent Vallar <val@zbla.net>

### DIFF
--- a/lib/kitchen/transport/ssh.rb
+++ b/lib/kitchen/transport/ssh.rb
@@ -508,7 +508,8 @@ module Kitchen
       #
       # Returns the correct host-key-verification option key to use depending
       # on what version of net-ssh is in use. In net-ssh <= 4.1, the supported
-      # parameter is `paranoid` but in 4.2, it became `verify_host_key`
+      # parameter is `paranoid` but in 4.2, it became `verify_host_key`.
+      # After 5.0 it is now `never`.
       #
       # `verify_host_key` does not work in <= 4.1, and `paranoid` throws
       # deprecation warnings in >= 4.2.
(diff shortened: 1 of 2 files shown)


Val:
@coderanger : You did close #1474 a bit quickly. Could you please reconsider ?

Val:
I will provide an use case and a proper fix. Closing.

coderanger:
The issue doesn't need to stay open for you to file a PR, the two are unrelated. Issues are for bug reports, PRs are for contributions.

PR #1529, opened 2019-02-26
Rename .kitchen.yml -> kitchen.yml

nqb:


tas50:
You found another one. Thanks @nqb

oscar123mendoza:
This is the worst change, ever. Now I have to go through and rename every `.kitchen.yml` file with no added benefit.

tas50:
@oscar123mendoza the existing .kitchen.yml files continue to work. You're free to keep using those

PR #1535, opened 2019-03-20
Allow the newly released net-scp

tas50:
This allows us to bring in net-ssh 5.x as well since net-scp 1.1 pinned to 4.0


tas50:
Looks like I created the same PR twice. Oops

PR #1538, opened 2019-04-05
Link to the docs page with log_level information

tas50:
This constantly burns me. Make it easier to find via google


tas50:
Thanks @cheeseplus

PR #1542, opened 2019-04-17
Require gems needed for ed25519 ssh key support 

tas50:
net-ssh 5 includes support for ed25519 without the need for the rbnacl
mess, but it still requires additional gems to make it work. See https://github.com/net-ssh/net-ssh/issues/565


damacus:
I guess this means I have to close out all those issues I opened (will do when I'm on desktop!)

Nice one

PR #1544, opened 2019-04-25
Chef provisioners should prompt for license acceptance

tyler-ball:
The provisioner itself is not protected under the Chef Software EULA. But the product it installs on provisioned machines (Chef Client, ChefDK, etc.) is. So we perform a workstation-side early check to ensure the current user has accepted the license. The acceptance flow is the same as it is for the clients. Acceptance is automatically passed to provisioned instances.

To accept the license the user can:
* Set the CHEF_LICENSE environment variable
* Set the 'chef_license' attribute in the kitchen.yml under the provisioner
* If these are not set and the license has not been previously accepted an interactive prompt will occur asking the user to accept.

Possible values for acceptance are "accept" or "accept-no-persist".

License acceptance is stored on the workstation (unless the user sets "accept-no-persist"). Once stored the license acceptance does not need to be accepted as a flag.

Depends on https://github.com/chef/license-acceptance/pull/36

### DIFF
--- a/test-kitchen.gemspec
+++ b/test-kitchen.gemspec
@@ -23,17 +23,18 @@ Gem::Specification.new do |gem|
 
   gem.required_ruby_version = ">= 2.3"
 
-  gem.add_dependency "mixlib-shellout", ">= 1.2", "< 3.0"
-  gem.add_dependency "net-scp",         ">= 1.1", "< 3.0" # pinning until we can confirm 3+ works
-  gem.add_dependency "net-ssh",         ">= 2.9", "< 6.0" # pinning until we can confirm 6+ works
-  gem.add_dependency "net-ssh-gateway", ">= 1.2", "< 3.0" # pinning until we can confirm 3+ works
-  gem.add_dependency "ed25519",         "~> 1.2" # ed25519 ssh key support
-  gem.add_dependency "bcrypt_pbkdf",    "~> 1.0" # ed25519 ssh key support
-  gem.add_dependency "thor",            "~> 0.19"
-  gem.add_dependency "mixlib-install",  "~> 3.6"
-  gem.add_dependency "winrm",           "~> 2.0"
-  gem.add_dependency "winrm-elevated",  "~> 1.0"
-  gem.add_dependency "winrm-fs",        "~> 1.1"
+  gem.add_dependency "mixlib-shellout",    ">= 1.2", "< 3.0"
+  gem.add_dependency "net-scp",            ">= 1.1", "< 3.0" # pinning until we can confirm 3+ works
+  gem.add_dependency "net-ssh",            ">= 2.9", "< 6.0" # pinning until we can confirm 6+ works
+  gem.add_dependency "net-ssh-gateway",    ">= 1.2", "< 3.0" # pinning until we can confirm 3+ works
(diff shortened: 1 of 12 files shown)


test-kitchen.gemspec
```diff
@@ -23,17 +23,18 @@ Gem::Specification.new do |gem|
...
+  gem.add_dependency "net-ssh",            ">= 2.9", "< 6.0" # pinning until we can confirm 6+ works
+  gem.add_dependency "net-ssh-gateway",    ">= 1.2", "< 3.0" # pinning until we can confirm 3+ works
+  gem.add_dependency "ed25519",            "~> 1.2" # ed25519 ssh key support
+  gem.add_dependency "bcrypt_pbkdf",       "~> 1.0" # ed25519 ssh key support
+  gem.add_dependency "thor",               "~> 0.19"
+  gem.add_dependency "mixlib-install",     "~> 3.6"
+  gem.add_dependency "winrm",              "~> 2.0"
+  gem.add_dependency "winrm-elevated",     "~> 1.0"
+  gem.add_dependency "winrm-fs",           "~> 1.1"
+  gem.add_dependency "license-acceptance", "~> 0.2"
```
tas50: it would be useful to add a simple comment that we need this for chef 15+ license acceptance

Annih:
@tyler-ball maybe I missed something but it doesn't seem that chef-zero/chef-solo/chef-apply accept a `--chef-license` argument. How is it supposed to work?

Since I bumped to TK2.2.0 my kitchen tests are failing with the following error:
> `parse_options': invalid option: --chef-license (OptionParser::InvalidOption)

N.B. I'm using an old version of chef, maybe this is the cause, but I didn't see anything like that in the chef master branch

cheeseplus:
@Annih the license flag will only apply to Chef 15+ and doesn't exist on anything prior

tas50:
We added a dummy flag to 14.12, but Chef 15 will require the flag. Are you specifying that directly in your kitchen.yml? This may be a bug but I'd need to know what your config looks like.

Annih:
I'm not specifying anything but at somepoint test-kitchen run the `license-acceptance` code and detect that I must accept it. I think this is related to the fact I'm not complying with the new kitchen settings (as per https://github.com/chef-boneyard/chef-rfc/blob/master/rfc091-deprecate-kitchen-settings.md)

Here is my `provisioner` config:
```yaml
provisioner:
  name: chef_zero
  install_msi_url: https://internal.url/chef-client.msi
  chef_omnibus_url: https://internal.url/omnibus_install.sh
  chef_license: false # this is a test
  client_rb:
    rubygems_url: https://internal.url
```

IMHO that's why `product_name` defaults to `chef` and `product_version` to `latest`. Adding some debug to kitchen I got this:
```D      License acceptance required for chef version: latest. Prompting
D      Searching for the following licenses: ["chef-infra", "inspec"]
D      Missing licenses remaining: ["chef-infra", "inspec"]
```

I know I'm late on chef versions/features but I would expect a minor release to be less breaking or at least more documented. Would you like me to create issue(s) about that?

PS. For some weird reason I have been asked first for "InSpec" (which I was not using - doing only a converge test - but OK), then I have been asked for "Chef-infra".
When I reset everything (kitchen + accepted licenses), I have been asked for both - which seems more normal to me.

PR #1546, opened 2019-04-30
Chef license checking should work with legacy require_chef_omnibus config

tas50:
We'd prefer if people would stop using this old config, but it's used in a lot of places. This adds a new method that determines where a person specified the version of chef they'd like to provision. Without this a person using require_chef_omnibus set to 15 or later will not receive a license prompt since we interpret that as a nil version that doesn't require a license.


tyler-ball:
LGTM so far

PR #1547, opened 2019-05-02
Moving Chef license acceptance to Chef config instead of command line argument

tyler-ball:
This is less conflicting with consumer libraries. It is also backwards
compatible with previous Chef versions (even though we check that).

Signed-off-by: tyler-ball <tball@chef.io>


tas50:
@tyler-ball specs are failing for this

tyler-ball:
@tas50 Specs fixed!

PR #1552, opened 2019-05-13
Let license errors raise without trying to cleanup sandbox

tyler-ball:
Before this change if the license acceptance failed and raised an
exception the 'cleanup_sandbox' was called in an ensure. But the
sandbox would not have been created yet so it would fail with an error
like:

Failed to complete #converge action: [Sandbox directory has not yet been
created. Please run Kitchen::Provisioner::ChefGithub#create_sandox
before trying to access the path.]

Signed-off-by: tyler-ball <tball@chef.io>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,7 +48,7 @@ matrix:
 
       script:
         - bundle exec kitchen --version
-        - git clone https://github.com/chef/proxy_tests.git
+        - git clone --branch license_acceptance https://github.com/chef/proxy_tests.git
         - rvmsudo -E bundle exec bash $PROXY_TESTS_DIR/run_tests.sh kitchen \* \* /tmp/out.txt
       after_script:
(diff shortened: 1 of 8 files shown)


.travis.yml
```diff
@@ -18,6 +18,7 @@ bundler_args: --with integration --without changelog debug docs
 script:
   - bundle exec rake
   - export KITCHEN_YAML=kitchen.dokken.yml
+  # - export CHEF_LICENSE=accept-no-persist
```
tas50: Looks like tests are failing without this

tyler-ball: :doh: this was a local change I made to test and checked it in

tyler-ball:
Down to just the proxy and appveyor tests failing

tas50:
Appveyor will pass when the new Chef 15 client ships later today

PR #1569, opened 2019-07-01
Add upload support to provisioner

tecracer-theinen:
In case of air-gapped or proxy-enforced environments, sometimes there is the need to upload installers to kitchen instances from the developer's machine. This patch replicates the `downloads` functionality of the base provisioner to include an `uploads` functionality as well.


tas50:
Thanks @tecracer-theinen. This is very cool to have now

PR #1571, opened 2019-07-10
Add Azure Pipelines CI support

smurawski:
Adds a `azure-pipelines.yml` and tweaks some of the test helpers to get tests running on the hosted build agents.

Sets up builds across Windows, Mac, and Linux and replicates the tests that are happening in Travis and AppVeyor.

Once the PR is merged, we can wire it up to Azure DevOps (along side the other CI environments) to compare behaviors/times.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,115 @@
+---
+trigger:
+  batch: true
+  branches:
+    include:
+    - master
+    - 1-stable
+pr:
+  autoCancel: true
+  branches:
+    include: 
+      - master
+      - 1-stable
+
+variables:
+  CHEF_LICENSE: accept-no-persist
+
+stages:
+  - stage: Build
+    jobs:
+      - job: Validate
+        strategy:
+          matrix:
+            Windows_Ruby25:
+              version: 2.5.3
+              imageName: 'windows-2019'
+              task: windows_ci.bat
+            Windows_Ruby26:
+              version: 2.6.1
+              imageName: 'windows-2019'
+              task: windows_ci.bat
+            Linux_Ruby25:
+              version: 2.5.5
+              imageName: 'ubuntu-16.04'
+              task: linux_ci.sh
+              KITCHEN_YAML: kitchen.dokken.yml
+            Linux_Ruby26:
+              version: 2.6.2
+              imageName: 'ubuntu-16.04'
+              task: linux_ci.sh
+              KITCHEN_YAML: kitchen.dokken.yml
+            Mac_Ruby25:
+              version: 2.5.5
+              imageName: 'macOS-10.13'
+              task: mac_ci.sh
+            Mac_Ruby26:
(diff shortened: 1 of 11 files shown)


azure-pipelines.yml
```diff
@@ -0,0 +1,115 @@
...
+              machine_port: 5985
+              KITCHEN_YAML: kitchen.appveyor.yml
+            Proxy_Integration:
+              version: 2.5.5
+              imageName: 'ubuntu-16.04'
+              task: proxy_ci.sh
+              MACHINE_USER: kitchen
+              MACHINE_PASS: K1tch3nY@ml!
+              MACHINE_PORT: 22
+              KITCHEN_YAML: ../../../../../../../kitchen.proxy.yml
```
stuartpreston: Is there a more descriptive way of describing this path? (presume the root of the repo - so $(Build.SourcesDirectory) ?

support/ci/proxy_ci.sh
```diff
@@ -0,0 +1,38 @@
...
+sudo -E usermod -aG sudo $MACHINE_USER
+
+# Make sure SSH can run on Azure DevOps hosted agent.
+sudo mkdir -p /var/run/sshd
+sudo service ssh restart
+
+# Install squid and git.
+sudo apt-get update
+sudo apt-get -y install squid3 git curl 
+git clone https://github.com/smurawski/proxy_tests.git
```
stuartpreston: Is this a permanent location?

support/ci/proxy_ci.sh
```diff
@@ -0,0 +1,38 @@
...
+
+# Add the ruby path to the secure path, so we can run the tests elevated (to manage the squid service)
+sudo echo 'Defaults	secure_path="/opt/hostedtoolcache/Ruby/2.5.5/x64/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin"' | sudo tee /etc/sudoers.d/kitchen
```
stuartpreston: Is there any way of making this so that when Ruby 2.5.5 disappears from the hosted agent it won't break this script?  Can $(Agent.ToolsDirectory) help, can the **Use Ruby Version** task set the secure path instead?

support/ci/proxy_ci.sh
smurawski: I had to make some updates to that repo given how the package installs on the build agent.  I'd be happy to push the changes back upstream, but I'm not sure what else depends on it.

support/ci/proxy_ci.sh
smurawski: @stuartpreston I'm following up with the Azure Pipelines team on that very thing. :)

azure-pipelines.yml
smurawski: I'll try that - I tried an absolute path before, but I don't think kitchen liked it.

PR #1577, opened 2019-07-25
adding lifecycle to instance diagnose

tarcinil:
This adds instance-level lifecycle hooks into the instance diagnostic hash.


tarcinil:
Gist of `kitchen diagnose`
https://gist.github.com/tarcinil/34e0ce3786dca1710f87736afb659d41

PR #1614, opened 2020-01-15
Add instance name to download path when downloading files

vexx32:
If the same download path is used for all instances, it may prove problematic, especially if it is configured to download specific files; the file would be overwritten and the data lost.

Adding the instance name to the path allows the feature to be used by multiple instances in succession without potentially overwriting files.

/cc @tas50 @smurawski @gep13

### DIFF
--- a/lib/kitchen/verifier/base.rb
+++ b/lib/kitchen/verifier/base.rb
@@ -78,7 +78,7 @@ module Kitchen
 
           info("Downloading files from #{instance.to_str}")
           config[:downloads].to_h.each do |remotes, local|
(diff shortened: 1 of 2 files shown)


vexx32:
I think that's all the tests passing... the rest of the CI appears to be failing much earlier in the chain, so I don't think it's related to my changes. Feel free to correct me if I'm wrong! 😊 💖

vexx32:
From what I can see from the logs, I don't think the CI error is related to this PR. Give me a ping if anything more is needed here. 🙂

lamont-granquist:
this is probably the correct thing to do, but in the event someone has figured out this works fine for single instances they may be happily using it and this would break any scripts they have build wrapping it.

PR #1617, opened 2020-01-19
Store the installer script in /var/tmp

tomhughes:
On a systemd system the /tmp directory is mounted with noexec which prevents the installer script being executed but there is no similar restriction for the /var/tmp directory.


stevenoneill:
Hmmmm

tomhughes:
The other way round it is to use `/bin/sh /tmp/chef-installer.sh` but that requires assuming that it's a shell script and strictly speaking that routine could be given anything that could be executed at the moment.

lamont-granquist:
/var/tmp should probably be the default given the directions that the usage of /tmp vs /var/tmp is going, but this absolutely needs to be a configurable and not hard coded.

tas50:
@tomhughes I agree with Lamont here. It would be great to get out of temp as it causes a lot of pain on Fedora and in restricted environments, but we need to make this a configuration option.

tomhughes:
To be honest this dates from when I was experimenting with running full system containers with kitchen-docker before I discovered kitchen-dokken and switched to that which has avoided the problem.

damacus:
Looks like this was fixed via https://github.com/test-kitchen/test-kitchen/pull/1662/ ?

Issue #923, opened 2016-01-24
chef_omnibus_root is nil on images pre-baked with a chef client


aespinosa:
On test-kitchen 1.5.0, the chef_client_path is set to /bin/chef-client if the chef_zero provisioner doesn't have to perform an omnibus install.

Here's my `kitchen diagnose` : https://gist.github.com/aespinosa/00d81f95d1a9e060c51d

zl4bv:
Experiencing this error too.

Relevant section from the log:
```
D      Transfer complete
D      [WinRM] plaintext::http://ip-xxx-xxx-xxx-xxx.ap-southeast-2.compute.internal:5985/wsman<{:user=>"administrator", :pass=>"************", :disable_sspi=>true, :basic_auth_only=>true}> ($env:PATH = [System.Environment]::GetEnvironmentVariable("PATH","Machine")

& \bin\chef-solo.bat --config $env:TEMP\kitchen\solo.rb --log_level debug --force-formatter --no-color --json-attributes $env:TEMP\kitchen\dna.json)
$$$$$$ &amp; : The term '\bin\chef-solo.bat' is not recognized as the name of a cmdlet,
$$$$$$ function, script file, or operable program. Check the spelling of the name, or
$$$$$$ if a path was included, verify that the path is correct and try again.
$$$$$$ At line:9 char:3
$$$$$$ + &amp; \bin\chef-solo.bat --config $env:TEMP\kitchen\solo.rb --log_level debug
$$$$$$ --forc ...
$$$$$$ +   ~~~~~~~~~~~~~~~~~~
$$$$$$     + CategoryInfo          : ObjectNotFound: (\bin\chef-solo.bat:String) [],
$$$$$$    CommandNotFoundException
$$$$$$     + FullyQualifiedErrorId : CommandNotFoundException
D      Cleaning up local sandbox in /tmp/default-windows-2012r2-sandbox-20160128-10604-143gxar
-----> Destroying <default-windows-2012r2>...
```

zl4bv:
Aside from when the user explicitly defines it in their config, `chef_omnibus_root` is only ever set for the Chef provisioners on [this line](https://github.com/test-kitchen/test-kitchen/blob/master/lib/kitchen/provisioner/chef_base.rb#L145), which does not get executed unless `require_chef_omnibus` is truthy.

[...]

PR #1619, opened 2020-01-30
adding defaults for chef_omnibus_root for windows and unix-like/linux systems

tarcinil:
This work patches #923 so that the issue will stop bugging people, and we can discuss implementing mixlib-install but that has downstream affects against cinc.

### DIFF
--- a/lib/kitchen/provisioner/chef_base.rb
+++ b/lib/kitchen/provisioner/chef_base.rb
@@ -44,6 +44,13 @@ module Kitchen
     class ChefBase < Base
       default_config :require_chef_omnibus, true
       default_config :chef_omnibus_url, "https://omnitruck.chef.io/install.sh"
+      default_config :chef_omnibus_root do |provisioner|
+        if provisioner.powershell_shell?
+          "/opt/chef" # I know that we don't need this, but it leaves things flexible.
+        else
+          "/opt/chef"
+        end
+      end
       default_config :chef_omnibus_install_options, nil
       default_config :chef_license, nil
       default_config :run_list, []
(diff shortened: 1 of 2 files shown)


lib/kitchen/provisioner/chef_base.rb
```diff
@@ -44,6 +44,13 @@ module Provisioner
     class ChefBase < Base
       default_config :require_chef_omnibus, true
       default_config :chef_omnibus_url, "https://omnitruck.chef.io/install.sh"
+      default_config :chef_omnibus_root do |provisioner|
+        if provisioner.powershell_shell?
+          "/opt/chef" # I know that we don't need this, but it leaves things flexible.
```
robbkidd: Any reason not to set this to the Windows default of `"$env:systemdrive\\opscode\\chef"`?

tarcinil: I based that on content that I had seen inside of the spec tests.

spec/kitchen/provisioner/chef_base_spec.rb#L604 which looks to expect the same path as unix-like/linux systems.  I did contemplate using `provisioner.windows_os?` to `"$env:systemdrive\\opscode\\chef"` entry.

tarcinil:
As I am setting the default, anything done in the later functions "should" overload the default_config.  I have been thinking adding additional unit tests that would overload this using lets say Cinc as an example.

lib/kitchen/provisioner/chef_base.rb
lamont-granquist: that should be c:\opscode\chef though not /opt/chef for windows

PR #1623, opened 2020-01-31
Pass in all local env vars to the instance

### LINKED ISSUE #1622: Feature - Enable pushing in Environment variables
## Description

Currently the contents of the `CI` environment variable is pushed into the created virtual machine from the host, while this works well it is limiting in terms of what you can push in

I would like to propose that any environment variables that start with `KITCHEN_` is pushed into the created VM the same way the `CI` environment variable is today, but with the `KITCHEN_` stripped off, this would make handling the chef license easy for example, `KITCHEN_CHEF_LICENSE=accept` becomes `CHEF_LICENSE=accept` without having to update every kitchen file.

Xorima:
## Description

This commit will add the ability to push in additional env vars from the host to the client, it will push all env vars which start KITCHEN_ to the machine

Signed-off-by: Jason Field <jason@avon-lea.co.uk>

## Issues fixed

#1622

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -225,7 +225,7 @@ stages:
         steps:
           - task: UseRubyVersion@0
             inputs:
-              versionSpec: 2.6.3
+              versionSpec: 2.6.5
               addToPath: true
           - script: |
               gem install bundler --quiet
--- a/lib/kitchen/configurable.rb
+++ b/lib/kitchen/configurable.rb
@@ -349,8 +349,8 @@ module Kitchen
       code_parts = resolve_proxy_settings_from_config
       code_parts << shell_env_var("TEST_KITCHEN", 1)
       code_parts << shell_env_var("CI", ENV["CI"]) if ENV["CI"]
-      ENV.select { |key, value| key.start_with?("KITCHEN_") }.each do |key, value|
-        env_var_name = "#{key}".sub!("KITCHEN_", "")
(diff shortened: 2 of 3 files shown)


lamont-granquist:
You can't use that as the prefix, there's already variables like `KITCHEN_YAML` which occupy that namespace, at a minimum that is confusing, and the side effect of having those variables all injected into the virt may cause problems in the future.

It needs to be relatively clear from the prefix that they are env vars which are being deliberately injected into the client, just from the name, in order to reduce confusion.

Xorima:
Hey,

Thanks for the feedback, totally forgot about those ones. 

Do you have any opinions on what the prefix should be @lamont-granquist, I personally am just after the feature so I can push in the `chef_license` env var (it is not pushed currently) and some very odd proxy settings, but also it may be more flexible than the current `ci` push. 

Some of my initial thoughts, 

- `kitchen_envpush` - this is at least very clean, but it is also very long named?
- `tk_vmvar` - this uses a different namespace and says it's a vm variable, but is that really clear enough?
-`ci_` - we already push in the `ci` env var so this would be similar behaviour (except we would be removing the `ci_` prefix) however the behaviour still not clear from the name

Xorima:
Make it `tk_` - tim

lamont-granquist:
`TKENV_FOO` maybe?

doesn't collide, has "env" in it, has "tk" in it, but isn't obnoxiously long to type.

Xorima:
done @lamont-granquist

Xorima:
Where should I doc these new features @tas50 ?

PR #1625, opened 2020-01-31
I like green builds

Xorima:
Signed-off-by: Jason Field <jason@avon-lea.co.uk>

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -90,6 +90,13 @@ stages:
             displayName: Check ruby exists
             env:
               SPEC_OPTS: --format progress
+
+          - script: |
+              dir /s /b C:\hostedtoolcache\windows\Ruby\
+            displayName: Check for Ruby
+            env:
+              SPEC_OPTS: --format progress
+
           - script: |
               C:\hostedtoolcache\windows\Ruby\2.5.5\x64\bin\ruby.exe "d:/a/1/.bundle/ruby/2.5.0/bin/rake"
             displayName: run the rake.bat by hand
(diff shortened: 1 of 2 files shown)


Xorima:
No idea how to fix this, it appears to be failing at this command:
```
              bundle exec rake unit && bundle exec rake quality
```
Help would be appreciated?

Xorima:
So on further inspection it cannot find the rake command in bundle exec at all, 

comparing the `bundle install` between this and the previous version the only difference is as follows:

|Green|Red|
|---|---|
|Using mime-types 3.3.1| Not found at all|
|Using pry-byebug 3.7.0| Fetching pry-byebug 3.8.0 Installing pry-byebug 3.8.0|

I can't see why this would stop the it from finding the rakefile, help would be appreciated

Xorima:
it appears that ruby is no longer where rake is expecting to find it: <https://dev.azure.com/test-kitchen/test-kitchen/_build/results?buildId=312&view=logs&j=a7cf9f6c-a724-5d40-b4d2-4e4eb95d75da&t=386653f8-6c12-5d7f-2c0f-9d176bfacf61> 
If you look at the Cat the rakefile this should be a dir of where ruby.exe is, but it doesn't show up so that seems to be the error,  
I had an error around Ruby 2.5.5 for the mac build, fix in this commit: https://github.com/test-kitchen/test-kitchen/pull/1625/commits/24b0c06dfad89a6217fdfc1341d8a87422f249fb 
maybe it's similar?

tas50:
Thanks @Xorima

PR #1626, opened 2020-02-19
[WIP] consider root_path given under provisioner option

Nimesh-Msys:
This is an in-progress PR and some fixes are needed.

Changes done so far:
- Passing tmp_dir and TMPDIR options to mixlib-install [WIP]
- The option would then reflect under [helpers.sh](https://github.com/chef/mixlib-install/blob/master/lib/mixlib/install/generator/bourne/scripts/helpers.sh.erb#L324) script.
- Other changes includes fixes done in addition to a previous [PR#1369](https://github.com/test-kitchen/test-kitchen/pull/1369) which was reverted due to some issue.
- Possible issue could be:  "sudo cat with redirect" creates an ambiguity for shell and so it raises permission denied error. An alternative is to use tee with pipe (|).
- There is a related discussion for the [reference](https://stackoverflow.com/questions/10134901/why-sudo-cat-gives-a-permission-denied-but-sudo-vim-works-fine).
- Stucked at passing params to mixlib-install's helper.sh
- Testing under progress.

Signed-off-by: Nimesh-Msys <nimesh.patni@msystechnologies.com>

Issues Resolved:
#1364 
[#customer-bugs#4](https://github.com/chef/customer-bugs/issues/4)

### DIFF
--- a/lib/kitchen/provisioner/chef_base.rb
+++ b/lib/kitchen/provisioner/chef_base.rb
@@ -509,6 +509,8 @@ module Kitchen
           channel: config[:channel].to_sym,
           install_command_options: {
             install_strategy: config[:install_strategy],
+            tmp_dir: config[:root_path],
+            "TMPDIR" => config[:root_path],
           },
         }.tap do |opts|
           opts[:shell_type] = :ps1 if powershell_shell?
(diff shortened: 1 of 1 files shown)


tyler-ball:
We completed this work in https://github.com/test-kitchen/test-kitchen/pull/1662 - thanks for getting it started!

PR #1634, opened 2020-04-01
include ohai directory in cookbooks

SAPDanJoe:
This changes allows test-kitchen to work with the documented placement of ohai plugins: https://docs.chef.io/cookbooks/#components


lamont-granquist:
dunno why azure is red in places but i doubt its this PR.

Issue #1265, opened 2017-08-28
powershell arguments to shell provisioner


bad-beat:
## Description

Issue [#917](https://github.com/test-kitchen/test-kitchen/issues/917) and PR [#943](https://github.com/test-kitchen/test-kitchen/pull/943) adds the ability to pass arguments to shell provisioner.

I added the arguments in my yaml file as shown below. I use the same arguments list for both linux and windows machines.
`provisioner: name: shell privileged: true arguments: '-version 01020304'`

In bootstrap.sh the arguments are concatenated after the script and they are passed along just fine.

In bootstrap.ps1 the arguments are concatenated after the script, but the whole set of script and arguments are encapsulated by double quotes. This is done by this code `code = powershell_shell? ? %{& "#{script}"} : sudo(script)` from lib/kitchen/provisioner/shell.rb on line 82.

By adding the arguments in my yaml file, the bootstrap.ps1 is no longer executed at all on the windows machine. When I change the shell.rb code so it does not encapsluate the script and arguments in double quotes, the arguments are passed along just fine.

## Kitchen Version
```ruby
Test Kitchen version 1.17.0
```

## ChefDK Version
```ruby
not installed
```

## Ruby Version
```ruby
ruby 2.3.4p301 (2017-03-30 revision 58214) [x86_64-linux]
```

## Platform Version
ubuntu-14.04-LTS-trusty-64bit

## Replication Case
I added the arguments in my yaml file as shown below. I use the same arguments list for both linux and windows machines:
`provisioner: name: shell privileged: true arguments: '-version 01020304'`

Now create a linux and windows machine and converge both. The linux bootstrap.sh is executed and shows the arguments. The windows bootstrap.ps1 is not executed at all and does not show the arguments.

It can also be replicated by just calling a powershell script with arguments from another powershell script and add or remove double quotes: `"command -version 01020304"` vs `command -version 01020304`

## Kitchen Output
```ruby
Not available at the moment of creating the issue.
It is however easily replicated and the error message of Windows shows that the command can not be found. Windows sees the double quoted string as a command instead of as a command with arguments and therefore cannot find the command.
```

## Kitchen Diagnose
Can be added later if wanted.

PR #1639, opened 2020-04-24
Windows Shell fixes

alanghartJC:
Currently, it's effectively impossible to run powershell scripts with the provisioner, or to pass arguments to powershell provisioners, which this PR fixes.

- Remove quotes from powershell commands, which was grouping
  the script name with the arguments, causing failures.
  e.g: `somescript.ps1 myarg1 myarg2` => `"somescript.ps1 myarg1 myarg2"`
  Powershell would complain that the executable doesn't exist.

- Don't use `Shellwords.join()` for powershell. This was generating
  `\$env:TEMP/somescript.ps1` (escaping the dollar sign) which
  powershell evaluated to `\C:\Users\someuser\...etc` leaving the
  backslash before `C:` which caused it not find the file.

- Update affected tests

### DIFF
--- a/lib/kitchen/provisioner/shell.rb
+++ b/lib/kitchen/provisioner/shell.rb
@@ -106,13 +106,17 @@ module Kitchen
 
         if config[:arguments] && !config[:arguments].empty?
           if config[:arguments].is_a?(Array)
-            script = Shellwords.join([script] + config[:arguments])
+            if powershell_shell?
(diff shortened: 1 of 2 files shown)


alanghartJC:
This should fix https://github.com/test-kitchen/test-kitchen/issues/1265

tas50:
Thanks for the work and the great explanation @alanghartJC

PR #1642, opened 2020-05-04
docs(guides): update list of provisioners

noelmcloughlin:
Updated list of provisioners in test-kitchen docs.


tas50:
Thanks @noelmcloughlin

PR #1662, opened 2020-07-08
Respect root_path when product_name is specified

### LINKED ISSUE #1364: Error executing /tmp/chef-installer.sh

clintoncwolfe:
# Description

Fixes the `root_path` config option so it is respected when the user specifies `product_name`. The specified directory path will be created without sudo, the install script created without privileges, and then executed with sudo. 

Tested with kitchen-vagrant and kitchen-ec2; kitchen-dokken was also verified not to break, though it does not use this code path. kitchen-openstack would also be good to test.

## Issues Resolved

Fixes #1364

This is an adoption of #1626.

Related PRs and Issues:
#1369 was a prior attempt to fix this, reverted in #1403. Reverted because a sudo command caused the folder to be created with root perms, then the download would occur as an unprivileged user, causing it to fail.
#1626 is a WIP PR to solve this, which this PR adopts. This PR uses the mixlib-install API changes, but keep the original script inject method using a heredoc, finding the sudo issue was located elsewhere. This PR also adds an error message.

Closes chef/chef-workstation#1319, an effort-tracking ticket.

## Check List

### DIFF
--- a/lib/kitchen/provisioner/chef_base.rb
+++ b/lib/kitchen/provisioner/chef_base.rb
@@ -510,6 +510,8 @@ module Kitchen
           channel: config[:channel].to_sym,
           install_command_options: {
             install_strategy: config[:install_strategy],
+            tmp_dir: config[:root_path],
+            "TMPDIR" => config[:root_path],
           },
         }.tap do |opts|
           opts[:shell_type] = :ps1 if powershell_shell?
@@ -535,7 +537,6 @@ module Kitchen
             # install.ps1 only supports http_proxy
             prox.delete_if { |p| %i{https_proxy ftp_proxy no_proxy}.include?(p) } if powershell_shell?
           end
-
           opts[:install_command_options].merge!(proxies)
         end)
         config[:chef_omnibus_root] = installer.root
@@ -554,13 +555,12 @@ module Kitchen
       end
 
       def install_from_file(command)
-        install_file = "/tmp/chef-installer.sh"
-        script = ["cat > #{install_file} <<\"EOL\""]
-        script << command
-        script << "EOL"
+        install_file = "#{config[:root_path]}/chef-installer.sh"
+        script = ["mkdir -p #{config[:root_path]}"]
+        script << "echo #{command} | tee #{install_file}"
         script << "chmod +x #{install_file}"
-        script << sudo(install_file)
-        script.join("\n")
+        script << install_file
+        script.map{ |cmd| sudo(cmd) }.join(";\n")
       end
 
       # @return [String] contents of version based install script
--- a/lib/kitchen/provisioner/chef_base.rb
+++ b/lib/kitchen/provisioner/chef_base.rb
@@ -558,7 +558,10 @@ module Kitchen
         install_file = "#{config[:root_path]}/chef-installer.sh"
         script = []
         script << "mkdir -p #{config[:root_path]}"
-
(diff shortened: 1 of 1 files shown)


tas50:
This is great. Can you add `root_path` to the documentation in this repo so that folks know how to work around everything landing on /tmp by default

lib/kitchen/provisioner/chef_base.rb
```diff
@@ -570,7 +577,7 @@ def script_for_omnibus_version
           config[:require_chef_omnibus], powershell_shell?, install_options
         )
         config[:chef_omnibus_root] = installer.root
```
marcparadise, line 579: Do you know offhand where else `:chef_omnibus_root` might be needed? I see it's set here and in `script_for_product`.  I've only seen it checked in `install_options` - but we set `install_options` above,  before we set the `config` value that it references.

lib/kitchen/provisioner/chef_base.rb
```diff
@@ -554,8 +555,14 @@ def omnibus_dir_option
       end
 
       def install_from_file(command)
```
marcparadise, line 557: This change looks right in context, but do you know why the original code doesn't directly execute the mixlib install script when it's available? 

It seems odd that we're executing a script that writes a script and executes it, instead of either - 1) uploading the script to the root path and executing it by name as the install_command, or 2) execute the install script itself directly, instead of writing to file at all.

tyler-ball: TK was originally written before mixlib-install existed - thats my only guess for why this is the way it is

lib/kitchen/provisioner/chef_base.rb
tyler-ball: I see it read in some of the other provisioners to get things like the `chef-apply` or `chef-solo` path

tyler-ball:
@tas50 I added the documentation as a [follow up commit](https://github.com/test-kitchen/test-kitchen/commit/41914a81ae9adb240d80abc7f2f4789a8266ec29)

PR #1664, opened 2020-07-23
Revert "Respect root_path when product_name is specified" due to regression

tas50:
The change in https://github.com/test-kitchen/test-kitchen/pull/1662 is causing failures on windows hosts. Reverting until we can come up with a fix for that.

### DIFF
--- a/lib/kitchen/provisioner/chef_base.rb
+++ b/lib/kitchen/provisioner/chef_base.rb
@@ -556,12 +556,8 @@ module Kitchen
 
       def install_from_file(command)
         install_file = "#{config[:root_path]}/chef-installer.sh"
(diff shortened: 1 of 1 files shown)


marcparadise:
Revert looks fine, but I'm not sure why the new copps are failing against the reverted code?  I don't see those utf-8 encoding lines being re-added in anything reverted.

tas50:
The utf-8 stuff is for a new chefstyle release. I have a PR in the repo to clean those up, but chefstyle should be pinned to the exact release we know works so that we handle cleanup of new cops + the bump in their own PRs and new cop cleanup doesn't leak into other PRs like this.

PR #1675, opened 2020-08-20
Optimize our requires

tas50:
Avoid requiring things that are already defined. Rubygems is very slow at traversing the filesystem.

### DIFF
--- a/lib/kitchen/cli.rb
+++ b/lib/kitchen/cli.rb
@@ -15,7 +15,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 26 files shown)


Xorima:
looks like the tests are totally broken, this will need fixing as well ...

lib/kitchen/cli.rb
```diff
@@ -15,7 +15,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-require "thor"
+require "thor" unless defined?(Thor)
```
Xorima, line 18:
```suggestion
require "thor"
```

Issue #1641, opened 2020-04-29
Support running recipes in lifecycle hooks


Lasering:
## Feature Request
Lifecycle hooks could support running a recipe.

## Scenario
We are creating a chef recipe for a service which talks to a database. This recipe is presented with the necessary credentials to be able to create databases (with a specific name) in a database cluster elsewhere. However in production we don't want to create a new database, but rather connect to an existing one.

A test cookbook is created that depends on the mariadb supermarket cookbook and installs the mariadb clients locally and creates the database. With just this setup the database cluster would accumulate garbage databases over time. The lifecycle pre_destroy hook helps us to solve this problem, but since we can only run commands we will have to duplicate the logic already offered by the mariadb cookbook. A more maintainable approach would be to create a new recipe in the test cookbook that just destroys the database, then ask the pre_destroy hook to invoke that recipe.

tas50:
I don't think this is a place we'd ever go with the lifecycle hooks. The point of the hooks is to provide a dead-simple shell execution that runs before/after. If we had to support a full recipe execution that would be a much more complex environment that would basically be a complete run of the chef-zero provisioner before running the provisioner.

Lasering:
Running chef-zero ends up being invoking the transport (usually ssh) and running a command (usually chef-zero). Its not a dead-simple shell execution, but its also not that far from it.

PR #1679, opened 2020-08-29
Allow includes/excludes filters for hooks

jasonwbarnett:
# Description

This extends the same idea used in suites to be used in lifecycle hooks, specifically the ability to include/exclude hooks based on platform name. This also goes into extracting hooks into their own class to help build a foundation for future things like https://github.com/test-kitchen/test-kitchen/issues/1641, potentially.

## This seems odd, what is the point of this and why can't you just define platform specific hooks in the platform definition?

The use case I'm trying to solve is someone who has complex platforms defined globally in a KITCHEN_GLOBAL_YAML and only defines suites in the KITCHEN_YAML file. In this case I have multiple platforms defined in the KITCHEN_GLOBAL_YAML file and I wish to define a unique pre_destroy hook per platform type (i.e. windows versus linux) in the  KITCHEN_YAML for the specific suite. Because of the way the YAML loader works with `#rmerge`, it can only merge hashes, not arrays (which totally makes sense technically). Because platforms are defined as an array it forces me to duplicate the platform content to all of my cookbooks in the mono repo.

With the feature layed out in the PR it would allow someone to do the following:

```yaml
# KITCHEN_GLOBAL_YAML
---
platforms:
  - name: redhat-7
    ...lots of content
  - name: windows-2012r2
    ...lots of content
  - name: windows-2016
    ...lots of content
```

```yaml
# KITCHEN_YAML
---
suites:
  - name: default
    run_list:
      - recipe[cookbook::default]
    pre_destroy:
      - remote: /opt/app/bin/cmd --delete
        includes:
          - redhat-7
      - remote: C:/app/bin/other-cmd --destroy
        excludes:
          - redhat-7
```

## Issues Resolved

N/A

## Check List

- [x] New functionality has been documented in the README if applicable.

### DIFF
--- a/lib/kitchen/lifecycle_hook/base.rb
+++ b/lib/kitchen/lifecycle_hook/base.rb
@@ -56,7 +56,7 @@ module Kitchen
 
       # @return [Kitchen::StateFile]
       def state_file
-        instance.state_file
+        lifecycle_hooks.state_file
       end
 
       # @return [Array<String>] names of excluded platforms
--- a/lib/kitchen/lifecycle_hook/base.rb
+++ b/lib/kitchen/lifecycle_hook/base.rb
@@ -1,0 +1,78 @@
+module Kitchen
+  class LifecycleHook
+    class Base
+      # @return [Kitchen::LifecycleHooks]
+      attr_reader :lifecycle_hooks
+
+      # return [String]
+      attr_reader :phase
+
+      # return [Hash]
+      attr_reader :hook
+
+      # @param lifecycle_hooks [Kitchen::LifecycleHooks]
+      # @param phase [String]
+      # @param hook [Hash]
+      def initialize(lifecycle_hooks, phase, hook)
+        @lifecycle_hooks = lifecycle_hooks
+        @phase = phase
+        @hook = hook
+      end
+
+      # return [void]
+      def run
+        raise NotImplementedError
+      end
+
+      # @return [TrueClass, FalseClass]
+      def should_run?
+        if !includes.empty?
+          includes.include?(platform_name)
+        elsif !excludes.empty?
+          !excludes.include?(platform_name)
+        else
+          true
+        end
+      end
+
+      # @return [Logger] the lifecycle hooks's logger
+      #   otherwise
+      # @api private
+      def logger
+        lifecycle_hooks.send(:logger)
+      end
+
+      private
+
+      # @return [Kitchen::Instance]
+      def instance
+        lifecycle_hooks.instance
+      end
+
+      # @return [Hash]
+      def config
+        lifecycle_hooks.send(:config)
+      end
+
+      # @return [Kitchen::StateFile]
+      def state_file
+        instance.state_file
+      end
+
+      # @return [Array<String>] names of excluded platforms
+      def excludes
+        @excludes ||= hook.fetch(:excludes, [])
+      end
+
+      # @return [Array<String>] names of only included platforms
+      def includes
+        @includes ||= hook.fetch(:includes, [])
+      end
+
+      # @return [String]
+      def platform_name
+        instance.platform.name
+      end
+    end
+  end
+end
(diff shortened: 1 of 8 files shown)


lib/kitchen/lifecycle_hook/base.rb
```diff
@@ -0,0 +1,78 @@
...
+      end
+
+      # @return [Hash]
+      def config
+        lifecycle_hooks.send(:config)
+      end
+
+      # @return [Kitchen::StateFile]
+      def state_file
+        instance.state_file
```
clintoncwolfe, line 59: When I tested this code, I get an exception thrown on this line because `state_file` is a private method of Kitchen::Instance. I reached here by having a Kitchenfile with a basic top-level lifecycle hook:
```
lifecycle:
  pre_create:
  - local: echo ....... in pre create .........
```

clintoncwolfe (changes requested):
Thanks for your contribution! I love this feature, and I really like the approach you've taken in implementing it.  However, when I tested the code, I encountered an error which needs to be addressed. Thanks!

lib/kitchen/lifecycle_hook/base.rb
jasonwbarnett: ahh... I haven't tested this at all at this point.

Well... using `#send` seems like a hack and probably not recommended, but what do you think could be done here? I'm not really sure given that I don't know test kitchen that well. In other words, I'm looking for some ideas here of the best way to resolve this issue.

clintoncwolfe: Well, I see three options, in order of my preference:
1. The `state_file` object is actually passed in to the initialize method of LifecyleHooks. Previously, it was repeatedly passed down until it reached where it was needed. It's not graceful, but that is the least disruptive approach.
2. Use `#send` to call `state_file`
3. Alter Instance to make state_file public. I'm most reluctant to do that because it is explicitly made private.

jasonwbarnett: Take a look: https://github.com/test-kitchen/test-kitchen/pull/1679/commits/013b9d5075329140510cea72b1c086eb05b17d90

spec/kitchen/lifecycle_hooks_spec.rb
```diff
@@ -27,151 +27,162 @@
...
-        "KITCHEN_SUITE_NAME" => "default",
-        "KITCHEN_PLATFORM_NAME" => "toaster-1.0",
-        "KITCHEN_INSTANCE_HOSTNAME" => "localhost",
-      },
-    })
+    hook = { local: "echo foo", cwd: "test" }
+    config.update(post_create: [hook])
+    expect_local_hook_generated_and_run(:create, hook, non_standard_local_opts: { cwd: "#{kitchen_root}/test" })
     run_lifecycle_hooks
   end
```
clintoncwolfe, line 118: As this change introduces two new options (includes and excludes) for lifecycle hooks, it would be good to add tests that cover those options.

clintoncwolfe (commented):
I've tested the code manually and at this point it seems solid. However I think we should probably add a test to cover the new options, and ideally an update to the lifecycle hooks docs as well - https://github.com/test-kitchen/test-kitchen/blob/master/docs/content/docs/reference/lifecycle-hooks.md

spec/kitchen/lifecycle_hooks_spec.rb
jasonwbarnett: I agree! I am going to put some time on my calendar this month to take care of this.

jasonwbarnett:
@clintoncwolfe I think we're ready to rock. I added a couple tests around includes/excludes (positive and negative) and added docs to describe the new feature. Please review and let me know if you see anything else that should be done.

clintoncwolfe (approved):
Hey, sorry this sat for so long - this looks good.

jasonwbarnett:
I just rebased and pushed back up.

tas50:
Nice work @jasonwbarnett. This is very handy

PR #1684, opened 2020-09-10
use autoload to improve perf of kitchen --version

mwrock:
This shaves about 1 second off a `kitchen --version`. Unfortunately it does not improve other commands too much. The way the transports load deps could be refactored so that the underlying (and large) deps are not loaded at plugin load time. And of course you need `YAML` to do just about anything.

Signed-off-by: mwrock <matt@mattwrock.com>


clintoncwolfe (approved):
Thanks!

PR #1725, opened 2021-02-27
Support chef-cli in the policyfile provisioner

lamont-granquist:
Bonus fix in here to spec_helper.rb which probably does most of the
work to fix t-k for ruby 3.0


test-kitchen.gemspec
```diff
@@ -33,6 +33,7 @@ Gem::Specification.new do |gem|
   gem.add_dependency "winrm",              "~> 2.0"
   gem.add_dependency "winrm-elevated",     "~> 1.0"
   gem.add_dependency "winrm-fs",           "~> 1.1"
+  gem.add_dependency "chef-utils",         ">= 16.4.35"
```
lamont-granquist, line 36: This version has a patch in it which supports windows PATHEXT in the which helper

Issue #1723, opened 2021-02-14
Switch the default provisioner in kitchen init from chef-solo to chef-zero


tas50:
As part of the eventual goal of removing the chef-solo provisioner, we need to switch the default configs over to chef-zero.

## Acceptance Criteria

- Running kitchen init produces a config with chef-zero configs
- Any documentation references to the default mention chef-zero

Issue #1708, opened 2021-01-11
Rename chef-zero provisioner to chef-infra


tas50:
chef-zero doesn't really mean anything to end users. We should rename this provisioner (with backwards compat) to chef-infra and make it the new default.

PR #1730, opened 2021-03-23
Add a new chef_infra alias to the chef_zero provisioner

i5pranay93:
# Description

We are renaming provisioner chef_zero to chef_infra and making it backward compatible,  also replacing chef_infra as the default provisioner from chef_solo.

## Issues Resolved

fixes https://github.com/test-kitchen/test-kitchen/issues/1708
fixes https://github.com/test-kitchen/test-kitchen/issues/1723

## Check List


.idea/workspace.xml
```diff
@@ -0,0 +1,860 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
tas50, line 1: You'll want to skip this file and probably add .idea to the gitignore

i5pranay93: Done.

i5pranay93:
The 4 failing test cases are because of dokken  provisioner which is based on chef_zero and is expecting to load the chef_zero file, which has been renamed

looks like ll need to update the dokken provisioner

https://github.com/test-kitchen/kitchen-dokken/blob/f4e267891780298e4bba9bc0f2652f6e04a7531f/lib/kitchen/provisioner/dokken.rb#L19-L28

I am creating a new issue for this in dokken repo, and will create pr for this as well? @marcparadise @tas50

lib/kitchen/provisioner.rb
```diff
@@ -34,7 +34,16 @@ module Provisioner
     # @return [Provisioner::Base] a provisioner instance
     # @raise [ClientError] if a provisioner instance could not be created
     def self.for_plugin(plugin, config)
+      plugin, config = change_provider_chef_zero(plugin, config)
       Kitchen::Plugin.load(self, plugin, config)
```
marcparadise, line 38: Instead of a separate function, what about moving the `if` block of `change_provider_chef_zero` into this function? It's a little more concise, falls into the domain of what we'd expect this function to do (determining the right plugin to load) and can be tested by verifying the parameters of Plugin.load for a given set of inputs.

marcparadise (changes requested):
Please also add a test to Provisioner that verifies that behavior of `self.for_plugin` for solo, infra. infra will also serve to verify the generic case of nothing being modified.

lib/kitchen/provisioner.rb
i5pranay93: @marcparadise it's updated now.

marcparadise (changes requested):
This looks good to me, but two additional issues have occurred to me since my first reading. 

First is that the class name is now inconsistent with the file name. Second is that this is a breaking API change for anyone who has inherited from ChefZero. Workstation is one that will need updating[1]. 

What about addressing both by renaming this file, but keeping a lib/provisioner/chef_zero.rb that requires provisioner/chef_infra, and declaring `Kitchen::Provisioner::ChefZero < ChefInfra`?

[1] https://github.com/chef/chef-cli/blob/master/lib/kitchen/provisioner/chef_zero_capture.rb#L48

i5pranay93:
Closing this and recreating it in https://github.com/test-kitchen/test-kitchen/pull/1771

PR #1765, opened 2021-06-16
Set always_update_cookbooks to true for Policyfiles/Berkshelf

damacus:
# Description

Without this the user doesn't get their updated cookbooks transferred to
the box.
The user shouldn't need to set this.

## Issues Resolved

n/a

## Check List


tas50:
@damacus can you look at the unit tests here?

damacus:
@tas50 these tests good now.

tas50:
This is setting it for both policyfiles and berkshelf, which we probably don't want since we only think we need this for policyfiles.

lamont-granquist (approved):

Issue #1783, opened 2021-07-06
Jan 1st 2022: Update all docs / generators to use `chef_infra` provisioner


tas50:
We want to make sure that people have pushed out newer Workstation releases that include the updated provisioned name, but on Jan 1st we should update the docs to all reference the new name.

tas50:
Done

PR #1784, opened 2021-07-06
Start the slow migration to chef_infra provisioner

tas50:
We need to make sure Workstation releases in the wild can support this before we cutover docs and generators, but step 1 is at least calling out the new name in docs.

Signed-off-by: Tim Smith <tsmith@chef.io>4

### DIFF
--- a/docs/content/docs/provisioners/chef.md
+++ b/docs/content/docs/provisioners/chef.md
@@ -7,7 +7,7 @@ menu:
     weight: 5
 ---
 
-Test Kitchen includes two provisioners for Chef Infra, `chef_solo` and `chef_zero`, which support nearly identical options. Note: In Test Kitchen 3.0 and later the `chef_zero` provisioner has been renamed to `chef_infra` for improved clarity. You may use either name going forward, but older releases of Test Kitchen will still need to use `chef_zero`.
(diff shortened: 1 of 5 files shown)


tas50:
On Jan 1st we'll cutover the docs for the new provisioned: https://github.com/test-kitchen/test-kitchen/issues/1783

docs/content/docs/provisioners/chef.md
```diff
@@ -7,7 +7,7 @@ menu:
     weight: 5
 ---
 
-Test Kitchen includes two provisioners for Chef Infra, `chef_solo` and `chef_zero`, which support nearly identical options.
+Test Kitchen includes two provisioners for Chef Infra, `chef_solo` and `chef_zero`, which support nearly identical options. Note: In Test Kitchen 3.0 and later the `chef_zero` provisioner has been renamed to `chef_infra` for improved clarity. You may use either name going forward, but older releases of Test Kitchen will still need to use `chef_zero`.
```
kagarmoe, line 10:
```suggestion
Test Kitchen includes two provisioners for Chef Infra, `chef_solo` and `chef_infra`, that support similar options.  `chef_zero` was renamed `chef_infra` in Test Kitchen 3.0. This change is backward-compatible, and both names work starting with Test Kitchen 3.0. In older releases, you will need to use `chef_zero`.
```

kagarmoe (changes requested):

PR #1786, opened 2021-07-06
Update Kitchen -> Test Kitchen in the docs

tas50:
Historically we've had way to many names for Test Kitchen. Use the full
name to avoid confusion.


kagarmoe (approved):
tenor-17468527

PR #1796, opened 2021-09-01
fix progress reporting in ssh transport

karmix:
Signed-off-by: Doug Knight <doug.knight@karmix.org>

# Description

Add some tests and make a trivial fix:

Uploaded files are not being logged in debug output as intended.  It's pretty obvious in this case that the author meant to pass a code block to `session.scp.upload` instead of `waits.push`.

## Issues Resolved

_This only corrects info in debug logs; no issue created, see description._

## Check List

- [ ] ~~New functionality has been documented in the README if applicable.~~ (_N/A: Bug Fix._)

### DIFF
--- a/lib/kitchen/transport/ssh.rb
+++ b/lib/kitchen/transport/ssh.rb
@@ -176,9 +176,11 @@ module Kitchen
             Array(locals).map do |local|
               opts = File.directory?(local) ? { recursive: true } : {}
 
-              waits.push session.scp.upload(local, remote, opts) do |_ch, name, sent, total|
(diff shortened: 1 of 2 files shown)


karmix:
The two test cases I added essentially reproduce the entire test that precedes them before they check the debug output.  This is not ideal, because there are situations where the new tests may fail early with an assert about a packet mismatch before it gets to check whether the debug logs are correct.  However I do not see a way to make `Net::SSH::Test` replay a `PacketStream` without it making assertions, and replaying a `PacketStream` appears to be the only way to trigger a call to the progress callback without mocking `Net::SCP#upload` and calling the callback directly.

tas50:
Thanks @karmix

PR #1799, opened 2021-09-07
Mention podman in docs and fix typos

tas50:


georgettica:
@tas50 how does podman work as a drop in replacement?

I didn't see a config option of using podman instead of docker

Issue #1729, opened 2021-03-18
Add option for ruby code for lifecyle hooks


mbaitelman:
### :speaking_head: Foreword

Every so often I stumble across a problem that could be easily solved with a touch of ruby code.

## :frowning_person: Problem Statement

There are times where running ruby code directly could solve issues I have when writing kitchen tests. Most recently I was trying to think about how to do an OS-agnostic sleep for 2 minutes after a destroy when reusing a hardcoded IP for AWS EC2. 
Being able to add a ruby `sleep 120` would have made this quite simple. 

## :grey_question: Possible Solution

I see this working similar to the not_if/only_if code in Chef. 
That would allow for running either scripts or ruby code.

PR #1822, opened 2021-11-02
Add support for Ruby hooks

mbaitelman:


mbaitelman:
This was very WIP for #1729 
I was playing around with it via the VScode in-browser tool and it created it as a PR. 
If I finish it up would it be considered for release?

damacus:
Probably something to bring up with the other maintainers. I'm personally not keen on Ruby the kitchen.yml. 

I'm also not sure why this branch looked stale 🤔 (which is why I deleted it)

Issue #650, opened 2015-03-25
Additional platforms in global/local configs don't get merged


mconigliaro:
I have a global config (`~/.kitchen/config.yml`) like this:

```
platforms:
  - name: aws
    driver:
      name: ec2
      aws_access_key_id: <%= ENV['AWS_ACCESS_KEY_ID'] %>
      aws_secret_access_key: <%= ENV['AWS_SECRET_ACCESS_KEY'] %>
      aws_ssh_key_id: mike
      ssh_key: <%= ENV['HOME'] %>/.ssh/mike.pem
      region: <%= ENV['AWS_DEFAULT_REGION'] %>
      availability_zone: <%= ENV['AWS_DEFAULT_REGION'] %>a
      flavor_id: t2.micro
      image_id: ami-6889d200
      username: ubuntu
```

I was hoping this would automagically create an aws platform in all my cookbooks, but these options don't seem to get merged the way I expected. FWIW, the driver options do seem to get merged. For example:

```
driver:
  aws_access_key_id: <%= ENV['AWS_ACCESS_KEY_ID'] %>
  aws_secret_access_key: <%= ENV['AWS_SECRET_ACCESS_KEY'] %>
  aws_ssh_key_id: mike
  ssh_key: <%= ENV['HOME'] %>/.ssh/mike.pem
  region: <%= ENV['AWS_DEFAULT_REGION'] %>
  availability_zone: <%= ENV['AWS_DEFAULT_REGION'] %>a
  flavor_id: t2.micro
  image_id: ami-6889d200
  username: ubuntu
```

The only problem is now I need to go to all my cookbooks and manually add an aws platform with the driver overridden:

```
platforms:
  - name: aws
    driver:
      name: ec2
```

Also, additional platforms in a `.kitchen.local.yml` seem to completely overwrite the platforms in a `.kitchen.yml` file.

Issue #778, opened 2015-07-14
Merge platform and suite run_list between project_config and local_config


patcon:
I keep a `.kitchen.digitalocean.yml` file that I copy to `.kitchen.local.yml` for doing travis tests. I keep all the digitalocean-specific things there, so I override the named platform versions with a new hash that has the `image` key:

``` yaml
# .kitchen.yml
driver: vagrant

platforms:
- name: centos-6
  driver_config:
    box: chef/centos-6.6
  run_list:
  - recipe[yum]
```

``` yaml
# .kitchen.digitalocean.yml
driver: digitalocean

platforms:
- name: centos-6
  driver_config:
    image: centos-6-5-x64
```

It seems that the run_list in the project_config is overridden by the empty run_list in local_config.

Before I dig in, is this expected behaviour? The merging seems to be pretty seamless and as-expected everywhere else :)

Thanks!

cheeseplus:
This seems like a bug to me but I'll have to ask some folks who know the code better. Can this be repro'd still?

ghostsquad:
Oh please... deep merge of all the things is so needed!
As an example:

`.kitchen.yml`

```yaml
---
driver:
  name: vagrant

provisioner:
  name: chef_zero
  require_chef_omnibus: 12.21.1
  always_update_cookbooks: true
  cookbooks_path:
    - ../../site-cookbooks
    - ../../librarian-cookbooks

verifier:
  name: inspec

platforms:
  - name: ubuntu-16.04

suites:
  - name: default
    run_list:
      - recipe[foo::default]
    verifier:
      inspec_tests:
        - spec/integration/recipes/default_spec.rb
    attributes:
      hello: world
      username: bob
      password: <redacted>
```

I would very much like to be able to do this:
`.kitchen.local.yml`

```yaml
---
suites:
  - name: default
     attributes:
       password: mypassword
```

nacc:
FWIW, I think I figured out why this happens and it's sort of hairy

[...]

Issue #1826, opened 2021-11-16
Unit tests are failing on master branch


ashiqueps:
## :speaking_head: Foreword

Thank for taking the time to fill this bug report fully. Without it we may not be able to fix the bug, and the issue may be closed without resolution.

## :ghost: Brief Description

Some of the unit tests are failing on master branch. This is causing the azure builds to fail for all the open pull requests. The tests are failing on Windows platform only. 

## Version

test-kitchen - v3.1.1

## Environment

Operating System - Windows
## Scenario

Unit tests are failing on windows platform causing the azure builds to fail on open PRs. 

## Steps to Reproduce

Run the unit tests on Windows OS.

## Actual Result
```
Finished in 22.246829s, 105.2285 runs/s, 165.6416 assertions/s.

1) Failure:
Kitchen::Driver::SSHBase::#converge::transferring files#test_0001_uploads files [D:/a/1/s/spec/kitchen/driver/ssh_base_spec.rb:453]:
not all expectations were satisfied
unsatisfied expectations:
- expected exactly once, invoked never: #<Mock:0x1baae2831c8>.upload(["/tmp/sandbox/stuff"], "/rooty")
satisfied expectations:
- allowed any number of times, invoked once: #<Mock:0x1baae249b58>.[](:downloads)
- allowed any number of times, invoked once: #<Mock:0x1baae249b58>.[](:root_path)

2) Failure:
Kitchen::Verifier::Base::#call#test_0006_uploads sandbox files [D:/a/1/s/spec/kitchen/verifier/base_spec.rb:220]:
not all expectations were satisfied
unsatisfied expectations:
- expected exactly once, invoked never: #<Mock:transport_connection>.upload(["/tmp/sandbox/stuff"], "/tmp/verifier")
satisfied expectations:
- allowed any number of times, invoked never: #<Mock:0x1baac9a0340>.shell_type(any_parameters)
- allowed any number of times, invoked once: #<Mock:0x1baac9a0340>.os_type(any_parameters)
- allowed any number of times, invoked never: #<Mock:0x1baacafea20>.name(any_parameters)
- allowed any number of times, invoked once: #<Mock:transport>.connection(any_parameters)

3) Failure:
Kitchen::Provisioner::Base::#call#test_0006_uploads sandbox files [D:/a/1/s/spec/kitchen/provisioner/base_spec.rb:212]:
not all expectations were satisfied
unsatisfied expectations:
[...]

PR #1831, opened 2021-12-14
HashRecursiveMerge: support arrays in recursive merge

nacc:
# Description

When HashRecursiveMerge currently encounters a non-Hash, it will simply return the "new" value. This is sub-optimal for cases where we might want to have some top-level defaults (e.g., for platforms) but certain cookbooks may want to define additional ones. The current code will only leverage the additional ones, instead of appending them to the global ones.

At this point, I'm looking for comments and if this seems like an appropriate direction, in which case I will add tests (and think harder about corner cases).

## Issues Resolved

https://github.com/test-kitchen/test-kitchen/issues/650
https://github.com/test-kitchen/test-kitchen/issues/778

## Check List

### DIFF
--- a/lib/kitchen/loader/yaml.rb
+++ b/lib/kitchen/loader/yaml.rb
@@ -114,11 +114,11 @@ module Kitchen
       # @api private
       def combined_hash
         y = if @process_global
-              normalize(global_yaml).rmerge(normalize(yaml))
+              normalize(global_yaml).rmerge(normalize(yaml), merge_arrays=true)
             else
               normalize(yaml)
             end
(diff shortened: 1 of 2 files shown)


nacc:
I realize that perhaps for safety purposes, we should maybe avoid changing the default behavior since it is used by the within-the-yaml merging (eg between suites, platforms). I think I can do that pretty easily by adding an opt-in parameter to rmerge.

nacc:
I think the CI failures are the same as https://github.com/test-kitchen/test-kitchen/issues/1826

PR #1840, opened 2022-01-28
[INFCT-7] Unit tests are failing on master branch

ashiqueps:
# Description

Some of the unit tests are failing on master branch. This is causing the azure builds to fail for all the open pull requests. The tests are failing on Windows platform only.

Due to the different folder structure in windows when compared to the Mac/Linux, the file util is creating a single file instead of the complete path. This is causing the expectation to fail. The fix for the issue is to stub the method that fetch the items in the directory as well. 

## Issues Resolved

https://github.com/test-kitchen/test-kitchen/issues/1826

## Check List

- [x] New functionality has been documented in the README if applicable.

### DIFF
--- a/lib/kitchen/provisioner/base.rb
+++ b/lib/kitchen/provisioner/base.rb
@@ -182,6 +182,13 @@ module Kitchen
           "trying to access the path."
       end
 
+      # Returns the list of items in the sandbox directory
+      #
+      # @return [String] path of items in the sandbox directory
+      def sandbox_dirs
+        Util.list_directory sandbox_path
+      end
+
       # Deletes the sandbox path. Without calling this method, the sandbox path
       # will persist after the process terminates. In other words, cleanup is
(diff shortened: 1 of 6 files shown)


lib/kitchen/provisioner/base.rb
```diff
@@ -182,6 +181,13 @@ def sandbox_path
           "trying to access the path."
       end
 
+      # Returns the list of items in the sandbox directory
+      #
+      # @return [String] path of items in the sandbox directory
+      def sandbox_dirs
+        Util.list_directory sandbox_path
```
tecracer-theinen, line 188: Bit picky, but might be good to adjust to the style on `lib/kitchen/verifier/base.rb` ?

ashiqueps: Updated. Thanks @tecracer-theinen

damacus (approved):

Issue #1904, opened 2022-07-19
Update net-ssh to support RHEL 9 / AlmaLinux 9 / RockyLinux 9


sspans-sbp:
## :speaking_head: Foreword

Thank for taking the time to fill this bug report fully. Without it we may not be able to fix the bug, and the issue may be closed without resolution.

## :ghost: Brief Description

The current version of net-ssh that's used by test-kitchen doesn't support SHA2 signatures.
The older SHA1 signatures have been deprecated and are not enabled in EL9 distributions.

The most recent net-ssh release 7.0 is the first version that supports the required signatures.

https://github.com/net-ssh/net-ssh/commit/a45f54fe1de434605af0b7195dd9a91bccd2cec5
https://github.com/net-ssh/net-ssh/commit/d9b2f09005bf95b6af3e263794f70dd6b9bda10a

OpenSSH deprecation notice: https://www.openssh.com/txt/release-8.7

## Version

3.3.1

## Environment

[Details about the environment such as the Operating System, cookbook details, etc...]

## Scenario

I'm trying to run 'kitchen converge' on an AlmaLinux 9 host.

## Steps to Reproduce

Run a kitchen test on an Almalinux 9 instance with the default sshd config.

## Expected Result

SSH connections to newer OpenSSH work.

## Actual Result

The SSH connection fails:
```
$ kitchen test --destroy=always $SUITE
-----> Starting Test Kitchen (v3.3.1)
-----> Cleaning up any prior instances of <default-almalinux-9>
-----> Destroying <default-almalinux-9>...
       Finished destroying <default-almalinux-9> (0m0.00s).
-----> Testing <default-almalinux-9>
-----> Creating <default-almalinux-9>...
       Bringing machine 'default' up with 'cosmic' provider...
       ==> default:  -- Start port : 22
       ==> default:  -- End port   : 22
       ==> default:  -- ICMP code  : 
       ==> default:  -- ICMP type  : 
       ==> default: Waiting for SSH to become available...
       ==> default: Machine is booted and ready for use!
       ==> default: Machine not provisioned because `--no-provision` is specified.
       Waiting for SSH service on x.x.X.131:56732, retrying in 3 seconds
       Waiting for SSH service on x.x.X.131:56732, retrying in 3 seconds
       Waiting for SSH service on x.x.X.131:56732, retrying in 3 seconds
       Waiting for SSH service on x.x.X.131:56732, retrying in 3 seconds
```

[...]

PR #1905, opened 2022-07-19
chore: bump net-ssh

testeddoughnut:
# Description

Update net-ssh requirement from >= 2.9, < 7.0 to >= 2.9, < 8.0 to support RSA keys on recent versions of OpenSSH, see https://www.openssh.com/txt/release-8.7 for more context.

## Issues Resolved

https://github.com/test-kitchen/test-kitchen/issues/1904

## Type of Change

Our release process assumes you are using [Conventional Commit messages](https://www.conventionalcommits.org/en/v1.0.0/).

The most important prefixes you should have in mind are:

- `_fix_`: which represents bug fixes, and correlates to a SemVer patch.
- `_feat_`: which represents a new feature, and correlates to a SemVer minor.
- `_feat!_`:, or `fix!:`, `refactor!:`, etc., which represent a breaking change (indicated by the !) and will result in a major version change.

If you have not included a conventional commit message this can be fixed on merge.

## Check List

- [ ] New functionality includes tests
- [x] All tests pass

### DIFF
--- a/test-kitchen.gemspec
+++ b/test-kitchen.gemspec
@@ -24,7 +24,7 @@ Gem::Specification.new do |gem|
 
   gem.add_dependency "mixlib-shellout",    ">= 1.2", "< 4.0"
   gem.add_dependency "net-scp",            ">= 1.1", "< 4.0" # pinning until we can confirm 4+ works
-  gem.add_dependency "net-ssh",            ">= 2.9", "< 7.0" # pinning until we can confirm 7+ works
+  gem.add_dependency "net-ssh",            ">= 2.9", "< 8.0" # pinning until we can confirm 8+ works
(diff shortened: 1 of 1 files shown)


tas50:
@testeddoughnut Hey sorry but I merged in the same change from dependabot before I saw this. It looks like we need the net net-scp to get release before 7.0 can actually get pulled in since net-scp pins to < 7 right now.

Stromweld:
@tas50 along the same lines as this, it looks like the future of scp is going away. Is this going to affect the use of net-scp and require a switch to using net-sftp instead? https://www.redhat.com/en/blog/openssh-scp-deprecation-rhel-9-what-you-need-know

PR #1910, opened 2022-08-23
Work with net-scp 4.0 when it ships

tas50:
This will allow pulling in the new net-ssh which supports new algos

Signed-off-by: Tim Smith <tsmith84@gmail.com>

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,12 +10,6 @@ group :integration do
   gem "kitchen-vagrant"
 end
 
-group :debug do
-  gem "pry", "~>0.12"
-  gem "pry-byebug"
-  gem "pry-stack_explorer"
-end
-
 group :chefstyle do
   gem "chefstyle", "2.2.2"
 end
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -10,12 +10,6 @@ group :integration do
   gem "kitchen-vagrant"
 end
 
-group :debug do
-  gem "pry", "~>0.12"
```
tas50, line 14: New ruby replaces pry so nuke all this

vkarve-chef (approved):

Stromweld:
net-scp just officially release 4.0 to rubygems. Is there anything that needs to be done to build a new release of test-kitchen with the latest net-scp and net-ssh to fix these ssh login issues with the latest releases of rhel and ubuntu?

tas50:
It's being held back by train from the InSpec team. They need to update the dep on net-scp and release train.

clintoncwolfe:
done - https://rubygems.org/gems/train-core/versions/3.10.7

PR #1916, opened 2022-10-21
Always download files even if verifier fails

### LINKED ISSUE #1643: File Downloads from a Verifier fail if the verify step fails
## Description

I'm trying to have Test Kitchen retrieve some log files from the instance after the verify step to help with diagnosing failures and publish some test results that the verifier generates on the target machine as XML.

Currently, it seems that if a verify step results in failure, no files can be downloaded once it's completed.

## Kitchen Version

```ruby
Test Kitchen version 2.4.0
```

## ChefDK Version

```ruby
ChefDK version: 4.5.0
Chef Infra Client version: 15.4.45
Chef InSpec version: 4.18.0
Test Kitchen version: 2.4.0
Foodcritic version: 16.1.1
Cookstyle version: 5.9.3
```

## Platform Version

MacOS Catalina & Windows Server 2019

## Replication Case

Create a simple `kitchen.yml` and basic test cases with a Windows-based agent and the `kitchen-pester` verifier. Something very simplistic will suffice for the tests, save as a PS1 file and reference in the yml:

```ps1
Describe "Minimal Test Case" {
    It 'should save a log file' {
        "test log contents" | Set-Content -Path "C:/TEMP/test.log"
    }

    It 'should fail the test' {
        throw
    }
}
```

Sample yml file:

```yml
---
driver:
  name: ec2
  region: us-east-1
  availability_zone: a
  retryable_tries: 120
  retryable_sleep: 5
  tags:
    Name: test instance

provisioner:
  name: chef_zero
  product_name: chef
  product_version: 15.3.14

verifier:
  name: pester

transport:
  name: winrm
  elevated: false

platforms:
  - name: windows-2019
    driver:
      image_id: <insert basic AMI image here, or just use a more basic VM setup>
      instance_type: t2.medium
      tags:
        Name: test-kitchen-repro
    verifier:
      downloads:
        "C:/Users/Administrator/AppData/Local/Temp/verifier/PesterTestResults.xml": testresults/TestResults.xml
        "C:/Temp": logs/test

suites:
  - name: chocolatey-tests
    provisioner:
      policyfile: policyfile.rb
    verifier:
      name: pester
      test_folder: ./test
    attributes:
```

Running `kitchen verify` at this point will run through the tests and fail to retrieve the requested files. However, if the failing test is removed, the files can be retrieved.

There have been numerous attempts to handle this in the verifier itself as well, but it seems that any attempts the verifier makes to download files is prevented by Test Kitchen registering the failure of the verifier.

If there's a way to handle this in the verifier (report failure without test kitchen preventing file downloads), please do let me know, it's very annoying trying to get at some basic logging information from the machine at present without wanting it printed directly to console output all of the time.

brycekahle:
# Description

Download files, even if the verifier run command fails.

## Issues Resolved

#1643 

## Type of Change

Fix

## Check List

- [x] All tests pass

### DIFF
--- a/lib/kitchen/verifier/base.rb
+++ b/lib/kitchen/verifier/base.rb
@@ -73,14 +73,17 @@ module Kitchen
           conn.upload(sandbox_dirs, config[:root_path])
           debug("Transfer complete")
           conn.execute(prepare_command)
-          conn.execute(run_command)
-
-          info("Downloading files from #{instance.to_str}")
-          config[:downloads].to_h.each do |remotes, local|
-            debug("Downloading #{Array(remotes).join(", ")} to #{local}")
-            conn.download(remotes, local)
+          
+          begin
+            conn.execute(run_command)
+          ensure
+            info("Downloading files from #{instance.to_str}")
+            config[:downloads].to_h.each do |remotes, local|
+              debug("Downloading #{Array(remotes).join(", ")} to #{local}")
+              conn.download(remotes, local)
+            end
+            debug("Download complete")
           end
-          debug("Download complete")
         end
       rescue Kitchen::Transport::TransportFailed => ex
         raise ActionFailed, ex.message
--- a/spec/kitchen/verifier/base_spec.rb
+++ b/spec/kitchen/verifier/base_spec.rb
@@ -250,6 +250,22 @@ describe Kitchen::Verifier::Base do
       cmd
     end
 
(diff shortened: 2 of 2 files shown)


brycekahle:
CI seems to be broken, so I cannot tell if the tests pass.

tas50 (approved):

PR #1931, opened 2023-02-13
If always_update config is setup - Then only chef update (which does a chef install)

clickthisnick:
# Description

Right now if you specify always_update in your kitchen.yml - If the policyfile lock is incorrect such that the command chef install fails, the kitchen tests will fail.

Instead if you are specifying always_update it should just run the chef update command which behind the scenes also does a [chef install](https://github.com/chef/chef-cli/blob/main/lib/chef-cli/command/update.rb#L102-L104).

This also makes it faster, currently it is doing chef install then running chef update which also calls install

## Issues Resolved

N/A

## Type of Change

Feat - this perhaps is a fix

## Check List

- [X] New functionality includes tests 
- [X] All tests pass
- [X] Commit message includes a [Conventional Commit Message](https://www.conventionalcommits.org/en/v1.0.0)


lib/kitchen/provisioner/chef/policyfile.rb
```diff
@@ -69,20 +69,20 @@ def resolve
...
-          run_command("#{cli_path} install #{escape_path(policyfile)} --chef-license #{license}")
-
-          if always_update
-            info("Updating policy lock using `#{cli_path} update`")
-            run_command("#{cli_path} update #{escape_path(policyfile)} --chef-license #{license}")
+            run_command("chef install #{escape_path(policyfile)}")
+            if always_update
+              info("Updating policy lock using `#{cli_path} update`")
+              run_command("chef update #{escape_path(policyfile)}")
+            end
```
Stromweld, line 86: Is this even needed? if a lockfile doesn't exist then chef-install will pull all of the current deps and thus update would be redundant. Subsequent kitchen-converges would be caught in the first if statement with && always_update enabled and then would run update as needed.

clickthisnick: you are right - will remove, thanks

clickthisnick:
It looks like the tests that are failing are also failing on main:
https://github.com/test-kitchen/test-kitchen/pull/1920

PR #2021, opened 2025-09-03
feat!: adds chef license for omnibus downloads and cinc support

Stromweld:
# Description

Adds support for chef's commercial download api requiring a license key to download chef products as well as future use of license key/server set as environment variables for chef-client runs. Also adds support for using cinc client and defaults to installing cinc if no product_name is given.

## Check List

- [x] All tests pass

### DIFF
--- a/lib/kitchen/generator/init.rb
+++ b/lib/kitchen/generator/init.rb
@@ -39,7 +39,7 @@ module Kitchen
       class_option :provisioner,
         type: :string,
         aliases: "-P",
-        default: "chef_solo",
+        default: "chef-infra",
         desc: <<-D.gsub(/^\s+/, "").tr("\n", " ")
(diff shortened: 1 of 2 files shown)


docs/content/docs/provisioners/chef.md
```diff
@@ -29,8 +37,10 @@ provisioner:
     - deploy_resource # deprecation key name
     - chef-23 # deprecation numeric ID
     - recipes/install.rb:22 # specific line in a file
-  product_name: chef # which package to install chef || chef-workstation
+  product_name: nil # defaults to installing cinc-client when set to cinc or nil, to install chef-client package set to chef || chef-workstation
```
tas50, line 40: why not just make this cinc as the default?

Stromweld:
outdated

PR #2022, opened 2025-09-04
fix: updates to fix tests

Stromweld:
# Description

fix tests and testing deprectations

## Check List

- [x] All tests pass


Gemfile
```diff
@@ -19,6 +19,8 @@ group :integration do
   gem "kitchen-dokken"
   gem "kitchen-inspec"
   gem "kitchen-vagrant"
+  gem "inspec", "~> 5.22" # Fix dependency conflicts for ruby < 3.4
+  gem "train", "~> 3.10" # Fix dependency conflicts for ruby < 3.4
```
damacus, line 23: Should we make this the last <3.4 release?

damacus (approved):

.github/workflows/lint.yml
```diff
@@ -46,15 +46,19 @@ jobs:
...
+          brew tap hashicorp/tap
+          brew update
+          brew install --cask hashicorp/tap/hashicorp-vagrant
+          brew install --cask virtualbox
       - uses: actions/checkout@v5
       - uses: ruby/setup-ruby@v1
         with:
           ruby-version: ${{ matrix.ruby }}
-          bundler-cache: true
+          bundler-cache: false
```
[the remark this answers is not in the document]
Stromweld: I changed that as a test, they were failing previous runs with it set to true

Gemfile
Stromweld: that'd make sense especially after chef-client 19 gets released as it'll also be based on ruby 3.4.

Gemfile
```diff
@@ -19,6 +19,8 @@ group :integration do
   gem "kitchen-dokken"
   gem "kitchen-inspec"
   gem "kitchen-vagrant"
+  gem "inspec", "~> 5.22" # Fix dependency conflicts for ruby < 3.4
```
tas50, line 22: inspec claims to work with ruby 3.1. Is that not true?

PR #2054, opened 2026-06-02
refactor: simplify logging, instance actions, and config munging

damacus:
## Summary
This refactor isolates several long-standing core seams without changing Test Kitchen's public CLI, plugin, or config surface.

## Reviewer Notes
- Logging now has explicit internal collaborators for sink fan-out, device creation, streamed line buffering, and prefix normalization. The existing logger API and output format are preserved, while block-style logging is forwarded correctly through both Kitchen::Logger and the Logging mixin.
- Instance action execution now runs through an ActionRunner that owns timing, state persistence, plugin serialization, and failure wrapping. Legacy Driver::SSHBase behavior moved into a small adapter so the normal action path is easier to follow.
- Command action fan-out now protects shared error collection and restores Thread.abort_on_exception after fail-fast runs.
- DataMunger still remains the facade and keeps destructive normalization semantics, but legacy conversion order and merge resolution are now named collaborators for easier follow-up work.

## Suggested Review Focus
- Confirm the logger refactor preserves stdout/file formatting and lazy block semantics.
- Check that legacy SSH adapter behavior still matches the previous converge/setup/verify/login paths.
- Check that DataMunger extraction keeps conversion ordering obvious and does not imply immutable input handling yet.


Gemfile
```diff
@@ -16,9 +16,11 @@ end
 
 group :integration do
   gem "chef-cli"
+  gem "cinc-auditor-bin", source: "https://rubygems.cinc.sh"
+  gem "inspec", source: "https://rubygems.cinc.sh"
   gem "kitchen-dokken"
-  gem "kitchen-vagrant"
   gem "kitchen-inspec"
```
Stromweld: I'd replace this with new kitchen-cinc-auditor gem

Gemfile
```diff
@@ -16,9 +16,11 @@ end
...
+  gem "kitchen-vagrant"
   gem "kitchen-omnibus-chef", ">= 1.0"
```
Stromweld: I'd replace this with new kitchen-cinc gem

.github/actions/install-cinc-client/action.yml
```diff
@@ -0,0 +1,99 @@
+---
```
Stromweld: replacing the gems with cinc versions in gemfile and updating kitchen.yml provisioner product to cinc would accomplish the same thing and not need this action.

lib/kitchen/verifier/cinc_auditor.rb
```diff
@@ -0,0 +1,27 @@
+#
```
Stromweld: is this needed if we have the new kitchen-cinc-auditor? I'd recommend adding an alias in kitchen-cinc-auditor inspec verifier to use CincAuditor.

Stromweld (changes requested):
LGTM - other than the switch from chef to cinc gems and kitchen config settings, I have to admit a lot of the code is a bit above my head.

PR #2068, opened 2026-06-03
fix(provisioner): download provisioner files even when converge fails (#2036)

### LINKED ISSUE #2036: Add ability to retrieve files after each step
### :speaking_head: Foreword

Thank for taking the time to fill this feature request fully. Without it we may not be able to , and the issue may be closed without resolution.

## :frowning_person: Problem Statement

I have run into several instances where our converge step has failed for "reasons". In order to assist with identifying these reasons, it would be incredibly helpful to be able to obtain log files from the SUT.

## :grey_question: Possible Solution

A possible solution is an additional option for files or directories to retrieve when a SUT action is completed. Perhaps one for retrieval on success and another on failure?

## :arrow_heading_up: Describe alternatives you've considered

My initial attempt was to include log files in our kitchen-pester configuration to retrieve them for us, and this works when everything works. However, when the converge step fails, then the verify step is never executed, and so we never retrieve the files.

## :heavy_plus_sign: Additional context

Add any other context about the problem here. e.g. related issues or existing pull requests.

damacus:
## What

Closes #2036.

Make the provisioner retrieve its configured `downloads:` files **even when converge fails**, by wrapping the converge run command in `begin/ensure`. Previously the download only ran after a *successful* converge, so log files needed to debug a failed converge were never retrieved.

## Why

[#2036](https://github.com/test-kitchen/test-kitchen/issues/2036) reports that when converge fails, the verify step never runs, so any file-retrieval attached to the verifier never happens — leaving no way to pull SUT logs that explain the failure.

This turned out to be a concrete asymmetry, not a missing feature:

- The **verifier** already downloads files even when its run fails — `Kitchen::Verifier::Base#call` wraps the run command in `begin/ensure` (added in #1916).
- The **provisioner** has the same `downloads:` feature, but ran the download loop *inline after* `execute_with_retry`, so a failed converge skipped it entirely.

This PR makes the provisioner mirror the verifier.

## Change

`lib/kitchen/provisioner/base.rb` — wrap `conn.execute_with_retry(...)` in `begin/ensure`, with the existing `downloads` loop in the `ensure`. On failure, files are downloaded, then the original exception propagates to the existing `rescue TransportFailed => raise ActionFailed`.

No new or changed configuration — it uses the existing `downloads:` key.

## Testing

- New spec `"downloads files when run fails"` in `spec/kitchen/provisioner/base_spec.rb`, mirroring the verifier's equivalent. Written test-first: it fails on the old code (downloads skipped on failure) and passes with the change.
- `provisioner` spec: **61 runs, 0 failures**. `verifier` spec: **40 runs, 0 failures**. `rake style`: clean.

## Reviewer notes

- **Known limitation (shared with the verifier):** if converge fails *and* a listed file can't be downloaded (e.g. it never got created), the download's `TransportFailed` surfaces instead of the converge error. The verifier already behaves this way; kept consistent here. Happy to address both separately if preferred.
- This branch also includes an internal design doc at `docs/superpowers/specs/2026-06-03-provisioner-downloads-on-converge-failure-design.md`. Let me know if you'd prefer it dropped from the PR.

### DIFF
--- a/docs/content/docs/provisioners/_index.md
+++ b/docs/content/docs/provisioners/_index.md
@@ -24,7 +24,7 @@ provisioner:
   wait_for_retry: 30
   uploads: # a Hash of local => remote file mappings to upload at the start of invocation
     "contrib/some_file.cfg": "/etc"
-  downloads: # a Hash of remote => local file mappings to download post-converge
+  downloads: # a Hash of remote => local file mappings to download after converge (downloaded even when converge fails, so logs can be retrieved)
   # if the local value is an existing dir, the file will be copied into it
   # if the local value does not exist, a file with that value as name will be created
     "/tmp/kitchen/client.rb": "./downloads"
--- a/docs/superpowers/specs/2026-06-03-provisioner-downloads-on-converge-failure-design.md
+++ b/docs/superpowers/specs/2026-06-03-provisioner-downloads-on-converge-failure-design.md
@@ -1,0 +1,141 @@
+# Provisioner `downloads` on converge failure — Design
+
+- **Date:** 2026-06-03
+- **Issue:** [test-kitchen#2036 — Add ability to retrieve files after each step](https://github.com/test-kitchen/test-kitchen/issues/2036)
+- **Status:** Approved design, ready for implementation plan
+
+## Problem statement
+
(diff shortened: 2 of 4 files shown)


Stromweld (approved):
that's simple enough

PR #2070, opened 2026-06-18
docs: record plugin contracts

damacus:
## Summary
- Initialize ADR tracking under docs/adr and add ADRs for the current driver, transport, provisioner, and verifier contracts.
- Add reference documentation for the current driver, transport, provisioner, and verifier plugin surfaces.
- Preserve the current messy behavior as documented baseline for future contract improvements.

## Test Plan
- adr list
- git diff --cached --check
- rg -n "[ 	]+$" docs/adr docs/content/docs/reference
- hugo not run: hugo is not installed locally

### DIFF
--- a/.adr-dir
+++ b/.adr-dir
@@ -1,0 +1,1 @@
+docs/adr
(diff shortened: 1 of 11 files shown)


damacus:
I'm aware that some of this is probably wrong. Some of you understand the workings of TK much deeper that I do currently. If there's anything obviously missing/wrong please let me know

PR #2074, opened 2026-06-22
feat: add structured kitchen logs

damacus:
## Summary
- add per-instance NDJSON logs with run, action, session, instance, suite, and platform metadata
- add `kitchen logs` for text-by-default log inspection, explicit `--format ndjson`, level/session filters, follow mode, and all-session reads
- expose driver live status and optional transport probing through `kitchen list`
- fix local dokken verification config to use the existing dokken provisioner

## Apache example
Default output is friendly text:

```sh
mise exec -- bundle exec kitchen logs --all-sessions --level error
```

```text
The `berkshelf' gem is missing and must be installed or cannot be properly activated. Run `gem install berkshelf` or add the following to your Gemfile if you are using Bundler: `gem 'berkshelf'`.
```

NDJSON output is available explicitly:

```sh
mise exec -- bundle exec kitchen logs --all-sessions --level error --format ndjson
```

```json
{"kitchen_run_id":"5bdf5e81-7adf-4f69-91b5-bce22e37b39b","instance":"default-almalinux-8","suite":"default","platform":"almalinux-8","timestamp":"2026-06-22T07:38:03.889943Z","level":"fatal","event_type":"log","message":"The `berkshelf' gem is missing and must be installed or cannot be properly activated. Run `gem install berkshelf` or add the following to your Gemfile if you are using Bundler: `gem 'berkshelf'`.","progname":"default-almalinux-8","sequence":1}
```

The NDJSON stream can be queried with `jq`:

```sh
mise exec -- bundle exec kitchen logs --all-sessions --format ndjson \
  | jq -r 'select(.level == "fatal") | [.level, .instance, .message] | @tsv'
```

```text
fatal	default-almalinux-8	The `berkshelf' gem is missing and must be installed or cannot be properly activated. Run `gem install berkshelf` or add the following to your Gemfile if you are using Bundler: `gem 'berkshelf'`.
```

## Verification
- `bundle exec ruby -Itest -Ilib spec/kitchen/command/logs_spec.rb`
- `bundle exec ruby -Itest -Ilib spec/kitchen/cli_spec.rb`
- `bundle exec rake unit` -> 1696 runs, 2600 assertions, 0 failures
- `git diff --check`
- apache2 smoke checks for default text and explicit NDJSON output

### DIFF
--- a/.github/scripts/kitchen/provisioner/dokken_ci.rb
+++ b/.github/scripts/kitchen/provisioner/dokken_ci.rb
@@ -1,9 +1,9 @@
 # frozen_string_literal: true
 
-require 'kitchen/provisioner/dokken'
-require 'kitchen/provisioner/cinc/berkshelf'
-require 'kitchen/provisioner/cinc/common_sandbox'
-require 'kitchen/provisioner/cinc/policyfile'
+require "kitchen/provisioner/dokken"
+require "kitchen/provisioner/cinc/berkshelf"
+require "kitchen/provisioner/cinc/common_sandbox"
+require "kitchen/provisioner/cinc/policyfile"
 
 module KitchenDokkenCI
   def self.patch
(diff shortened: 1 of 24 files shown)


Stromweld (approved):
LGTM, do we want to add an integration test to see the output and verify functionality?

